PR #8, opened 2017-02-24
pass arguments from prophet() to Stan

davharris:
Being able to pass arguments to Stan via `...` is useful when one wants to control things like the number of chains/iterations/cores or the HMC control parameters.  

This should not change any existing functionality.


seanjtaylor:
Looks good to me.  My only hesitation is if we want to use the `...` in some other way later, but I can't imagine any other uses right now.  @bletham do you have any thoughts?  I guess the alternative is something like `stan.args = list()` but that's a lot clunkier.

bletham:
I think this is great for now, if in the future we need `...` for something else we can handle it in a future version.

PR #11, opened 2017-02-24
invisibly return component plots

davharris:
This will make it easier for users to change the theme or other aspects of the plots, and shouldn't affect any existing behavior because the returned object is invisible.


seanjtaylor:
Thanks for the contribution @davharris! I'm going to let @bletham make the call on this one since he wrote this original code.  Sounds reasonable to me, though!

PR #25, opened 2017-02-26
Make plot axis label names configurable

arnesund:
The plot is currently hardcoded to use 'ds' and 'y' as label names. I'd like to be able to customize them and added two optional parameters to the plot() method to do so.


bletham:
Thanks!

Issue #27, opened 2017-02-26
Missing tick in weekly forecast components plot (Python API)


Tafkas:
In the Python API, the first tick seems to be missing in the weekly forecast components plot. You can see this comparing the two images in the documentation:

 - Python API
     ![](https://facebookincubator.github.io/prophet/static/quick_start_files/quick_start_14_0.png)
     
- R API
    ![](https://facebookincubator.github.io/prophet/static/quick_start_files/quick_start_28_0.png)
    
The result is, that all other ticks move by one position, leaving the last one empty.

bletham:
Fixed in #28 . Thanks!

PR #28, opened 2017-02-26
Fix missing x-axis tick in weekly forecast component plot

Tafkas:
A quick fix for #27


bletham:
Thanks for this bugfix. It looks like matplotlib is putting an xtick at -1 even though it is out of the plot range. For clarity moving forward, I'd prefer to fix this by  explicitly specifying the xticks to be at the right locations. So,  leave line 724 as was but insert
`ax.set_xticks(range(len(days)))`
the line before. Could you update your commit accordingly and I'll merge it in?

PR #54, opened 2017-03-01
Update forecaster.py

gautam1858:
Adding a new forecast algorithm


bletham:
Thanks for the PR. This is a really different type of model than Prophet, and we aren't looking to expand the type of models included in the package right now. We also require the R and Python versions to have the same features, so any new features would need to simultaneously be implemented in R.

Issue #78, opened 2017-03-03
Automatically detect if it makes sense to create weekly seasonality or not


datafool:
Since, creating weekly and yearly seasonality is default feature, it would be good that we check if the data under consideration actually supports weekly seasonality or not. All data which are aggregated at weekly or monthly level will never support weekly and monthly seasonality. 

So it's important that we check if data is aggregated at weekly or monthly level, before setting `self.weekly_seasonality =True`. Agree, it can't be done in class constructor but we can reset the value once we are in fit method and it's pretty easy to use. I have given solution below after providing the error which creeps in because of this. 

On weekly or monthly aggregated data if i call `m.plot_components(forecast)`, the error is `KeyError: u'the label [Monday] is not in the [index]`. This error happens in `plot_components()` method while executing below code block
` days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday',
                    'Friday', 'Saturday']
            y_weekly = [df_s.loc[d]['weekly'] for d in days]
            y_weekly_l = [df_s.loc[d]['weekly_lower'] for d in days]
            y_weekly_u = [df_s.loc[d]['weekly_upper'] for d in days]
`

Now, as weekly or monthly aggregated data will never has these days, plot will always fail, when called with default parameters.

Potential fix:
Step 1: Sort the input dataframe by date, which is being done anyways
Step 2: Get the difference between consecutive date values
Step 3: If min difference is more than 7, then data is not daily and we should set `self.weekly_seasonality =False`

A sample code snippet with the data and its working

```
df = pd.DataFrame({'ds': pd.date_range('2017-03-01', periods=10, freq='W'),
                  'y':[1,2,3,4,5,6,7,8,9,10]})
df = df.sort_values('ds')

if np.min(df.ds - df.ds.shift(1)).days >= 7:
    self.weekly_seasonality =False
```
In above case as data is weekly, we will not create weekly seasonality.

I will create pull request for the same if this problem seems to be one that needs to be addressed

LaurentBerder:
I agree, and also just posted a comment in that sense:
https://github.com/facebookincubator/prophet/issues/76

seanjtaylor:
Awesome!  Thanks for the patch @datafool.  Keeping this issue open until we can an R implementation.

bletham:
I think there are two issues here.

The first is a bug in the plotting which makes it fail if `weekly_seasonality=True` and there is a weekday that is entirely missing from the data. This should be fixed.

[...]

PR #80, opened 2017-03-03
#78 Automatic detection of data aggregation for creating weekly seasonality 

datafool:
#78 
- Changed setup_dataframe, so that if minimum difference between two  dates of data is more than 7 days then weekly seasonality is not created
- This will rectify the error in plot_component, when weekly aggregated data is used. For more detail refer to issue #78


datafool:
Raising another pull request to fix issue with this pull request

PR #96, opened 2017-03-08
Make stan code windows-compatible.

rolando:
The vector/matrix operations fail to compile in windows. Related https://github.com/stan-dev/pystan/issues/308

I think this is due to Eigen library compatibility with MSVC. I have applied this patch to the release on pypi and the build succeed and passes the tests. See here: https://ci.appveyor.com/project/rolando/staged-recipes/build/1.0.111 (conda build)

However, even though this uses ``dot_product`` there is a performance penalty for not using full vectorized matrix operations.

So the question is whether this patch makes sense only for windows (more difficult to maintain) or apply it to all platforms (performance penalty). Of course the alternative is to not have windows support.

**Update:** Latest patch choose non-matrix version for windows builds.


bletham:
I did a performance comparison of the current vs. this version using the Peyton Manning dataset from the quickstart. In the current, over 25 fits it took an average of 2.2s per fit. With this version, over 25 fits it took an average of 4.0s per fit.

I actually expected a bigger performance hit than just ~2x, it's still definitely fast enough!

I think we should be able to have the best of both worlds by shipping both stan files, and then putting a try/except in the model compilation in `setup.py` that catches on `distutils.errors.CompileError`, and in that case falls back to using the this non-matrix version. @rolando  is this something you could add to this PR and check that it works in a Windows install?

@seanjtaylor if we can get it to work shipping both, I'd favor immediately releasing a v0.1.1 on Pip to get this working for Windows now. Then hopefully if in the future PyStan gets matrix multiplication working and it will immediately start using the matrix version for new windows installs.

rolando:
@bletham yes, that's doable. It might be better to check the platform on `setup.py` and non-matrix version when building on windows, this because it takes a while to reach to the failure and also catching `CompileError` could hide other errors.

rolando:
I got successful builds for windows in all python versions (conda-based): https://ci.appveyor.com/project/rolando/staged-recipes/build/1.0.117

In linux I needed to apply #98 to fix the py27 build, and got all success: https://circleci.com/gh/rolando-contrib/staged-recipes/118

bletham:
This is perfect except the Unix versions should be switched back to the matrix code.

rolando:
@bletham Oh, I didn't copied the original files into the unix directory. Fixed the changes and test builds are OK:

https://ci.appveyor.com/project/rolando/staged-recipes/build/1.0.118 
https://circleci.com/gh/rolando-contrib/staged-recipes/126

bletham:
This is wonderful, thanks so much!

PR #100, opened 2017-03-10
Allow to build models in-place.

rolando:
This fixes the commands ``python setup.py develop`` and ``pip install
-e``.


bletham:
Awesome, thanks!

PR #105, opened 2017-03-13
Create weekly seasonality only if user wants to or when data is aggregated weekly

datafool:
changed `self.weekly_seasonality='auto`' in the constructor and if value of this is auto we check for data aggregation, if data is aggregated weekly then we create weekly seasonality else we don't

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -127,14 +127,20 @@ class Prophet(object):
 
     def setup_dataframe(self, df, initialize_scales=False):
         """Create auxillary columns 't', 't_ix', 'y_scaled', and 'cap_scaled'.
+        If minimum difference between two dates of data is more than 7, then 
+        we don't create weekly seasonality
 
         These columns are used during both fitting and prediction.
         """
         if 'y' in df:
             df['y'] = pd.to_numeric(df['y'])
         df['ds'] = pd.to_datetime(df['ds'])
-
+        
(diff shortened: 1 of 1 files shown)


bletham:
This is definitely along the lines of what I was thinking. I'd like to make this a little bit more general though since we'll want a similar thing for yearly seasonality. Here's what would be ideal:
1. I'd like a separate method `set_auto_seasonalities(self)` that has this logic for setting weekly seasonality if 'auto', and then also would be able to set yearly_seasonality if 'auto'. (For that, I'm thinking set to True if the history spans > 1.5 years).
2. This method should be called only during `fit`, not inside `setup_dataframe` which is used also for predictions. It should go right after the line `self.history = history`. That way the method can use `self.history` to set the auto seasonalities.
3. We will also need unit tests for this feature. We should test 4 conditions, one where the true outcome is T/F for each of weekly and yearly seasonalities.
4. This will need to be implemented in the same place/way in the R version and merged together at the same time to keep consistency between the two versions.

bletham:
If your background is primarily Python, I could do this in the R version once this is ready to merge.

One other thing, if you look at line 138 in your PR you will see that it added a line with 4 blank spaces. You can configure your editor to strip spaces from blank lines, which is what we'd need.

bletham:
This is now done by https://github.com/facebookincubator/prophet/commit/d937f47612c54107aa6adb52fa2891ec03bda3eb

PR #137, opened 2017-03-28
[docs][bug] Remove semi-colon from Quick Start page

nkconnor:


bletham:
The documentation is autogenerated from Jupyter notebooks. You can see the notebook for the Quickstart here:
https://github.com/facebookincubator/prophet/blob/master/notebooks/quick_start.ipynb
Any documentation changes thus need to be made in the notebook first, and then if you cd into `docs` and run the command `make notebooks` it will regenerate the md file with the changes.

The reason for the semi-colons being there is because by default Jupyter will print the output of whatever the last line in each block is, which clutters up the documentation. The semi-colon suppresses that print in Jupyter.

nkconnor:
got it, thanks!

PR #148, opened 2017-04-04
Minor typos and spellchecks in quickstart docs

cdharris:
Removed semicolon from python sources and fixed a grammar typo.

### DIFF
--- a/docs/_docs/quick_start.md
+++ b/docs/_docs/quick_start.md
@@ -70,12 +70,12 @@ df.head()
 
 
 
-We fit the model by instantiated a new `Prophet` object.  Any settings to the forecasting procedure are passed into the constructor.  Then you call its `fit` method and pass in the historical dataframe. Fitting should take 1-5 seconds.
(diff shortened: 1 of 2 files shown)


cdharris:
Contributor License Agreement signed :)

bletham:
Thanks for the grammar fix!

The documentation file is autogenerated from the notebook by cd'ing into `docs` and running `make notebooks`. The reason for the semi-colons being there is because by default Jupyter will print the return of whatever the last line in each block is if unassigned, which clutters up the documentation. If you run `make notebooks` then you should see a bunch of extra output in the generated md file. The semi-colon suppresses that print in Jupyter.

It seems that this may seem odd to people who haven't used Jupyter, but I don't know of a good alternative for the plotting blocks, where we want to suppress the return but want to keep the figure. One alternative is to assign the output, `artists = m.plot_components(fcst)` but I'm not a fan of that since it may make it seem necessary to assign the output, which you would generally have no need to do.

bletham:
I pushed the grammar fixes in https://github.com/facebookincubator/prophet/commit/03d3ec9f2edd2f246ad2c072188150320bbc58ce but left the semi-colons in for now.

PR #156, opened 2017-04-06
ADD kwargs param for matplotlib figure

lsternlicht:


bletham:
I'm wondering if it might be a bit more general to actually update `plot` to use the same logic as `plot_trend`, `plot_holidays`, and `plot_weekly`, where the user can actually just pass in the matplotlib axes that the plot should be made on. And if None, then we create a figure with default settings as is currently done. This would allow users to do things like change the figure size or even put it inside a subplot. Would this work for whatever your use case is for wanting kwargs?

seanjtaylor:
Thanks for the contribution @lsternlicht but I tend to agree with @bletham here.  I think @lemonlaug contributed the original matplotlib axes code, maybe he has an opinion?

lemonlaug:
I'm in agreement with @bletham as well, layout and plotting should be separate concerns though with an option for a default. This is consistent with pandas and matplotlib's recommendations FWIW.

If you make that change, you could make another change that brings us closer to consistency with the the `plot_components` family of methods by returning mpl `Artist` objects instead of the figure.

bletham:
This was done in https://github.com/facebookincubator/prophet/commit/26ca2f7af7a4dad2606ef1071342fd6d966310a5

PR #160, opened 2017-04-08
Set up Travis to run the python tests.

mbruggmann:
Hey

We would really like to start using fbprophet from iPython and Cloud Datalab notebooks. At the moment installing the python version requires a C++ compiler to compile the pystan models though. We could avoid that by building and releasing wheel packages to PyPI instead. 

I am proposing to use [Travis](https://travis-ci.org) to continuously build and test the Python fbprophet package. On top of that, we could easily [automate uploading wheels to PyPI](https://docs.travis-ci.com/user/deployment/pypi/) for every release. If this sounds like a good idea I'm happy to provide an additional PR for that.


seanjtaylor:
This seems great @mbruggmann, can you sign your CLA so we can marge the pull request?

mbruggmann:
@seanjtaylor This should be good to go now.

bletham:
Thanks for the update!

PR #180, opened 2017-05-03
Fix typo in comment

jkabrg:
`n.changepoints` changed to `n_changepoints`.
The first follows R's naming convention but is
incorrect Python.


bletham:
Thanks!

PR #184, opened 2017-05-09
Highlight active item in the main menu

DanilBaibak:
Highlight active item in the main menu according to the using page


bletham:
Thanks!

PR #197, opened 2017-05-19
Set verbosity for Printing Meta Information

### LINKED ISSUE #188: Set verbosity for Printing Meta Information
`print` statements such as one in [forecaster.py](forecaster.py)  line number - 380 and 381 should either write to stderr instead of stdout or such there should an option to turn off these messages.

When such a statement is executed in a distributed environment such as hive with a custom script using `fbprophet`, output includes such rougue statements as well.

```python
print('Disabling yearly seasonality. Run prophet with '
                      'yearly_seasonality=True to override this.')
```

Hasil-Sharma:
Changed `print` statements to `logger.info` and `logger.error`. 

This redirects such messages from stdout to stderr and doesn't mess up systems that depend on reading from stdout for output. 

Fixes #188

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -383,7 +383,7 @@ class Prophet(object):
             if last - first < pd.Timedelta(days=730):
                 self.yearly_seasonality = False
                 logger.info('Disabling yearly seasonality. Run prophet with '
-                      'yearly_seasonality=True to override this.')
+                      		'yearly_seasonality=True to override this.')
             else:
                 self.yearly_seasonality = True
         if self.weekly_seasonality == 'auto':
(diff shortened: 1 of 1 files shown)


python/fbprophet/forecaster.py
```diff
@@ -387,7 +392,7 @@ def set_auto_seasonalities(self):
             if ((last - first < pd.Timedelta(weeks=2)) or
                     (min_dt >= pd.Timedelta(weeks=1))):
                 self.weekly_seasonality = False
-                print('Disabling weekly seasonality. Run prophet with '
+                logger.into('Disabling weekly seasonality. Run prophet with '
```
bletham: typo, `info`

python/fbprophet/forecaster.py
```diff
@@ -377,7 +382,7 @@ def set_auto_seasonalities(self):
         if self.yearly_seasonality == 'auto':
             if last - first < pd.Timedelta(days=730):
                 self.yearly_seasonality = False
-                print('Disabling yearly seasonality. Run prophet with '
+                logger.info('Disabling yearly seasonality. Run prophet with '
                       'yearly_seasonality=True to override this.')
```
bletham: this should be indented to be inline with the 'Disabling...'

python/fbprophet/forecaster.py
```diff
@@ -387,7 +392,7 @@ def set_auto_seasonalities(self):
             if ((last - first < pd.Timedelta(weeks=2)) or
                     (min_dt >= pd.Timedelta(weeks=1))):
                 self.weekly_seasonality = False
-                print('Disabling weekly seasonality. Run prophet with '
+                logger.into('Disabling weekly seasonality. Run prophet with '
                       'weekly_seasonality=True to override this.')
```
bletham: indent

Issue #210, opened 2017-06-01
Monthly Seasonality: How to incorporate ?


vaibhavkaushal:
Hi,
My time series data has weekly as well as monthly seasonality i.e. sales tends to be higher on weekends and again same is applied for month end also. However my time series doesn't have yearly seasonality. How do i incorporate these changes ? Please explain.

bletham:
We'd like to add a method for adding arbitrary seasonalities, something like:
```
m.add_seasonality(period=30.4, components=5, name='monthly')
```
You can see right here where we can easily generate the corresponding X matrix for a specified period and number of components: https://github.com/facebookincubator/prophet/blob/master/R/R/prophet.R#L387

It shouldn't be too much effort to add this functionality.

BrashEQLibrium:
I feel compelled to write an endorsement for this enhancement, which would make forecast immensely more valuable for sales forecasting. I could definitely convince my company to use prophet in much of our sales analytics if this feature were added. Do you need any help with developing it given your other priorities?

bletham:
I would definitely accept a PR for this :-)

fanlu:
@bletham When to release a new version？

bletham:
@fanlu I don't have a good timeline for the next version atm.

1mike12:
Along the same note, I would like to see hourly seasonality, It's the strongest in my data compared to day of week or month

1mike12:
I took a horrible stab at trying to add it but unfortunately I don't know what I'm doing for python plotting. If someone who actually knows what they're doing, I think my PR for hourly is really close. https://github.com/facebookincubator/prophet/pull/235

bletham:
This is now pushed to the v0.2 branch. https://github.com/facebookincubator/prophet/commit/8be35c2f34a41b653e3966bb6a5391aefc6554f8 and https://github.com/facebookincubator/prophet/commit/707c88527593bd65f02e99e03704bf84545495d1 add a method/function `add_seasonality` for adding a custom seasonality. See the example here for using monthly seasonality:
https://github.com/facebookincubator/prophet/blob/v0.2/notebooks/seasonality_and_holiday_effects.ipynb

Install the v0.2 branch from github and try it out!
In R you can use
```
devtools::install_github('facebookincubator/prophet', subdir='R', ref='v0.2')
```

fanlu:
@bletham How to use python version?

bletham:
[...]

PR #235, opened 2017-07-01
Attempt to add hourly seasonality

1mike12:
I know this is very bad practice, but the code does not work at this point. But it's very close and probably trivial for someone who actually knows python could fix my plot_hourly method,

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -924,7 +924,7 @@ class Prophet(object):
         return fig
 
     def plot_components(self, fcst, uncertainty=True, plot_cap=True,
-                        weekly_start=0, yearly_start=0):
+                        hourly_start=0, weekly_start=0, yearly_start=0):
         """Plot the Prophet forecast components.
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the patch. I just merged into branch v0.2 a PR that adds daily seasonality. There's also been a request for monthly seasonality (#210), and we decided for that we will keep the built-in seasonalities to just yearly/weekly/daily, and other seasonalities (monthly, and also hourly here) will be handled through a method to add custom seasonalities. If you're interested in working on custom seasonalities we can discuss the API more in #210, but in the meantime we've decided against having a built-in hourly seasonality component.

bletham:
https://github.com/facebookincubator/prophet/commit/8be35c2f34a41b653e3966bb6a5391aefc6554f8 and https://github.com/facebookincubator/prophet/commit/707c88527593bd65f02e99e03704bf84545495d1 add the custom seasonality option, which I think should also handle the monthly seasonality case.

Issue #201, opened 2017-05-22
Failed building wheel for fbprophet


sonic3366:
pip install fbprophet  
Collecting fbprophet
  Using cached fbprophet-0.1.1.tar.gz
Requirement already satisfied: matplotlib in /root/anaconda2/lib/python2.7/site-packages (from fbprophet)
Requirement already satisfied: numpy in /root/anaconda2/lib/python2.7/site-packages (from fbprophet)
Requirement already satisfied: pandas>=0.18.1 in /root/anaconda2/lib/python2.7/site-packages (from fbprophet)
Requirement already satisfied: pystan>=2.14 in /root/anaconda2/lib/python2.7/site-packages (from fbprophet)
Requirement already satisfied: six>=1.10 in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: python-dateutil in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: functools32 in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: subprocess32 in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: pytz in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: cycler>=0.10 in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: pyparsing!=2.0.4,!=2.1.2,!=2.1.6,>=1.5.6 in /root/anaconda2/lib/python2.7/site-packages (from matplotlib->fbprophet)
Building wheels for collected packages: fbprophet
  Running setup.py bdist_wheel for fbprophet ... error
  Complete output from command /root/anaconda2/bin/python -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-8Yl5PE/fbprophet/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmpICUNmGpip-wheel- --python-tag cp27:
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib
  creating build/lib/fbprophet
  creating build/lib/fbprophet/stan_models
[...]

PR #260, opened 2017-07-24
Anaconda users should use conda

lincolnfrias:
Just add the instruction given in the issue #201 to the README and to the docs.


bletham:
This is already specified in the docs, see the bottom of https://facebookincubator.github.io/prophet/docs/installation.html. I do think that adding this to the README is good though.

lincolnfrias:
Ops, I stopped at the pip. My bad, @bletham. I'm closing this then. Thanks for the awesome package!

PR #261, opened 2017-07-28
Implement cross-validation of time series(a rolling forecast origin)

teramonagi:
I made a functionality of cross-validation of time series based on [this conversation](https://github.com/facebookincubator/prophet/issues/257).

The concept of cross-validation of time series is explained in detail in the following link:

- https://robjhyndman.com/hyndsight/tscv/

If this looks good, I will add more tests and implement it in R language as well.
I'm not sure how I name this function and would like to change it through discussion.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -32,6 +32,7 @@ def _copy(model):
     Prophet class object with the same parameter with model variable
     """
 
+    # TODO(@teramonagi) this function should be moved into Prophet class as a method.
     result = Prophet(
         growth=model.growth,
         n_changepoints=model.n_changepoints,
@@ -46,6 +47,42 @@ def _copy(model):
         uncertainty_samples=model.uncertainty_samples
     )
     return result
+
+
+def _cutoffs(df, horizon, k, period):
+    """Generate cutoff dates
+
+    Parameters
+    ----------
+    df: pd.DataFrame with historical data
+    horizon: pd.Timedelta.
+        Forecast horizon
+    k: Int number.
+        The number of forecasts point.
+    period: pd.Timedelta.
+        Simulated Forecast will be done at every this period.
+
+    Returns
+    -------
+    list of pd.Timestamp
+    """
+    # Allocate memory for result list
+    result = [None] * k
+    # Last cutoff is 'latest date in data - horizon' date
+    cutoff = df['ds'].max() - horizon
+    result[0] = cutoff
+
+    for i in range(1, k):
+        cutoff -= period
+        # If data does not exist in data range (cutoff, cutoff + horizon]
+        if not (((df['ds'] > cutoff) & (df['ds'] <= cutoff + horizon)).any()):
+            # Next cutoff point is 'closest date before cutoff in data - horizon'
+            closest_date = df[df['ds'] <= cutoff].max()['ds']
+            cutoff = closest_date - horizon
+        result[i] = cutoff
+
+    # Sort lines in ascending order
+    return reversed(result)
 
 
 def simulated_historical_forecasts(model, horizon, k, period=None):
@@ -72,7 +109,7 @@ def simulated_historical_forecasts(model, horizon, k, period=None):
     df = model.history.copy().reset_index(drop=True)
     horizon = pd.Timedelta(horizon)
     period = 0.5 * horizon if period is None else pd.Timedelta(period)
-    cutoffs = [df['ds'].max() - horizon - period * i for i in reversed(range(k))]
+    cutoffs = _cutoffs(df, horizon, k, period)
     predicts = []
     for cutoff in cutoffs:
         # Generate new object with copying fitting options
@@ -80,7 +117,7 @@ def simulated_historical_forecasts(model, horizon, k, period=None):
         # Train model
         m.fit(df[df['ds'] <= cutoff])
         # Calculate yhat
-        index_predicted = (df['ds'] >= cutoff) & (df['ds'] <= cutoff + horizon)
+        index_predicted = (df['ds'] > cutoff) & (df['ds'] <= cutoff + horizon)
         yhat = m.predict(df[index_predicted][['ds']])
         # Merge yhat(predicts), y(df, original data) and cutoff
         predicts.append(pd.concat([
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -14,57 +14,99 @@ import logging
 
 logger = logging.getLogger(__name__)
 
+import numpy as np
 import pandas as pd
 from functools import reduce
-
 from fbprophet import Prophet
 
+def _copy(model):
+    result = Prophet(
+        growth=model.growth,
+        n_changepoints=model.n_changepoints,
+        yearly_seasonality=model.yearly_seasonality,
+        weekly_seasonality=model.weekly_seasonality,
+        holidays=model.holidays,
+        seasonality_prior_scale=model.seasonality_prior_scale,
+        changepoint_prior_scale=model.changepoint_prior_scale,
+        holidays_prior_scale=model.holidays_prior_scale,
+        mcmc_samples=model.mcmc_samples,
+        interval_width=model.interval_width,
+        uncertainty_samples=model.uncertainty_samples
+    )
+    return result
 
-def cv(model, periods, horizon=1):
-    """Computes the forecast errors obtained by applying predict function to subsets of
-    the time series y using a rolling forecast origin.
+
+def _cutoff_points(te, horizon, period, k):
+    return [te - horizon - period * i for i in reversed(range(k))]
+
+
+def shf(model, horizon, k, period=None):
+    """Simulated Historical Forecasts.
 
     Parameters
     ----------
-    model: Prophet class object
-    periods: Int number of test periods of a rolling forecast origin.
-    horizon: Forecast horizon
+    model: Prophet class object. Fitted Prophet model
+    horizon: string which has pd.Timedelta compatible style.
+        Forecast horizon
+    k: Int number. The number of forecasts point.
+    period: string which has pd.Timedelta compatible style or None, default None
+        Simulated Forecast will be done at every this period.
+        0.5 * horizon is used when it is None.
 
     Returns
     -------
-    A pd.DataFrame with the forecast errors (error = y - yhat)
+    A pd.DataFrame with the forecast, actual value and cutoff.
     """
+    pd.to_timedelta
     df = model.history.copy().reset_index(drop=True)
-    size_history = len(df)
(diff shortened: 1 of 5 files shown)


teramonagi:
Let me know if I should implement the metrics mentioned in https://github.com/facebookincubator/prophet/issues/194 .

Ic3fr0g:
@teramonagi Please do! It's a great feature to have!

teramonagi:
@Ic3fr0g 
I think that we should implement it in a different branch and as a different class because cross-validation and its measure are not the same thing esentially.
How you think about my idea?

Ic3fr0g:
Agreed. Would you like me to create a PR ?

bletham:
This is great, thanks for the PR! A few comments:

- I think this would be better as a separate function as opposed to a method of Prophet, and in a separate file. Let's have a file `diagnostics.py` or something like that, and then this function would be `cv(m, ...)` where `m` is a fitted Prophet model.

- The current approach for a dataset of length N is to start at N - horizon - periods, and then one observation at a time move forward in the time series, and make a forecast for horizon periods. There are a couple of things here that I think would be better if done slightly differently.

First, it'd be nice to not have to require regular spacing of the time series for this to work. The horizon could be given as a unit of time (5 days, 10 hours, etc.) and then we would construct the future dataframe to hold whatever observations there were in the history within that time window. In the final output we would convert those datetimes to be time deltas from the cutoff (end of the training data). The `horizon` input instead of being an int could be a string in the form expected by pd.Timedelta. So something like '5 days' or '10 minutes' could be passed directly in.

The second thing is about how we choose those cutoffs. This PR implements what Rob Hyndman recommends which is to use a cutoff at every observation. This is completely reasonable, but at the same time the prediction errors from one cutoff to the next are very correlated: typically adding one observation to the training data will not change the forecast very much. This just means that there is not a whole lot of value to making cutoffs one observation apart, and we can get a pretty-much-as-good estimation of prediction error by spacing out the cutoffs a bit. This of course is desirable because it means fewer models to be trained and so CV is much faster. I'd like this to be the default strategy, with an ability to use cutoffs every day should the user wish. The interface could be something like
```
cv(m, horizon, periods=10, initial=None)
```
where `initial` would be a pd.Timedelta-style string that indicates the minimum amount of time to be used for training (that is, the distance in time from the first observation in the history to the first cutoff). If not provided (None) we could default to something reasonable like 3 horizons. So if t0 is the start of the time series and T the end, the first cutoff would be at the first observation after t0+initial. The last would be at the last observation before T-horizon. And we would uniformly space the remaining periods-2 cutoffs in between those two. If periods is more than the number of observations inside [t0+initial, T-horizon], then we will have a cutoff at each observation as in the current implementation. Thoughts?

- Please rebase this on the v0.2 branch, that is where the new features are going in preparation for that launch. And please don't change the version number, we'll increment that when the release is ready.

python/fbprophet/forecaster.py
```diff
@@ -36,7 +36,6 @@
 # fb-block 2
 
 
-
```
teramonagi: I got a message `PEP8: too many blank lines` from IDE(PyCharm).

teramonagi:
I understood the situation and realized that it is the same thing `3.3 Simulated Historical Forecasts` in [the prophet paper](https://facebookincubator.github.io/prophet/static/prophet_paper_20170113.pdf).

Is my understanding correct?

teramonagi:
I understood the situation and realized that it is a special version of [`3.3 Simulated Historical Forecast`](https://facebookincubator.github.io/prophet/static/prophet_paper_20170113.pdf).

Then, I will implement Simulated Historical Forecasts ( `shfs` ) at first.
After that, I will implement time-series cross validation as a special case of that.

teramonagi:
@bletham 
I implemented `shf` function (Simulated Forecast) based on [the paper of Prophet](
https://facebookincubator.github.io/prophet/static/prophet_paper_20170113.pdf) following your advice.

And I also re-newed my `cv` implementation using `shf`.
It is completely the same except for these arguments.

Does it sound good for you?
I would like to write a more document and test if it is.

python/fbprophet/diagnostics.py
```diff
@@ -0,0 +1,112 @@
...
+    cutoffs = _cutoff_points(df['ds'].max(), horizon, period, k)
+
+    predicts = []
+    for cutoff in cutoffs:
+        # Generate new object with copying fitting options
+        m = _copy(model)
+        # Train model
+        m.fit(df[df['ds'] <= cutoff])
+        # Calculate yhat
+        index_predicted = (df['ds'] >= cutoff) & (df['ds'] <= cutoff + horizon)
```
bletham: Should be `df['ds'] > cutoff` so that the cutoff date isn't in both training and validation data.

python/fbprophet/diagnostics.py
```diff
@@ -0,0 +1,112 @@
...
+        yhat = m.predict(df[index_predicted][['ds']])
+        # Merge yhat(predicts), y(df, original data) and cutoff
+        predicts.append(pd.concat([
+            yhat,
+            df[index_predicted][['y']].reset_index(drop=True),
+            pd.DataFrame({'cutoff': [cutoff] * len(yhat)})
+        ], axis=1))
+
+    # Combine all predicted pd.DataFrame into one pd.DataFrame
+    return reduce(lambda x, y: x.append(y), predicts).reset_index(drop=True)
```
bletham: Since there is no ground truth for the components, I think it should be safe to limit the output columns here to 'ds', 'cutoff', 'y', 'yhat', 'yhat_lower',and 'yhat_upper'.

python/fbprophet/diagnostics.py
teramonagi: It seems that [the original Prophet's paper](https://facebookincubator.github.io/prophet/static/prophet_paper_20170113.pdf) 
allows this case.
You can see it in Figure 5 and Table 4.

bletham: In the paper at the cutoff we were not using the cutoff date in the history (so strict inequality in line 71 here). Definitely either L71 or L73 should be strict, depending on if we want "cutoff" to mean data up to and including that date, or if we want it to mean up to but not including that date. I personally think it is more natural to have the cutoff date included in the training as is currently implemented in L71 (and so not in the validation). We are about to update the paper and I can make sure it is consistent there. If you think it is more natural to have the cutoff date not included in the history (and so in the validation) I'm open to that also.

python/setup.py
```diff
@@ -117,7 +118,7 @@ def with_project_on_sys_path(self, func):
         'develop': DevelopCommand,
         'test': TestCommand,
     },
-    test_suite='fbprophet.tests.test_prophet',
+    test_suite='fbprophet.tests',
```
teramonagi: Change this to run entire test.

teramonagi:
@bletham 
I update codes to reflect your comment and add tests & documentations.
In addition to that, I change the name of functions ( `shf` -> `simulated_historical_forecasts`, `cv` 
 -> `cross_validation`  ) to be more clear for users.
Could you merge this PR if it looks good to you ? :)

I will implement the same functionality in R language if necessary.

python/fbprophet/diagnostics.py
```diff
@@ -0,0 +1,159 @@
...
+
+    Returns
+    -------
+    Prophet class object with the same parameter with model variable
+    """
+
+    # TODO(@teramonagi) this function should be moved into Prophet class as a method.
+    result = Prophet(
+        growth=model.growth,
+        n_changepoints=model.n_changepoints,
```
bletham: There are two interfaces for changepoints: The `n_changepoints` input that is copied over here, but there is also the `changepoints` input, which allows the user to manually specify the locations of any potential changepoints. If `changepoints` is not None, then `n_changepoints` is ignored.

The correct thing to do here is to copy over both, but one subtlety is that we require the user-specified changepoints to fall within the history (see `Prophet.validate_inputs`). We thus need to filter `model.changepoints` to be just those that are <cutoff (if `model.changepoints` is not None). This should be straightforward by passing `cutoff` along to `_copy`. Making it an optional input would make sense if it is a method of Prophet.

python/fbprophet/diagnostics.py
```diff
@@ -0,0 +1,159 @@
...
+    result[0] = cutoff
+
+    for i in range(1, k):
+        cutoff -= period
+        # If data does not exist in data range (cutoff, cutoff + horizon]
+        if not (((df['ds'] > cutoff) & (df['ds'] <= cutoff + horizon)).any()):
+            # Next cutoff point is 'closest date before cutoff in data - horizon'
+            closest_date = df[df['ds'] <= cutoff].max()['ds']
+            cutoff = closest_date - horizon
+        result[i] = cutoff
```
bletham: This is a good place to handle the case where k * period is more than the length of the time series. Instead of pre-allocating `result`, just initialize it as `result = [cutoff]` and then here in L82:
```
if cutoff < df['ds'].min():
    logger.warning('Not enough data for requested number of cutoffs! Using {}.'.format(k))
    break
result.append(cutoff)
```

python/fbprophet/diagnostics.py
```diff
@@ -0,0 +1,159 @@
...
+    cutoffs = _cutoffs(df, horizon, k, period)
+    predicts = []
+    for cutoff in cutoffs:
+        # Generate new object with copying fitting options
+        m = _copy(model)
+        # Train model
+        m.fit(df[df['ds'] <= cutoff])
+        # Calculate yhat
+        index_predicted = (df['ds'] > cutoff) & (df['ds'] <= cutoff + horizon)
+        yhat = m.predict(df[index_predicted][['ds']])
```
bletham: if `model.growth == 'logistic'`, we need to include `df['cap']` here also or it will error.

python/fbprophet/tests/test_diagnostics.py
```diff
@@ -0,0 +1,76 @@
...
+                self.assertAlmostEqual(np.sum((df_merged['y_x'] - df_merged['y_y']) ** 2), 0.0)
+
+    def test_simulated_historical_forecasts_default_value_check(self):
+        m = Prophet()
+        m.fit(self.__df)
+        # Default value of period should be equal to 0.5 * horizon
+        df_shf1 = diagnostics.simulated_historical_forecasts(m, horizon='10 days', k=1)
+        df_shf2 = diagnostics.simulated_historical_forecasts(m, horizon='10 days', k=1, period='5 days')
+        print(df_shf1)
+        print(df_shf2)
```
bletham: remove prints

teramonagi:
@bletham 
Change codes to reflect your comments and add some tests for `Prophet.copy` method!

bletham:
This is huge, thanks!

If you'd like to translate this commit over to the R version that'd be great, otherwise I'll get to it in the next couple weeks.

PR #270, opened 2017-08-07
Add a function to pick practical change points (R)

hoxo-m:
Hello.
I suggest a function to pick practical change points from prophet model object.
Because prophet sparse estimates changes in the growth rates, we usually get change points contains nearly zero changes.

```r
> df <- read.csv("https://raw.githubusercontent.com/facebookincubator/prophet/master/examples/example_wp_peyton_manning.csv")
> df$y <- log(df$y)
> 
> library(prophet)
> m <- prophet(df, n.changepoints = 15)
> m$changepoints
 [1] "2008-05-18" "2008-11-13" "2009-04-17" "2009-09-18" "2010-03-05"
 [6] "2010-08-14" "2011-01-16" "2011-06-20" "2011-11-24" "2012-04-29"
[11] "2012-10-02" "2013-03-05" "2013-08-08" "2014-01-11" "2014-06-15"
> as.vector(m$params$delta)
 [1] -6.967408e-06  8.049281e-01 -4.754921e-08 -2.404752e-01 -2.007219e-01
 [6] -1.844382e-07  2.658313e-01  1.094812e-01  9.937361e-08 -8.180386e-01
[11]  1.398169e-06  5.498220e-01  2.451281e-07 -3.389944e-01 -1.024679e-07
```
It is more clear using `round` function: 
```r
> round(as.vector(m$params$delta), digits = 2)
 [1]  0.00  0.80  0.00 -0.24 -0.20
 [6]  0.00  0.27  0.11  0.00 -0.82
[11]  0.00  0.55  0.00 -0.34  0.00
```

Practically, we want to get only non-zero changes.
So I implemented `prophet_pick_changepoints` function:

```r
> prophet_pick_changepoints(m, digits = 2)
  changepoints growth_rate      delta
1   2007-12-10 -0.37995357  0.0000000
2   2008-11-13  0.42496753  0.8049211
3   2009-09-18  0.18449230 -0.2404752
4   2010-03-05 -0.01622959 -0.2007219
5   2011-01-16  0.24960149  0.2658311
6   2011-06-20  0.35908272  0.1094812
7   2012-04-29 -0.45895581 -0.8180385
8   2013-03-05  0.09086762  0.5498234
9   2014-01-11 -0.24812665 -0.3389943
```

How about it?

### DIFF
--- a/R/R/prophet_pick_changepoints.R
+++ b/R/R/prophet_pick_changepoints.R
@@ -1,0 +1,40 @@
+#' Pick changepoints from prophet object
+#'
+#' @param model Prophet model object.
+#' @param digits Integer, indicating the number of decimal places to be used. Default 2.
+#'
+#' @return A data frame consists of changepoints, growth rates and delta (changes in the growth rates).
+#'
+#' @examples
+#' \dontrun{
+#' m <- prophet(df)
+#' prophet_pick_changepoints(m)
+#' }
+#'
+#' @export
+prophet_pick_changepoints <- function(model, digits = 2) {
+  df <- data.frame(cp = model$changepoints,
+                   cp.t = model$changepoints.t,
+                   delta = as.vector(model$params$delta))
+  while(nrow(df) > 1 && any(abs(df$delta) < 10^-digits)) {
+    ind <- which.min(abs(df$delta))
+    if (ind == 1) {
+      pos <- 2
+    } else if (ind == nrow(df) || 2 * df$cp.t[ind] < df$cp.t[ind-1] + df$cp.t[ind+1]) {
+      pos <- ind - 1
+    } else {
+      pos <- ind + 1
+    }
+    df$delta[pos] <- df$delta[pos] + df$delta[ind]
+    df <- df[-ind, , drop = FALSE]
+  }
+  if (nrow(df) == 1 && abs(df$delta) < 10^-digits) {
+    df <- data.frame()
+  }
+  cp <- c(model$start, df$cp)
+  delta <- c(0, df$delta)
+  growth_rate <- model$params$k + cumsum(delta)
+  result <- data.frame(changepoints = cp, growth_rate = growth_rate, delta = delta)
+  class(result) <- c("prophet_changepoint", class(result))
+  result
+}


hoxo-m:
Also, we can draw a plot of change points using it.

```r
> fore <- predict(m, df)
> cpts <- prophet_pick_changepoints(m)

> plot(m, fore) + geom_line(aes(ds, trend), color="red") +
>   geom_vline(xintercept = as.integer(cpts$changepoints[-1]), color="red", linetype="dashed")
```

change_plot

bletham:
Thanks for the PR!

This function takes any changepoints whose delta is <0.01 and basically folds them into a neighboring changepoint, by adding its delta to that of the neighbor, and then removing it.

There are a couple aspects of this procedure that make it seem to me like it will not be generally useful. 

The first is that adding the values of the deltas doesn't give an accurate representation of the delta for that whole segment. Let's say we have two changepoints at t1=1 and t2=2, with delta1=2 and delta2=1. Suppose y=0 and k=0, then at t=3, we have y=5. Now, if we follow this procedure for removing delta2 it would leave us with a single changepoint at t1=1 with delta1=3. Then at t=3, y=6, not 5. Obviously a correct combining of delta2 into delta2 would be 2.5, but the math becomes more complicated with the logistic growth trend.

The second issue is that the parameters `m` are computed using `delta` to make sure that the trend is piecewise continuous. Updating the deltas but not correctly computing the m's means that we couldn't actually use these deltas to make predictions, and so I'm not sure they'd be generally useful enough to include in the package. So unless I'm misunderstanding some aspect of this, I'm going to not merge this PR.

There is value in identifying the "significant" deltas for the purpose of visualization, but that could be done with a simple thresholding and there is no need for this combining of deltas logic. I'd accept a PR that produced the changepoints plot above, but had a threshold on delta for which ones are marked with the vertical lines. It could go in a utils file and be called something like `plot_changepoints`.

hoxo-m:
Thank you for your comment.
I realized you are right.

The reason why I implemented the function to add the values of the deltas is for such case as follows:
```
library(prophet)
library(ggplot2)

set.seed(314)
trend <- c(rep(1, 50), rep(-1, 50))
bias <- c(rep(0, 50), rep(100, 50))
t <- 1:100
y <- trend * t + rnorm(100, bias, 3)

df <- data.frame(ds = as.Date("2017-01-01") + t - 1, y)
m <- prophet(df, yearly.seasonality = FALSE)
fore <- predict(m, df)

cpts1 <- m$changepoints[abs(m$params$delta) >= 10^-1]
plot(m, fore) + geom_vline(xintercept=as.integer(cpts1), color="red", linetype="dashed")
```

rplot

We sometimes want to thin out the change points.

```
cpts2<- m$changepoints[abs(m$params$delta) >= 10^0]
plot(m, fore) + geom_vline(xintercept=as.integer(cpts2), color="red", linetype="dashed")
```

rplot01

In this case, we lost all of them.
It seems the function that I implemented has done well.

```
cpts3 <- prophet_pick_changepoints(m, digits = 0)
plot(m, fore) + geom_vline(xintercept=as.integer(cpts3$changepoints[-1]), color="red", linetype="dashed")
```
rplot02

However, I think that simple way is OK as follows:
```
cpts4 <- m$changepoints[abs(m$params$delta) >= 0.9]
plot(m, fore) + geom_vline(xintercept=as.integer(cpts4), color="red", linetype="dashed")
```
rplot03

Sorry to consume your time.
I will try to remake PR to concentrate at visualization following your advice.

Issue #179, opened 2017-05-03
Feature Request: Matplotlib should be an optional dependency


joincamp:
A little background might help explain why this would be a nice feature.

From the Prophet whitepaper:
> It is worth noting that we are ignoring the typical considerations that scale
implies: computation and storage. **We have found the computational and infrastructure
problems of forecasting a large number of time series to be relatively
straightforward** – typically these fitting procedures parallelize quite easily and
forecasts are not difficult to store in relational databases.

(emphasis mine)

Which totally makes sense, and in my very narrow, but arguably common use-case, I'm on AWS and Lambda is the clear path for me. Lambda is amazing for scaling little things up and getting a lot of compute done without worrying so much about things like having to pay for servers when you aren't using them.

So here is the rub... The dependency chain in this project is like an iceberg. And I get it, but many of the libraries just aren't even used in this type of use case, where you are just dealing with the dataframes and you don't need the plots. The base package size for all the dependencies is >700mb. In order to run on lambda I have to get that down to 250mb uncompressed, and I have been able to do this by monkey patching pystan and fbprophet. I'd rather not have to maintain a separate fork(of this project at least) and would love if having matplotlib as an optional dependency could make it into the mainline. I have no idea if there is an elegant way of having a smaller pystan, but I'll take on that burden).

If I were to find a good way of doing this, would you consider merging it in to master?

joincamp:
I have also filed an issue with stan-dev/math that should help out with overall distributable size if they are amenable. https://github.com/stan-dev/math/issues/545

joincamp:
Based on https://github.com/stan-dev/math/issues/545#issuecomment-299038368, it looks like after the stan models have been pickled, we can actually create a deployable package with a lot of the bulk deleted.

bletham:
I'm amenable to making matplotlib optional for those that do not need the plotting. I would though want to be sure that when users install with pip it does install/upgrade matplotlib by default. It could be confusing for someone to install Prophet and then not understand why the plotting isn't working.

bletham:
If someone would like to make a PR for this, we can just put the matplotlib import in a try/except and disable the plotting methods if it did not import. matplotlib should remain a dependency in pip though, so it will be installed by default with a normal pip install.

jpcaruana:
or maybe we could make another package, say `fbprophet-plot`, for those who want to plot stuff

robertdknight:
[...]

PR #271, opened 2017-08-11
Remove matplotlib dependancy in the main package

jpcaruana:
Following the discussion on issue #179, here is a small **proposal** creating 2 python packages, `fbprophet` and `fbprophet-plot`.

If you want to plot, import `ProphetPlot` from `fbprophet-plot.forecaster` instead of the regular import.

there is also a static method for creating a `ProphetPlot` instance from an existing `Prophet` instance : `plot_instance = ProphetPlot.fromProphet(some_instance)`

ps : I don't know much about python packaging


bletham:
I'm going to close this per discussion in #272 and #179.

PR #272, opened 2017-08-11
python: proposition: import matplotlib only within plot functions

jpcaruana:
python: proposition: import matplotlib only within plot functions

another way of dealing with the issue #179 (see PR #271 for another proposition): we keep only one package in the end and people who don't want to plot won't be bothered by this I suppose  

(still not a pro with the python setup tools by the way)

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -16,10 +16,6 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-from matplotlib import pyplot as plt
-from matplotlib.dates import MonthLocator, num2date
-from matplotlib.ticker import FuncFormatter
-
 import numpy as np
 import pandas as pd
 
@@ -892,6 +888,10 @@ class Prophet(object):
         -------
         A matplotlib figure.
         """
+        from matplotlib import pyplot as plt
+        from matplotlib.dates import MonthLocator, num2date
+        from matplotlib.ticker import FuncFormatter
+
         if ax is None:
             fig = plt.figure(facecolor='w', figsize=(10, 6))
             ax = fig.add_subplot(111)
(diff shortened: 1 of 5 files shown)


bletham:
Moving the imports inside the plot functions is a good solution. It looks like `MonthLocator`, `num2date`, and `FuncFormatter` are only for used for `plot_yearly` so those imports can be removed from the other functions. Also, please put matplotlib back in requirements.txt and I'll merge!

bletham:
and install_requires

python/fbprophet/forecaster.py
```diff
@@ -892,6 +888,10 @@ def plot(self, fcst, ax=None, uncertainty=True, plot_cap=True, xlabel='ds',
         -------
         A matplotlib figure.
         """
+        from matplotlib import pyplot as plt
+        from matplotlib.dates import MonthLocator, num2date
+        from matplotlib.ticker import FuncFormatter
+
```
msoedov: We can generalize it with one single function. This suppose to work with both py2 and py3
```python
def import_matplotlib():
    from matplotlib import pyplot as plt
    from matplotlib.dates import MonthLocator, num2date
    from matplotlib.ticker import FuncFormatter
    global plt, MonthLocator, num2date, FuncFormatter # escape local function scope

```

ajbouh:
Is there any chance this will be merged?

ajbouh:
What can I do?

bletham:
I'm going to close this since it is stale and needs to be re-based anyway.

The hard requirement on our end is that `pip install fbprophet` installs matplotlib. This means not changing requirements.txt. 

My preferred solution would be to just wrap the matplotlib imports in a try/except:
```
try:
    from matplotlib import pyplot as plt
    from matplotlib.dates import MonthLocator, num2date
    from matplotlib.ticker import FuncFormatter
except ImportError:
    logger.error('Unable to import matplotlib. Plotting will not work.')
```
And that's all.
But we can discuss this further in #179 to see if that works for those who need this feature.

PR #273, opened 2017-08-12
Add function to draw significant changepoints

hoxo-m:
As a continuation of the pull request (#270),  I created a function to overlay significant change points on prophet forecast plot.
The `layer_changepoints()` returns ggplot2 layer object, and you can use as follows:

```
plot(m, fcst) + layer_changepoints(m)
```

I think the way is more flexible than like `plot_changepoint`.
For example, if the package will have a mechanism to detect outliers, we can create similar function.

```
plot(m, fcst) + layer_outliers(m)
```

Then we can draw that combination easy.

```
plot(m, fcst) + layer_changepoints(m) + layer_outliers(m)
```

The function name is merely my idea.
I checked up the manner of ggplot2 for such cases, but I couldn't find.
Only `autolayer()` is a method for a similar situation.
http://ggplot2.tidyverse.org/reference/autolayer.html
The function can use from next version of ggplot2.

Changepoints where `abs(delta) >= threshold` are significant, and you can change threshold (default `0.01`).
You can specify line color (default `"red"`) and line type (default `"dashed"`).
And you can choose whether to draw the trend line.

### DIFF
--- a/utils
+++ b/utils
@@ -1,0 +1,32 @@
+#' Get layers to overlay significant changepoints on prophet forecast plot.
+#'
+#' @param m Prophet model object.
+#' @param threshold Numeric, changepoints where abs(delta) >= threshold are significant. (Default 0.01)
+#' @param cp_color Character, line color. (Default "red")
+#' @param cp_linetype Character or integer, line type. (Default "dashed")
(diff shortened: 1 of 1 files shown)


bletham:
This is cool, thanks!

PR #278, opened 2017-08-21
External regressors in R

sukwkim:
@bletham  Hi Ben, I am translating your python code of adding external regressor to R. 
But I am stuck on line 984.  I am little confused that do I need to set components as list or dataframe?
I think I need to use list.  But, I need your advice.


bletham:
Very exciting, this is looking great! For your question, In line 978 `components` is created as a dataframe with columns `col` (column number of `seasonal.features` matrix) and `component` (the name of the component that that column represents). So for instance `components` might look something like
```
col | component
---------------
1   | weekly
2   | weekly
3   | yearly
4   | yearly
5   | yearly
6   | christmas
7   | christmas
8   | new_years
9   | new_years
```
We then add in additional components that are sums of other components. So for the `holidays` component we append a dataframe
```
col | component
---------------
6   | holidays
7   | holidays
8   | holidays
9   | holidays
```

This dataframe is created by `add_group_component`. We would run
```
components <- add_group_component(m, components, 'holidays', unique(m$holidays$holiday))
```
and the result would be the combined dataframe (`components` as it was input, plus the new rows for `holidays`):
```
col | component
---------------
1   | weekly
2   | weekly
3   | yearly
4   | yearly
5   | yearly
6   | christmas
7   | christmas
8   | new_years
9   | new_years
6   | holidays
7   | holidays
8   | holidays
9   | holidays
```
Does that make sense?

Also, I know this is a pain, but could you change this PR to be on the v0.2 branch (not master) and rebase on that?

bletham:
@sukwkim, we're going to be pushing out v0.2 this week and I'm really looking forward to having this as part of it. If you're planning to finish this up in the next couple days let me know, otherwise I'll just merge it in as-is and will take it from there on Wednesday. This was a really big commit and what you've done with it already is fantastic.

sukwkim:
@bletham  I will try my best to finish it by Tuesday. So far I implemented most of them except plot part (but still got some errors when I tried to build them). Today I will change this PR to be on the v0.2 branch.

PR #287, opened 2017-09-01
Update DESCRIPTION

DerekYves:
This should perhaps be updated so one can be clear from session info that they actually running the development branch.


bletham:
Thanks, this should have been done a long time ago but anyway now we have actually pushed v0.2 to CRAN and so it is updated.

PR #290, opened 2017-09-05
Add some dataset

Goffic:


bletham:
The datasets that we include with the package are all used for documentation examples. Unless there is something about this datasets that makes them interesting documentation examples I'm inclined to not merge this.

bbrewington:
You may want to check out the pull request I just submitted, #348 

Fixed an issue with the link to the CSV file in the R package getting started Vignette - maybe this is a type of dataset the OP is referring to?

PR #299, opened 2017-09-13
Adding changepoint threshold

### LINKED ISSUE #217: Feature Request: Expose Potential Changepoint 80% Threshold?
I think it would make sense to expose a class variable to allow the user to change the 80% threshold for potential automatic changepoints. For long time series, the 20% runway may be longer than is necessary. 

Thanks,
Scott

zippeurfou:
This should fix #217 
Maybe we should have a v0.3 branch?

### DIFF
--- a/R/R/prophet.R
+++ b/R/R/prophet.R
@@ -28,7 +28,9 @@ globalVariables(c(
 #' @param n.changepoints Number of potential changepoints to include. Not used
 #'  if input `changepoints` is supplied. If `changepoints` is not supplied,
 #'  then n.changepoints potential changepoints are selected uniformly from the
-#'  first 80 percent of df$ds.
+#'  first `changepoint.threshold` percent of df$ds.
+#' @param changepoint.threshold Parameter controling where to select the changepoints.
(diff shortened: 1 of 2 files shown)


bletham:
Thanks for this PR, and sorry for the long delay on the merge.

The name `changepoint.threshold` makes me think of a threshold on the value of the changepoint. I wonder if a name like `changepoint.range` might be a bit clearer, what do you think?

Right now there is logic that sets `changepoint.threshold` to 0.8 if it falls outside of [0, 1]. I'd prefer to move the check for [0, 1] inside the `validate_inputs` function, and then throw an error rather than automatically switching to default.

Related to that, the docstring says "first `changepoint.threshold` percent of df$ds." which might make it seem like it should be 80 to get 80% (rather than 0.8). I think using "proportion" instead of "percent" would be less likely to cause confusion.

bletham:
Not merged, needs the validation change described. Updating would be great.

bletham:
I'm going to go ahead and merge this and make another commit to update the validation.

PR #308, opened 2017-09-18
efficiency and robustness improvements for R package

bfgray3:
increase efficiency and robustness of the R package by doing the following, along with other small changes:

* minimize the number of function calls
* favor `*apply` over `for` loops when possible
* favor `seq_len` and `seq_along` over `1:*`
* favor lower-level functions over higher-level abstracted functions

### DIFF
--- a/R/DESCRIPTION
+++ b/R/DESCRIPTION
@@ -22,7 +22,8 @@ Imports:
     rstan (>= 2.14.0),
     scales,
     stats,
-    tidyr (>= 0.6.1)
+    tidyr (>= 0.6.1),
+    utils
 Suggests:
     knitr,
     testthat,
--- a/R/R/diagnostics.R
+++ b/R/R/diagnostics.R
@@ -79,10 +79,10 @@ simulated_historical_forecasts <- function(model, horizon, units, k,
     # Copy the model
     m <- prophet_copy(model, cutoff)
     # Train model
-    history.c <- df[df[["ds"]] <= cutoff, , drop = FALSE]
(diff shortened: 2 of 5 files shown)


bletham:
Thanks for this PR, and sorry for the delay on the merge.
Overall this looks fine, although there is a bug in `get_changepoint_matrix` that is causing it to return an empty matrix. If you run the unit tests (`Rscript testthat.R` in R/tests) you will see that they are failing because of this.

Generally we try to keep the R version and the Python version as closely in sync as possible since it makes it much easier to apply updates to both. There is of course then some tension between doing things in the most R or Pythonic way, vs. the way that translates clearly to both. The changes to `make_holiday_features` I'm fine with because anyway there the R and Python versions had to be slightly different.  But I do request that the changes in logic to `get_changepoint_matrix`, `piecewise_linear`, and `piecewise_logistic` be reverted, since here they current match the python version nearly exactly.

Actually, the changes to `piecewise_linear` and `piecewise_logistic` also seem to be substantially slower. On the model fitted to the quickstart data, `piecewise_linear`, 1000 evaluations of the current `piecewise_linear` takes 0.86s, while from the version in this PR it takes 9.8s. So let's definitely keep those as they are.

bfgray3:
Thanks for your understanding.  I will work on this and rectify the issues over the weekend.

bfgray3:
@bletham I have made the requested changes and I see the tests passing on my local machine.  The PR is minimal now, but I hope it is still useful.

bletham:
Great, thanks for the contribution!

PR #320, opened 2017-10-02
Added dygraphs functionality.

wjhrdy:
I added the function dyplot.prophet

and 

extracted all the graphing related funcitons to plot.r


R/R/plot.R
```diff
@@ -0,0 +1,412 @@
...
+    forecastCols <- c('yhat_lower', 'yhat', 'yhat_upper')
+  } else
+  {
+    colsToKeep <- c('y', 'yhat')
+    forecastCols <- c('yhat')
+  }
+  
+  # convert to xts for easier date handling by dygraph
+  # haven't dealt with cap yet
+  dfTS <- xts(dplyr::select_(df, .dots=colsToKeep), order.by=df$ds)
```
bletham: should be xts::xts

R/R/plot.R
```diff
@@ -0,0 +1,412 @@
...
+  # base plot
+  dyBase <- dygraphs::dygraph(dfTS)
+  
+  presAnnotation <- function(dygraph, x, text) {
+    dygraph %>%
+      dyAnnotation(x, text, text, attachAtBottom = TRUE)
+  }
+  
+  dyBase <- dyBase %>% 
+    dyOptions(colors = RColorBrewer::brewer.pal(3, "Set1")) %>%
```
bletham: dygraphs::dyOptions

R/R/plot.R
```diff
@@ -0,0 +1,412 @@
...
+    # plot forecast and ribbon
+    dygraphs::dySeries(forecastCols, label=forecast.label) %>% 
+    # allow zooming
+    dygraphs::dyRangeSelector() %>% 
+    # make unzoom button
+    dygraphs::dyUnzoom() %>% 
+  if (!is.null(x$holidays)) {
+    for (i in 1:nrow(x$holidays)) {
+      # make a gray line
+      dyBase <- dyBase %>% dyEvent(x$holidays$ds[i],color = "rgb(200,200,200)", strokePattern = "solid")
```
bletham: dygraphs::dyEvent

R/R/plot.R
```diff
@@ -0,0 +1,412 @@
...
+      dyBase <- dyBase %>% dyAnnotation(x$holidays$ds[i], x$holidays$holiday[i], x$holidays$holiday[i], attachAtBottom = TRUE)
```
bletham: dygraphs::dyAnnotation

R/R/plot.R
```diff
@@ -0,0 +1,412 @@
...
+  
+  # convert to xts for easier date handling by dygraph
+  # haven't dealt with cap yet
+  dfTS <- xts(dplyr::select_(df, .dots=colsToKeep), order.by=df$ds)
+  # base plot
+  dyBase <- dygraphs::dygraph(dfTS)
+  
+  presAnnotation <- function(dygraph, x, text) {
+    dygraph %>%
+      dyAnnotation(x, text, text, attachAtBottom = TRUE)
```
bletham: dygraphs::dyAnnotation

bletham:
This is great, I look forward to trying it out.

It looks like the `prophet_copy` function ended up over in `plot.R`, could you move it back?
I patched this and try to run dyplot.prophet on the quickstart example, and got
```
Error in if (.) !is.null(x$holidays) else { : 
  argument is not interpretable as logical
```

Since xts and dygraphs are both used only for this one non-core function, I think I'd prefer to have them as optional dependencies in "Suggests:" and then when dyplot.prophet is called, check for both of them with `require()` and raise an error that they should be installed to use that functionality. If dygraphs turns out to be the more popular choice for plotting we'll get a components plot and make it a required dependency.

jaredlander:
@bletham the `xts` argument can be skipped altogether by simply arranging the data in a proper `data.frame`. `dygraphs` takes _either_ and `xts` object __or__ a properly formatted `data.frame`.

Assuming other functions are built using `dygraphs` there would be more of a need to have `dygraphs` as an imports.

bletham:
I definitely agree that with more functions built on dygraphs it should be an import. If the intention is to get there then I suppose we can import it now.

wjhrdy:
@jaredlander I've been looking into how to import the df into dygraph without the use of the xts function. dygraphs effectively relies on xts package to see if the input is xtsible. 

After wasting a bunch of time trying to do this without the xts package I've decided to just use the xts package.

Here are the options I've tried:

1. using tseries::irts function. but this relies on the tseries package.
2. using timetk::tk_xts to export df to xts. but this relies on the timetk package.

jaredlander:
@wjhrdy `dygraphs` can be used with a `data.frame` or `tbl` if they're set up correctly, but it is easier and nicer with `xts`. Out of the packages you mentioned `xts` would be the cleanest. I can check about the conversion when I'm near a computer.

wjhrdy:
@jaredlander I think one thing that is needed for proper input of a data.frame or tbl is to have dates as rownames but I can't figure out the proper configuration. here is a sample dataset that is xtsible but I dont know how to make a df that follow this format.

bletham:
beautiful, thank you!

jaredlander:
Looks like this got merged in, that's awesome. Did you end up using `xts` or a `data.frame` setup to `dygraphs`' liking?

bletham:
It's using xts.

Issue #334, opened 2017-10-17
ValueError when sigma=0


sinanonur:
In `sample_model` function noise is calculated as:
`noise = np.random.normal(0, sigma, df.shape[0]) * self.y_scale`
this results in a `ValueError` because of `numpy` which is natural.

But I think this shouldn't prevent the forecasting. noise could simply be zero in that case.

Thank you

sinanonur:
apparently this is fixed in newer versions of `numpy` this results in the expected zeros.
I don't know if this should be changed or `numpy` dependency should be updated.

bletham:
Thanks, like you said this was fixed in later versions of numpy but is an issue in 1.10, which is our minimum requirement. I ran into this a few weeks ago and fixed it in the github version here:  https://github.com/facebook/prophet/commit/58173861b1f975fe272432d600fe18c79d76250c . So installing the latest github version will fix the issue, or of course upgrading numpy is the better fix if possible.

bletham:
This issue is fixed in the current PyPI version for the constant-history case. If it arises in a different context we can re-open this.

PR #335, opened 2017-10-17
Prevent value error when sigma=0 (Issue #334)

sinanonur:
Replace the `noise` with zeros in `sample_model` when sigma=0 which normally results in a `ValueError` and prevents forecasting.


bletham:
I really appreciate identifying the issue and pushing a fix for it! I expect that we will only run into this issue when the history is constant, in which case https://github.com/facebook/prophet/commit/58173861b1f975fe272432d600fe18c79d76250c fixes it. If you ran into this issue without a constant history then let me know and we will swap the fix out for this.

bletham:
The fix for the constant history case has been pushed to PyPI. I think that should resolve the issue, so I'm going to close this PR. But if you run into this issue outside of having a constant history we can look into this again.

Issue #340, opened 2017-10-28
Support custom stan models


jlindenger:
We're using the prophet library with some of our own custom stan models, but the way that the current "linear" and "logistic" models are loaded on import and managed in the local scope of the Prophet.fit method makes this rather challenging.

I propose making model a property on the Prophet class that gets set appropriately to "linear", "logistic", or custom during __init__, and tweaking models.py to load the pickled stan models lazily upon request to avoid the memory impact of loading them on import.

I'm happy to put together a PR for this. We've already done most of it to make our own code work.

bletham:
This is a great idea, I'd love to see a more convenient interface for custom models.

The main challenge that I see is handling the additional model-specific code that is not in Stan. In particular,
* Initialization for the optimization ([linear](https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L677) and [logistic](https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L701))
* Evaluation of the trend model ([linear](https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L888) and [logistic](https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L915)).

Perhaps the best thing to do will be to make the model a class that would include these two methods, and the stan code / compiled pickle. The complicated part is that these functions already have slightly different inputs (logistic expects `cap`, linear does not), and the parameters k, m, delta, beta, and sigma_obs are hard-coded throughout. It would be a bit of a refactor to support Stan code that has different parameters. That said, I think even requiring these same parameters would still allow flexibility in things like the likelihood model (#337) and so would be useful. Also see #264 for adding some new models, and the type of things that I would hope would be cleaner with this. I'd be curious to hear about what sort of model changes you have found useful.

bletham:
I think #501 will make this much easier by making the Stan code the only place that you have to worry about modifying the model, we can revisit this once that is finished.

bletham:
In light of #501 not being feasible, I don't expect this will be either. I did add to the documentation that PR #1466 provides a good template for what needs to be done to change the trend (https://facebook.github.io/prophet/docs/additional_topics.html#flat-trend-and-custom-trends) . For now I think the best course will be to leave it to the user to follow that PR if a different type of trend function is desired.

PR #341, opened 2017-10-29
WIP Support custom stan models

jlindenger:
Related to #340. This makes model a parameter on the Prophet object, and lets the user provide their own model at initialization. If no model is specified, it uses that same built-in "linear" and "logistic" models that the library currently provides. It also changes the import process for the built-in models slightly so that they only get loaded into memory if they are needed. They should still only load once within a program.

Note: I realize this needs to be implemented for the R side as well because it includes an API/process change. That still needs to be done, hence the WIP tag.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -23,10 +23,6 @@ from matplotlib.ticker import FuncFormatter
 import numpy as np
 import pandas as pd
 
-# fb-block 1 start
-from fbprophet.models import prophet_stan_models
-# fb-block 1 end
-
 try:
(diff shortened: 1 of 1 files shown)


bletham:
(Just to add some context for anyone else interested in this PR), as discussed in #340 we think there is value in having custom models, but the non-stan model-specific code will have to also be implemented in a modular way.

I'm going to go ahead and close this since as-is it would be applicable only for a very specific set of model changes, and the chances of someone using it wrong and getting bad results is very high. We can discuss further in #340 what the more general solution should look like.

PR #348, opened 2017-11-07
Update quick_start.Rmd

bbrewington:
Change relative link to data.csv to absolute link to file in github repo

### DIFF
--- a/R/vignettes/quick_start.Rmd
+++ b/R/vignettes/quick_start.Rmd
@@ -23,7 +23,7 @@ First we read in the data and create the outcome variable.
 
 ```{r, results= "hide"}
 library(readr)
-df <- read_csv('../tests/testthat/data.csv')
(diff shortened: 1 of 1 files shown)


bbrewington:
Hello Facebook bot, I just signed the agreement

bletham:
Switching it to the github link means that anyone can copy-and-paste this code and it will run, which is great. However, it means that the R package will no longer build unless you are connected to the Internet. Internally we build this on servers that can only access the internet via proxies. I'd imagine there could be others who want to make changes to the code and then build it in an environment without direct Internet access. I also frequently find myself working on Prophet while traveling without Internet, and this would be a major annoyance then.

I'd hate to create an issue with net connection just for the vignette, especially since the online quickstart at https://facebook.github.io/prophet/docs/quick_start.html is more fleshed out anyway, uses a dataset with better illustrative qualities, and includes a link to the CSV. I'm leaning towards keeping this as-is.

Issue #321, opened 2017-10-04
Quality of model


beloteloff:
Do you planing to add model's quality metrics like R^2 adjusted in ARIMA models? Metrics MAPE/MAE is not good evaluate model's accuracy

teramonagi:
You can use `diagnostics` functionality to calculate the metrics like R^2.
https://facebook.github.io/prophet/docs/diagnostics.html

beloteloff:
Of course, but I'm interesting for decision from "box".

bletham:
I think it'd be nice to have a whole collection of various metrics. I for one do like MAPE, but it'd be good to have a bunch built-in. I'm not exactly sure what an adjusted R^2 would look like for this model, we'd have to give it a try. This is being tracked in #194.

bletham:
I'm going to close this and leave the discussion on useful cross-validation metrics to #194.

PR #384, opened 2017-12-18
Implement metrics for time series forecasts

teramonagi:
## Summary
Implement metrics for time series forecasts (only for R now).
It enables prophet users to know how their model fit the original time series data.

## Details
This PR includes the following popular metrics:
- Mean Error (ME)
- Mean Squared Error (MSE)
- Root Mean Square Error (RMSE, 
- Mean Absolute Error (MAE)
- Mean Percentage Error (MPE)
- Mean Absolute Percentage Error (MAPE)

## Related issues
- https://github.com/facebook/prophet/issues/194
- https://github.com/facebook/prophet/issues/321

If this PR looks good to you, let me know.
I will implement python version also :)


bletham:
This is fantastic.

I do think we might benefit from a different interface though. This takes in model, and then returns the metric evaluated on the in-sample fit. I'm really excited for having these metrics available for assessing out-of-sample fit, which would be done using the cross_validation function. Like `predict`, that function returns a dataframe with columns `y` and `yhat`, except the `yhat` estimates are all out-of-sample.

What do you think about `me`, `mse`, etc. all taking in a forecast dataframe? The workflow would then be
```
forecast <- predict(m)
mse(forecast)
```
for insample fit, or for out-of-sample fit
```
forecast <- cross_validation(m, 10, 'days')
mse(forecast)
```
It's an extra line in the in-sample case, but I think the out-of-sample case will be extremely useful.

Related to that, these performance metrics are pretty cheap to compute so it might be nice to have a function that just computes all of them and returns a list. What do you think?

bletham:
(Note, a function to compute all of them would be in addition to having them all individually accessible. Just as a convenience thing).

teramonagi:
I completely agree with you.
It makes sense.

> Related to that, these performance metrics are pretty cheap to compute so it might be nice to have a function that just computes all of them and returns a list. What do you think?

Sounds good!

I will mention to you after these modifications.

teramonagi:
@bletham 

I update my PR that

- Add `all_metrics()` which calculate all metrics  I mentioned above
- All metrics functions (e.g. `me`, `mse` ) take `fcst` (forecast result) as an argument instead of `m` (prophet object)

Now, you can use these functionalities like the following:
```r
df <- read_csv('./tests/testthat/data.csv')
m <- prophet(df)
future <- make_future_dataframe(m, periods = 365)
forecast <- predict(m, future)
all_metrics(forecast)
df.cv <- cross_validation(m, horizon = 100, units = 'days')
all_metrics(df.cv)
```

R/R/metrics.R
```diff
@@ -0,0 +1,141 @@
...
+}
+
+#' @rdname metrics
+#' @export
+all_metrics <- function(fcst)
+{
+  # Define all metrics functions as a character
+  metrics <- c("me", "mse", "rmse", "mae", "mpe", "mape")
+  # Convert character to function and evalate each metrics
+  result <- purrr::map(metrics, ~ rlang::eval_tidy(rlang::sym(.x))(fcst))
```
teramonagi: `rlang::eval_tidy(rlang::sym(.x))` is more *modern* R programming style compared to `eval(parse(text=.x))`

bletham:
Sorry, my comment about how this would be the same for in-sample and out-of-sample was incorrect since, as you saw, `predict()` does not return `y`.

I'm really hesitant to change the default columns returned by `predict` since that will break things for anyone who is stashing forecasts into databases with fixed schemas.

I'd prefer if we left `predict` as is. Leave the metric functions as they are in the latest commit here, so that they directly take the dataframe with `y` and `yhat`. And then to make it more convenient to compute in-sample error metrics we could have a function `all_insample_metrics(m)` that takes in the prophet object and then does what this PR was originally doing: Compute predict on the history, merge with y from m$history, and then run all_metrics() on that. The workflows would then be
```
# In-sample
metrics <- all_insample_metrics(m)

# out-of-sample
forecast <- cross_validation(m, 10, 'days')
metrics <- all_metrics(forecast)
```
How does that seem? Sorry for the incorrect request earlier.

teramonagi:
Looks good!
I think it might be convenient for users to use the same interface.
```
# In-sample
metrics <- all_metrics(m)

# out-of-sample
forecast <- cross_validation(m, 10, 'days')
metrics <- all_metrics(forecast)
```

I already know that we can do that in R.
In python, we will use `isinstance(m, Prophet)` for this purpose.
(I'm not sure that it is a good manner in Python...)

I try this first and would like to hear your opinition!

teramonagi:
@bletham 
I changed the interface of R's function.
Now, you can write like this:
```R
df <- read_csv('./tests/testthat/data.csv')
m <- prophet(df)

# In-sample
all_metrics(m)

#Out-of-sample
df_cv <- cross_validation(m, horizon = 100, units = 'days', initial=400)
all_metrics(df_cv)
```

How about this?
...and, Have a happy new year!

teramonagi:
```
df <- read_csv('./tests/testthat/data.csv')
m <- prophet(df)
future <- make_future_dataframe(m, periods = 365)
forecast <- predict(m, future)
all_metrics(forecast)
df.cv <- cross_validation(m, horizon = 100, units = 'days')
all_metrics(df.cv)
```

teramonagi:
I accidentally closed this PR...

bletham:
Thanks, happy new year :-)

Sorry to keep nit picking at this. We try to avoid having functions that accept/return objects of variable types. (in this case, either a model object or a cross validation dataframe). I do like having one function for computing all metrics, but perhaps we could make the model a required input and then the df_cv an optional input. This would then mimic what is done with the `predict` function, where if a future dataframe is provided we use those dates,  otherwise we just do in-sample predictions. So the function would be `all_metrics(m, df_cv = NULL)` and use would look like

```
df <- read_csv('./tests/testthat/data.csv')
m <- prophet(df)

# In-sample
all_metrics(m)

#Out-of-sample
df_cv <- cross_validation(m, horizon = 100, units = 'days', initial=400)
all_metrics(m, df_cv)
```

teramonagi:
I understood your point! I fixed what you pointed out kindly. 
Now, you can write
```r
m <- prophet(df)
# In-sample 
all_metrics(m=m)
# Out-of-sample
df_cv <- cross_validation(m, horizon = 30, units = 'days', initial=65)
all_metrics(df=df_cv)
# Out-of-sample also with some warning
df_cv <- cross_validation(m, horizon = 30, units = 'days', initial=65)
all_metrics(m=m, df=df_cv)
```

You can also do the same thing for each metrics.
```r
# Example: Mean Absolute Error
mae(m=m)
mae(df=df_cv)
```

bletham:
Beautiful, thanks!

teramonagi:
Python version's PR is [here](https://github.com/facebook/prophet/pull/399).

kashif:
thanks for the PR... may I also ask for MASE? See: https://www.otexts.org/fpp/2/5

teramonagi:
@kashif 
I read the article you posted and realized that we need to do many changes for our interface.
Is it possible to implement it in the future?

PR #405, opened 2018-01-08
Document name argument of plot_seasonality method in Python API

aktech:
ping @bletham


bletham:
Awesome, thanks, and sorry for the delay :-)

PR #399, opened 2017-12-29
Implement Cross Validation metrics for Forecasts in Python

toddbot:
(note I added and closed a PR by accident on the main prophet github...apologies!)
In the spirit of #384 I'd like to modify the diagnostics.py file to produce a summary of fitting metrics so that analysts can quickly optimize their implementation directly from the Prophet interface. However, if this is has already been in done in Python let me know and I'll maybe check out something else!

PR #414, opened 2018-01-14
Error metrics python v2

toddbot:
This is a version of the pull request on https://github.com/facebook/prophet/pull/399. The new base branch is on v0.3. I saw the suggestion from @teramonagi to change interface to match R. @bletham Happy to do that if that also makes sense to you.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -179,4 +179,4 @@ def all_metrics(df):
     mpe = (df['yhat'] - df['y']).div(df['y']).sum()*(1/len(df))
     mape = (df['yhat'] - df['y']).div(df['y']).abs().sum()*(1/len(df))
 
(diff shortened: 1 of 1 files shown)


teramonagi:
I think that we need
- individual metrics function (e.g. `me`, `mae`, `mpe`)
- metrics functions take Simulated Forecast Object as a argument

It might be helpful for you to see the following code: https://github.com/facebook/prophet/pull/384/files#diff-e78e1108bb37c00e8ee57983b5a5987bR110

teramonagi:
I forgot to mention about test.
I wrote unit test in R like [this](https://github.com/facebook/prophet/blob/v0.3/R/tests/testthat/test_metrics.R).
You can add your unit test file to [this directory](https://github.com/facebook/prophet/tree/v0.3/python/fbprophet/tests) in Python.

bletham:
This is good, I'll merge this as-is and I can add the unit tests.

bletham:
Thanks!

Issue #415, opened 2018-01-15
vectorize Stan models


bob-carpenter:
The Stan models 

* `R/inst/stan/prophet_logistic_growth.stan`
* `R/inst/stan/prophet_linear_growth.stan`

have some additional opportunities for vectorization, which will shorten the programs and make them a bit faster.

Are the scales of the parameters predictable in the posterior?  If so, it will help to rescale so that adaptation is faster and less error prone (it initializes using unit scales everywhere).

For example, if `k` is expected to take on values like `10` as indicated by `normal(0, 5)`, then it would speed up adaptation if 
```
  k ~ normal(0, 5);
```
was recoded as
```
  k_raw ~ normal(0, 1);
  real k = 5 * k_raw;
```
If this is just meant to be a weakly informative prior and you expect the posterior `k` to usually be in the `(-2, 2)` range, then it shouldn't be changed as it's already on unit scale.

I'll submit a pull request with the vectorization, but will need some guidance w.r.t. posterior scales and how you carry out testing.

PR #416, opened 2018-01-15
[WIP] vectorize linear/logistic growth Stan models

bob-carpenter:
This is the PR for issue #415 

I updated the Stan models for vectorization, which makes them a bit shorter and should make them a bit faster.

I'd like to update for prior scale if the posteriors aren't all roughly unit scale in expectation.  That can make a big difference in speeding up adaptation.  

I also need to know how to test that I didn't break the models in refactoring.  The log densities shouldn't change (though might not be the same to 16 decimal points due to vagaries of floating point non-associativity).

Code copyright owner:  Columbia University
License:  BSD-3 Clause

Columbia has given us a permanent exemption to release Stan-related code open source in perpetuity.

### DIFF
--- a/R/inst/stan/prophet_linear_growth.stan
+++ b/R/inst/stan/prophet_linear_growth.stan
@@ -1,40 +1,35 @@
 data {
-  int T;                                // Sample size
-  int<lower=1> K;                       // Number of seasonal vectors
-  vector[T] t;                            // Day
-  vector[T] y;                            // Time-series
-  int S;                                // Number of changepoints
-  matrix[T, S] A;                   // Split indicators
-  real t_change[S];                 // Index of changepoints
-  matrix[T,K] X;                // season vectors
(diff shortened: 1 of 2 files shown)


bob-carpenter:
That license is too scary for me.  Good luck.

bletham:
Ugh. Well thanks for trying to contribute, and generally for all your support of Stan. I wonder if there is something on our end that could make this easier, (@seanjtaylor ?) but this is the sort of thing I expect them to be sticklers about.

bob-carpenter:
Sorry about that.  If I'd seen the contract first, I'd have stopped after creating the issue.

When someone asks me to sign a long contract to make a volunteer contribution and I don't immediately understand all the language, I bail rather than getting legal advice.  I have the same reaction to NDAs before giving invited talks.

I can totally understand why you're doing this.  Facebook has deep enough pockets and gazillions of contributors, so you're totally exposed on the IP front.

PR #439, opened 2018-02-11
Feature Request: Add changepoint_range resolve #217

### LINKED ISSUE #217: Feature Request: Expose Potential Changepoint 80% Threshold?

MIsmailKhan:
Provide necessary changes to pull request "Adding changepoint threshold #299"

* Rename changepoint_threshold to changepoint_range 
* Modified changepoint_range Docstring 
* Make changepoint_ range validation in validate_inputs function.

### DIFF
--- a/R/R/prophet.R
+++ b/R/R/prophet.R
@@ -28,7 +28,9 @@ globalVariables(c(
 #' @param n.changepoints Number of potential changepoints to include. Not used
 #'  if input `changepoints` is supplied. If `changepoints` is not supplied,
 #'  then n.changepoints potential changepoints are selected uniformly from the
-#'  first 80 percent of df$ds.
+#'  first `changepoint.range` proportion((0,1]) of df$ds.
+#'  @param changepoint.range Parameter controlling where to select the changepoints.
+#'  Not used if input `changepoints` is supplied.
 #' @param yearly.seasonality Fit yearly seasonality. Can be 'auto', TRUE,
 #'  FALSE, or a number of Fourier terms to generate.
 #' @param weekly.seasonality Fit weekly seasonality. Can be 'auto', TRUE,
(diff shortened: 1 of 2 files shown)


MIsmailKhan:
There is some doubt on the interval for changepoint_range. @bletham has stated it as [0,1] while @zippeurfou's pull request has defined it as (0,1].

Suggestion: In the changepoint_range validation, perhaps it would be better if a warning is thrown instead of an error followed by a re-input from user.

R/R/prophet.R
```diff
@@ -443,9 +447,14 @@ set_changepoints <- function(m) {
...
     }
   } else {
-    # Place potential changepoints evenly through the first 80 pcnt of
+    # Place potential changepoints evenly through the first changepoint.range proportion of
     # the history.
-    hist.size <- floor(nrow(m$history) * .8)
+    if (m$changepoint.range > 1 || m$changepoint.range <= 0){
+      m$changepoint.range <- .8
+      message('changepoint.range greater than 1 or less than equal to 0. Using ',m$changepoint.range)
+    }
```
bletham: This block should be replaced by a check inside `validate_inputs` that raises an error if not valid, like in the Python version in this commit.

bletham:
Sorry for the delay on this review too. I'll merge this in asap once these remaining things are resolved.

bletham:
I went ahead and merged #299 and made the necessary updates in https://github.com/facebook/prophet/commit/cb0b47994b1d2d6612f75eff4b9d4bec6f5ad846 so this feature is good-to-go now.

Issue #396, opened 2017-12-28
Components plots doubling in Jupyterlab


brylie:
When plotting model components charts in Jupyterlab, they appear twice:

![screenshot from 2017-12-28 11-03-16](https://user-images.githubusercontent.com/17307/34405930-c4752a30-ebbe-11e7-9887-62ba74d980f5.png)

In the notebook, I have set `%matplotlib inline`, so the charts are plotted directly below notebook cells.

sweb:
You can place a ';' behind your statement in order to get the plot only once. I think this is due to the fact that the plot is already rendered in the function.

bletham:
Correct, it is because jupyter will automatically render any matplotlib plots, and will in addition to that print the output of the last line in the cell (in this case the figure, hence the duplication). Semicolon will suppress printing the last line in the cell so you'll get it only once. Alternatively you can assign the output of plot to a variable and it won't print the 2nd time:
```
fig = model.plot_components(forecast)
```

brylie:
OK. It is worth noting that this behavior seems inconsistent when compared with plotting the timeseries data using `model.plot()`. In other words `model.plot()` and `model.plot_components()` would ideally behave the same.

bletham:
that's interesting, for me they do behave the same (both print duplicates without ;, and print one with ;), which is what I would expect since both of these method return a matplotlib fig object. What behavior are you experiencing?

brylie:
In Jupyter **Lab** the model components plot doubles while the model plot prints normally (without the trailing `;` in each case). I am not sure if [JupyterLab](https://github.com/jupyterlab/jupyterlab) is the deciding factor, but that is just what I am using in this case.

bletham:
Oh interesting, I haven't tried JupyterLab. Both functions return matplotlib figures so I can't imagine why JupyterLab would be treating the outputs differently.

bletham:
I'm going to close this since I don't think there is any changes to be made here, at least until JupyterLab usage picks up.

PR #468, opened 2018-03-15
Eliminate semicolons from Python samples

akarve:
In the case of the `plot*` commands the semicolon actually suppresses output in Jupyter

### DIFF
--- a/docs/_docs/quick_start.md
+++ b/docs/_docs/quick_start.md
@@ -75,7 +75,7 @@ We fit the model by instantiating a new `Prophet` object.  Any settings to the f
 ```python
 # Python
 m = Prophet()
-m.fit(df);
+m.fit(df)
 ```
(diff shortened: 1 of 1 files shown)


bletham:
These documentation files are generated automatically from the notebooks in https://github.com/facebook/prophet/tree/master/notebooks , so changes to the documentation have to be made there and then re-built as described in https://facebook.github.io/prophet/docs/contributing.html.

Most of the time notebooks are configured by default with `%matplotlib inline` which automatically prints any generated figures. Without the semicolon, they are then actually printed twice. This is a relatively frequent point of confusion, e.g. #396. Removing the semicolons in the notebooks and then running the make command from above will produce duplicated figures in the documentation.

It's certainly awkward to have the semicolons there, so maybe in the future we should switch to preventing the double printing by assigning the figure output, but one of those two things is needed.

PR #488, opened 2018-04-05
Make matplotlib non mandatory

PierreGe:
Hey guys,

Yesterday, I tried to parallelize the computation of prediction for a thousand time series on Apache Spark using user defined aggregate function that include fbprophet (pandas_udf).

I quickly run in the issue of matplotlib and its dependency that were required but not used. Indeed, it was a pain because while I had the right to load library on Spark executors I did not had the right to install system wide package (Tkinter needed by matplotlib).

So I forked the repo and made the import of matplotlib non blocking on failure (warning when import fail). It worked and computation were lighting fast on Spark for a thousand time series. I believe making the import of matplotlib non blocking on failure is better/useful for productizing fbprophet.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -17,9 +17,13 @@ import logging
 
 logger = logging.getLogger(__name__)
 
-from matplotlib import pyplot as plt
-from matplotlib.dates import MonthLocator, num2date
-from matplotlib.ticker import FuncFormatter
+try:
+    from matplotlib import pyplot as plt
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR!
This is actually ready-to-go in the v0.3 branch, which will be the next version, right here:
https://github.com/facebook/prophet/blob/v0.3/python/fbprophet/forecaster.py#L21-L26

You can follow along the issue for making matplotlib optional in #179

PierreGe:
That's awesome, thanks for the feedback !

PR #504, opened 2018-04-26
Removed semicolons from quick start docs

anshuman23:
Going through the documentation, I felt that the semicolons were very unusual, especially for Python. I then found out that this was the case because the documentation was being generated as markdown files, from Jupyter notebooks. But for users who read the docs who do not know that this is because of Jupyter, this would be considered un-Pythonic.

And the semicolons for the figures can be removed by catching the figures at the time of plotting. I also tried to remove the semicolon for ```m.fit(df)``` by using the ```%%capture``` IPython cell magic command, but then that would show up in the docs as well and would be even more jarring than the semicolon. So I have let the semicolon be for that statement and only handled the plots.

I have only done this for the quick start notebook as of now, but if this is fine, I will do it for all the notebooks/documentation.

Cheers!


anshuman23:
Understood! I've made a PR to the `v0.3` branch. Here it is: #509

PR #509, opened 2018-05-03
Removed semicolons from quick_start docs

anshuman23:
- Follows from #504 
- Shifted branch from `master` to `v0.3`

PR #520, opened 2018-05-16
Docs - removed semicolons from Python, fixed capitalised table headers

alexlouden:
Just a few minor changes - I removed the semicolons from the python code (unnecessary), and overrode the CSS on the table headers so they wouldn't be capitalised. I haven't checked other pages - they might need fixing too. This is what it looks like at the moment (a bit deceptive because Prophet expects it to be lowercase `ds` and `y`):

screen shot 2018-05-16 at 2 09 10 pm
 
Also I just noticed the repo's [contributing guidelines](https://github.com/facebook/prophet/blob/master/docs/CONTRIBUTING.md) are a default template and need editing.

### DIFF
--- a/docs/_docs/quick_start.md
+++ b/docs/_docs/quick_start.md
@@ -34,8 +34,8 @@ df.head()
   <thead>
     <tr style="text-align: right;">
       <th></th>
-      <th>ds</th>
-      <th>y</th>
+      <th style="text-transform: none">ds</th>
+      <th style="text-transform: none">y</th>
     </tr>
   </thead>
   <tbody>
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR. The documentation is actually all generated from ipython notebooks (https://github.com/facebook/prophet/tree/master/notebooks) and the semicolons are used to suppress Jupyter from printing outputs to the screen. For the figures, for instance, by default Jupyter will print out any generated figure but then will also print out the last returned item in the cell if not assigned, so without the semicolon the figure prints twice. This was actually fixed in the v0.3 branch in #509 by assigning the figure outputs.

Thanks for pointing out the CONTRIBUTING.md, I'll have to figure out what's up with that.. at least on the site we did put something (https://facebook.github.io/prophet/docs/contributing.html).

PR #536, opened 2018-05-24
Seasonal Trends that Vary by Arbitrary Covariate

caseyolives:
This commit adds a feature which allows for the seasonal trend(s) to vary by an arbitrary input. The specific use case in mind: modeling multiple time series simultaneously where we believe the over all trend is the same but seasonal trends may vary by an important feature.

To do so, Ive altered the behavior of the `add_regressor` function, which now takes an additional argument, `modify_seasonality`.

```
m <- add_regressor(m, 'new_input', modify_seasonality = TRUE)
```

When the flag is set to `TRUE`, the new regressor is included BOTH in the intercept term AND as an interaction with ALL seasonal trends.

### DIFF
--- a/R/R/prophet.R
+++ b/R/R/prophet.R
@@ -607,7 +607,7 @@ make_holiday_features <- function(m, dates) {
 #' @return  The prophet model with the regressor added.
 #'
 #' @export
-add_regressor <- function(m, name, prior.scale = NULL, standardize = 'auto'){
+add_regressor <- function(m, name, prior.scale = NULL, modify_seasonality = FALSE, standardize = 'auto'){
   if (!is.null(m$history)) {
     stop('Regressors must be added prior to model fitting.')
   }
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR, this is an interesting feature. Do you have an example of the sort of time series this is directed at? I'd love to see a forecast plot that uses this.

bletham:
I see, this is actually closely related to #538 - there it is seasonality that depends on things like weekend vs. weekday, but I think the solution for #538 could be made general enough to support this use-case also.

I really like the simplicity of the interface here and the idea of just directly interacting existing seasonalities with extra regressors. But I think the interface will need to be a bit more general to handle the situations in #538. In particular, we might want to interact only a subset of seasonalities with any given regressor. I wonder if something like
```
m <- add_seasonality_interaction(seasonality='daily', regressor='is_weekend')
```
would be the right direction, where 'daily' is already a seasonality in the model and 'is_weekend' is already a regressor. One challenge here is that the auto seasonalities are not set until model fitting, but we can figure out how to resolve that. We'll also need to think carefully about how multiplicative vs. additive seasonalities/regressors will need to be considered here. (Related to that, all of the new features are going on the v0.3 branch).

Let's discuss further in #538 what the interface should look like for this.

bletham:
I'm going to close this since as discussed above it'd be good to solve the more general problem in #538, and we can discuss what the implementation should look like there.

Issue #541, opened 2018-05-29
prophet_plot_components: make the plotting optional?


dmhv:
Hello,

When deploying a [Shiny](https://shiny.rstudio.com/) application which is using Prophet to a server, the `prophet_plot_components`  function causes an error, as the server does not have a graphical device to render the plot. What I want to get out of this function is just the list of ggplot objects which I can pass into proper Shiny objects and then display to the user. 

Looking at the code, it seems this wouldn't require much changes - just adding an extra function argument that defaults to rendering the plots but allows to bypass it, and wrapping the plotting part at the end in a conditional.

I haven't yet had the experience of contributing to open source projects, but am fairly sure I could manage this :) Do you think this is something that should be implemented? If so, what, if anything, should be done to the Python equivalent of this function, `plot_components`? If anyone has tried adding this to [Dash](https://plot.ly/products/dash/) apps, some comment would be great!

Kind regards,
Dmitrij

PR #547, opened 2018-05-30
R make rendering of plots by method plot_components optional (#541)

dmhv:
See #541.


bletham:
This looks great, except please make this change on the v0.3 branch (https://github.com/facebook/prophet/tree/v0.3) and not the master branch. I expect it'd be easiest for you to make a new PR with this change on that branch. You'll see that the plotting was refactored in that branch to now be in a plot.R file.

Issue #556, opened 2018-06-01
Package is overriding log output


collincburton:
When the fbprophet module is loaded, my log entries don't write to my designated log file. Info level logs print to the console. If you comment out the import Prophet line, this writes to a log as expected.

```
import logging
from fbprophet import Prophet


logging.basicConfig(filename='log_test.log', level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')

logging.debug("test debug message")
logging.info("Info message")
```

collincburton:
This is my first attempt at a bug fix on a package, but I'm going to take a try on it.

bletham:
Thanks! As an FYI, we are about to push out the v0.3 branch, so you'd want to do any investigating there and not on master.

collincburton:
I think I found the issue. It looks like prophet is importing the logging package and using it to raise exceptions, warnings, info statements, etc. We can replace those with the equivalent functions in base python and I think that should fix it. It's been a great learning experience so far.

bletham:
The logging package is the standard way to log in Python, so I'd be very hesitant to not use it. I think the right thing to do here is determine if there is something in fbprophet that needs to be done to enbale changing the default logging config like in your example.

collincburton:
I get that the logging package is the standard way to log, but is it standard practice to put it in the package itself as opposed to letting the user import logging and create their own logs as they want to? If a user has started their log and the fbprophet package raises an exception, warning, etc., the user can record the event as they see fit as opposed to how it is defined within fbprophet. I'm not sure what best practice is in this matter and I'll do a little more research.

Teoretic6:
Found this issue too today, although in different form.

I set up logging only into file, but when I import Prophet, **logging information also starts to log into terminal**, in which I'm rinning my code.

**Sample code**
If you comment out Prophet import in code below, debug lines 'log this' and 'log that' stop to log into terminal and start to log only to log file as intended.

`import logging
from fbprophet import Prophet

if __name__ == '__main__':
    # Setup logging to file
    root_logger = logging.getLogger()
    root_logger.setLevel(logging.DEBUG)
    handler = logging.FileHandler('logfile.log',
                                  'w',
                                  'utf-8')
    handler.setFormatter(logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s'))
    root_logger.addHandler(handler)

    logging.debug('log this')
    logging.debug('log that')`

I think that there is something with logging setup in Prophet, can investigate further.

[...]

PR #569, opened 2018-06-12
Fix to logging

collincburton:
This addresses issue #556 

The problem occurs because the basicConfig method is called in two of the modules. As noted in the logging docs:

> Note It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

[https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library](url)

Incidentally, this same mistake was also made in the pystan package, so even when this fix is implemented, the bug will still occur until pystan fixes their version. I have a fork of pystan with the basicConfig call removed at https://github.com/collincburton/pystan if you want to try it with the fix implemented.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -34,7 +34,7 @@ from fbprophet.plot import (
     plot_seasonality,
 )
 
-logging.basicConfig()
+
 logger = logging.getLogger(__name__)
 warnings.filterwarnings("default", category=DeprecationWarning)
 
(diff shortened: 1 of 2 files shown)


bletham:
This was handled in https://github.com/facebook/prophet/commit/f44285038cafe34611c9fa95ad47dd54f9d91119 the same way it is being handled in pystan.

PR #577, opened 2018-06-15
Explicitly use 64-bit integers in plot functions

khwilson:
If this is installed on a 32-bit system (rare nowadays, but they still exist) then the conversion to `int` from `timedelta64[ns]` in `plot_cross_validation_metric` will fail. This patch explicitly uses an `np.int64` for this conversion to save some people headaches.

Thanks for making this project available.


bletham:
Awesome, thanks for the fix!

PR #613, opened 2018-07-12
Fix broken internal links in hamburger menu

fasouto:
While browsing the docs in mobile I realize that some internal links are broken. For example if you are in [quick-start](https://facebook.github.io/prophet/docs/quick_start.html) and click on 'Docs' (in the hamburger menu) you'll get redirected to https://facebook.github.io/prophet/docs/docs/ and get a 404.

Fixed using the `relative_url` jekyll filter for internal links.


bletham:
Thanks for the fix!

PR #618, opened 2018-07-19
Add hdays to Prophet 

iamxiaodong:
Add hdays to Prophet, which can calculate nationwide official public holidays in 49 countries.
Add make_holidays method to Prophet, which returns a dataframe of 'ds' and 'holiday' with given list of years and list of countries.
Note: hdays is built upon an open source package [holidays](https://github.com/iamxiaodong/python-holidays) on Github.


python/fbprophet/forecaster.py
```diff
@@ -33,6 +33,8 @@
     plot_yearly,
     plot_seasonality,
 )
+import fbprophet.hdays as hdays
+import holidays
```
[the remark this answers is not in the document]
iamxiaodong: There is no R equivalent version, so we may use different approaches for this

melvinev:
This is useful, thanks!

iamxiaodong:
1. Added `append_holidays` to `Prophet` class, then add a boolean flag `calculate_holidays`, meaning that the user wants to calculate holidays with hdays module.
2. Added option to calculate holidays with `calculate_holidays == True` in both `fit` and `predict`.

python/fbprophet/forecaster.py
```diff
@@ -15,8 +15,10 @@
 import logging
 import warnings
 
+import logger as logger
```
bletham: I think this is unused? `logger` is defined in L43

python/fbprophet/forecaster.py
```diff
@@ -33,6 +35,9 @@
     plot_yearly,
     plot_seasonality,
 )
+# import fbprophet.hdays as hdays
+import hdays as hdays
+import holidays
```
bletham: It doesn't look like `holidays` package needs to be imported here since it is imported in `hdays`. If we do need to import it, then it should be under a different name since we have a variable below named `holidays`.

python/fbprophet/forecaster.py
```diff
@@ -123,6 +131,10 @@ def __init__(
         self.weekly_seasonality = weekly_seasonality
         self.daily_seasonality = daily_seasonality
 
+        if (holidays is not None) and (append_holidays is not None):
+            raise ValueError("User cannot set up both holidays and append_holidays!")
```
bletham: Does it seem reasonable to allow users to specify both `holidays` and `append_holidays` (or whatever its final name will be) and then just append the `append_holidays` to any holidays they provided. That would make a really easy way for people to start from the hdays list and add minor customizations that they need for their application.

python/fbprophet/forecaster.py
```diff
@@ -922,6 +943,11 @@ def fit(self, df, **kwargs):
         -------
         The fitted Prophet object.
         """
+        if self.calculate_holidays:
+            # calculate years range
+            year_list = list(set([x.year for x in pd.to_datetime(df['ds'])]))
+            self.holidays = self.make_holidays(year_list=year_list, country_list=self.append_holidays)
+
```
bletham: I think the best place for this is inside the method `make_holiday_features`. At the very start of that function we could do something like
```
all_holidays = self.holidays
if self.append_holidays is not None:
    year_list = list(set([x.year for x in dates]))
    append_holidays_df = self.make_holidays(year_list=year_list, country_list=self.append_holidays)
    all_holidays = pd.concat((all_holidays, append_holidays_df))
```
And then in the code that follows (L460 on this revision) , use `all_holidays` in the place of `self.holidays.

Like we talked about in the meeting, there is some extra work needed there to handle the fact that `append_holidays_df` there might return different sets of holidays depending on the years that are passed in. The `make_holiday_features` method is the place where this issue has to be addressed. Right now, the holidays dataframe is fixed as `self.holidays` and the only variable input is `dates`. These are used to produce a dictionary from holiday name to binary indicators (an element for each date), `expanded_holidays` in the code. `expanded_holidays` looks like
```
defaultdict(<function __main__.<lambda>>,
            {'playoff_delim_+0': array([0., 0., 0., ..., 0., 0., 0.]),
             'playoff_delim_+1': array([0., 0., 0., ..., 0., 0., 0.]),
             'superbowl_delim_+0': array([0., 0., 0., ..., 0., 0., 0.]),
             'superbowl_delim_+1': array([0., 0., 0., ..., 0., 0., 0.])})
```
We then do
```
holiday_features = pd.DataFrame(expanded_holidays)
```
Other code relies critically on the fact that the order of the columns in `holiday_features` will be the same every time we run it. So that is the thing we have to be sure of when adding these computed holidays: The columns and column order of `holiday_features` used in fitting the model must be the same as the columns/column order that we get from this in `predict`, when a different set of `dates` are passed in.

python/fbprophet/forecaster.py
```diff
@@ -1037,6 +1063,11 @@ def predict(self, df=None):
                 raise ValueError('Dataframe has no rows.')
             df = self.setup_dataframe(df.copy())
 
+        if self.calculate_holidays == True:
```
bletham: As described in the comment above, if this is done inside `make_holiday_features` then it doesn't have to be in both fit and predict.

python/fbprophet/forecaster.py
```diff
@@ -134,6 +146,15 @@ def __init__(
...
 
+        if append_holidays is not None:
+            if not (
+                    isinstance(append_holidays, str)
+                    or isinstance(append_holidays, list)
+            ):
+                raise ValueError("append_holidays must be a string \
+                                      or a list of strings")
+                self.calculate_holidays = True
+                self.append_holidays = append_holidays
```
bletham: put `self.append_holidays = append_holidays` outside the loop so that the attribute will exist and be None if there are no append_holidays.

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1505,57 @@ def copy(self, cutoff=None):
             DeprecationWarning,
         )
         return prophet_copy(m=self, cutoff=cutoff)
+
+    def make_holidays(self, year_list, country_list):
```
bletham: This doesn't actually use `self`, so let's go ahead and make it not a method of Prophet (I've been trying to reduce the number of methods on Prophet since it is quite long). It could either go in `hdays` and be imported from there, or we could make a new `utils.py` file that it could go in and be imported from.

python/fbprophet/forecaster.py
```diff
@@ -100,6 +107,7 @@ def __init__(
             weekly_seasonality='auto',
             daily_seasonality='auto',
             holidays=None,
+            append_holidays=None,
```
bletham: Go ahead and add this to the `prophet_copy` function here: https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py#L161

python/fbprophet/forecaster.py
bletham: I think we can ensure `holiday_features` keeps the same columns by tracking the names in `all_holidays['holiday']` when we first run `make_holiday_features`, and then in subsequent runs remove any rows from `all_holidays` whose holiday is not in the initial fitted list.

We can also sort `holiday_features` by column name prior to returning it, to be extra sure that things are staying correctly ordered. We just have to be sure to do the sort before `prior_scale_list` is computed.

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1505,57 @@ def copy(self, cutoff=None):
...
+        year_list: a list of years
+        country_list: a list of countries
+
+        Returns
+        -------
+        Dataframe with 'ds' and 'holiday', which can directly feed
+        to 'holidays' params in Prophet
+        """
+
+        if isinstance(country_list, str):
```
bletham: I think this can be simplified by doing
```
if isinstance(country_list, str):
    country_list = [country_list]
```
and then cutting directly to the code in L1539

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1505,57 @@ def copy(self, cutoff=None):
...
+                        except:
+                            raise AttributeError(
+                                "Holidays in {} are not currently supported!".format(country))
+                    temp_df = pd.DataFrame(list(temp.items()),
+                                           columns=['ds', 'holiday'])
+                    all_hdays.append(temp_df)
+        res = pd.concat(all_hdays, axis=0, ignore_index=True)
+
+        res = res.groupby(['ds']) \
+            .first() \
```
bletham: Is this only allowing one holiday per `ds`? Do we need/want that?

bletham:
This got merged in its final form in #644. Thanks @iamxiaodong for getting it in place!

PR #634, opened 2018-08-02
Fix table header case in quick_start.md

madaan:
In the Python API section, the table headers are being rendered in uppercase (DS, Y) as opposed to lowercase (ds, y). I can confirm that the behavior is consistent across several platforms. I've proposed one sample fix that may help, but there could be other ways of formatting this correctly.

### DIFF
--- a/docs/_docs/quick_start.md
+++ b/docs/_docs/quick_start.md
@@ -8,7 +8,7 @@ permalink: /docs/quick_start.html
 
 Prophet follows the `sklearn` model API.  We create an instance of the `Prophet` class and then call its `fit` and `predict` methods.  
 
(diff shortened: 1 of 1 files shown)


bletham:
This would be great to fix but unfortunately it cannot be fixed by just editing the md file. The markdown files for all of the documentation are actually generated directly from the example notebooks, here: https://github.com/facebook/prophet/blob/master/notebooks/quick_start.ipynb

The code used for generating the markdown is in https://github.com/facebook/prophet/blob/master/docs/Makefile

They are lowercase in the ipynb but uppercase in the md; it must be nbconvert at fault, and we'll have to figure out how to fix it either in this command or in the nbconvert template (https://github.com/facebook/prophet/blob/master/docs/nbconvert_template.tpl)

PR #642, opened 2018-08-09
Adding title of cross-validation metric in plot

IronistM:
Since cross-validation metrics are a handful and could probably be generated in bulk when having a lot of data series, having the metric in the title of the plot generated by `plot_cross_validation_metric` function by default will be a nice side-effect.

```R
prophet_performance$cross_val %>% 
  mutate(cov_plots = map(.,plot_cross_validation_metric,metric = 'coverage'), # this is the existing function
         mape_plots = map(.,plot_cross_validation_metric_custom,metric = 'mape')) # this is the proposed one
```

image

### DIFF
--- a/R/R/plot.R
+++ b/R/R/plot.R
@@ -497,6 +497,7 @@ plot_cross_validation_metric <- function(df_cv, metric, rolling_window=0.1) {
   gg <- (
     ggplot2::ggplot(df_none, ggplot2::aes_string(x = 'x_plt', y = metric)) +
     ggplot2::labs(x = paste0('Horizon (', dts[i], ')'), y = metric) +
+    ggplot2::ggtitle(paste0("Evolution of ", ggplot2::quo_name((metric)), " over horizon"))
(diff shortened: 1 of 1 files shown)


bletham:
I prefer to keep the amount of text on the figures to the minimum required to know what is going on, since it seems many people end up overriding the axes labels anyway to suit their needs (especially to get non-English labels). Here the metric name is given in the ylabel, but I can see there is still some value to adding it as a title if you are generating many of these. That said, it is also very easy to add this inside the loop that is generating them since the plot function returns the ggplot object. So it is only one additional line to do
```
gg <- plot_cross_validation_metric(df.cv, metric = 'mape')
gg <- gg + ggplot2::ggtitle('Title for MAPE')
```

Does that work well enough for your use case? If so then I prefer to not merge this PR, especially since we do not have titles for any of the other plots.

IronistM:
I hear your "minimal" approach to the plot style and can definitely live with it ;)

PR #644, opened 2018-08-13
Modification of holiday features 

ziye666:
Add hdays to Prophet, which can calculate nationwide official public holidays in 49 countries. (work done by @iamxiaodong).
Allow user to choose among self-defining, automatically generate holidays or both way. (holidays != None and append_holidays != None)
Make sure holiday DataFrames used in both fit and predict have the same set of holiday names

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -766,6 +766,7 @@ class Prophet(object):
             raise Exception('A bug occurred in seasonal components.')
         # Compare to the training, if set.
         if self.train_component_cols is not None:
+            component_cols = component_cols[self.train_component_cols.columns]
             if not component_cols.equals(self.train_component_cols):
                 raise Exception('A bug occurred in constructing regressors.')
         return component_cols, modes
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -19,7 +19,7 @@ import pandas as pd
 
 from fbprophet.diagnostics import prophet_copy
 from fbprophet.models import prophet_stan_model
-from fbprophet.make_holidays import make_holidays
+from fbprophet.make_holidays import get_holiday_names, make_holidays_df
 from fbprophet.plot import (
     plot,
     plot_components,
@@ -69,8 +69,7 @@ class Prophet(object):
         lower_window=-2 will include 2 days prior to the date as holidays. Also
         optionally can have a column prior_scale specifying the prior scale for
         that holiday.
-    append_holidays: List of country names or abbreviations; if only one country
-        needed, this can be specified as a String.
+    append_holidays: country name or abbreviation; must be string
     seasonality_mode: 'additive' (default) or 'multiplicative'.
     seasonality_prior_scale: Parameter modulating the strength of the
         seasonality model. Larger values allow the model to fit larger seasonal
@@ -141,10 +140,8 @@ class Prophet(object):
         if append_holidays is not None:
             if not (
                     isinstance(append_holidays, str)
-                    or isinstance(append_holidays, list)
             ):
-                raise ValueError("append_holidays must be a string \
-                                      or a list of strings")
+                raise ValueError("append_holidays must be a string")
         self.append_holidays = append_holidays
 
         self.seasonality_mode = seasonality_mode
@@ -228,6 +225,10 @@ class Prophet(object):
                 name in self.holidays['holiday'].unique()):
             raise ValueError(
                 'Name "{}" already used for a holiday.'.format(name))
+        if (check_holidays and self.append_holidays is not None and
+                name in get_holiday_names(self.append_holidays)):
+            raise ValueError(
+                'Name "{}" is a holiday name in {}.'.format(name, self.append_holidays))
         if check_seasonalities and name in self.seasonalities:
             raise ValueError(
                 'Name "{}" already used for a seasonality.'.format(name))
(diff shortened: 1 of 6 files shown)


python/fbprophet/forecaster.py
```diff
@@ -199,7 +216,7 @@ def validate_column_name(self, name, check_holidays=True,
             raise ValueError('Name cannot contain "_delim_"')
         reserved_names = [
             'trend', 'additive_terms', 'daily', 'weekly', 'yearly',
-            'holidays', 'zeros', 'extra_regressors_additive','yhat',
+            'holidays', 'zeros', 'extra_regressors_additive', 'yhat',
             'extra_regressors_multiplicative', 'multiplicative_terms',
         ]
         rn_l = [n + '_lower' for n in reserved_names]
```
bletham: Down in L231, we should pull out all of the holiday names from the countries in append_holidays to be sure that we don't add seasonalities or extra regressors with names that conflict with holidays that will be added.

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1524,58 @@ def copy(self, cutoff=None):
             DeprecationWarning,
         )
         return prophet_copy(m=self, cutoff=cutoff)
+
+    def make_holidays(self, year_list, country_list):
```
bletham: let's make this not a class method since this file is already way too long. This can be moved to a separate file and just imported here, since it doesn't actually use `self` at all.

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1524,58 @@ def copy(self, cutoff=None):
...
+        year_list: a list of years
+        country_list: a list of countries
+
+        Returns
+        -------
+        Dataframe with 'ds' and 'holiday', which can directly feed
+        to 'holidays' params in Prophet
+        """
+
+        if isinstance(country_list, str):
```
bletham:
```
if isinstance(country_list, str):
    country_list = [country_list]
```
and then go directly to L1558

python/fbprophet/forecaster.py
```diff
@@ -1474,3 +1524,58 @@ def copy(self, cutoff=None):
...
+                        except AttributeError:
+                            raise AttributeError(
+                                "Holidays in {} are not currently supported!".format(country))
+                    temp_df = pd.DataFrame(list(temp.items()),
+                                           columns=['ds', 'holiday'])
+                    all_hdays.append(temp_df)
+        res = pd.concat(all_hdays, axis=0, ignore_index=True)
+
+        res = res.groupby(['ds']) \
+            .first() \
```
bletham: Am I correct in understanding that this `.first()` means that we only allow one holiday per ds? That would not be necessary, and I think not what we want to do.

python/fbprophet/forecaster.py
```diff
@@ -100,6 +105,7 @@ def __init__(
             weekly_seasonality='auto',
             daily_seasonality='auto',
             holidays=None,
+            append_holidays=None,
```
bletham: This also needs to be added to the prophet_copy function here: https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py#L161

python/fbprophet/forecaster.py
```diff
@@ -450,13 +447,22 @@ def make_holiday_features(self, dates):
...
                                     country_list=self.append_holidays)
+            #Validates names of generated holidays.
+            if self.train_holiday_names is None:
+                for name in append_holidays_df.holiday:
+                    if name in self.seasonalities:
+                        raise ValueError(
+                            'Seasonality name "{}" already used for a holiday.'.format(name))
+                    if name in self.extra_regressors:
+                        raise ValueError(
+                            'Regressor name "{}" already used for a holiday.'.format(name))
```
bletham: This block from L453 to L461 needs to be removed. We want to validate names when seasonalities/regressors are added, not during fitting. (At this point it is too late to change the name without creating a whole new model).

What we ideally want is:
* In the make_holidays.py, make a function `get_holiday_names(append_holidays)` that returns a set of all holiday names that could possibly be returned for the holidays in append_holidays. Import that function here.
* In L227 here, you can see where there is an if block for validating that the name doesn't overlap with a holiday name. Add a block right below that one to validate that it doesn't overlap with an append_holidays name, like this:
```
if (check_holidays and self.append_holidays is not None and
        name in get_holiday_names(self.append_holidays
):
    raise ValueError(
        'Name "{}" already used for a built-in holiday.'.format(name)
    )
```

python/fbprophet/make_holidays.py
```diff
@@ -0,0 +1,50 @@
...
+    Returns
+    -------
+    Dataframe with 'ds' and 'holiday', which can directly feed
+    to 'holidays' params in Prophet
+    """
+
+    if isinstance(country_list, str):
+        country_list = [country_list]
+    all_hdays = []
+    for country in country_list:
```
bletham: Question: Are there overlapping holiday names across countries? If so, do they have identical dates?
If there are overlapping holidays across countries, we want to be sure to grab only one copy of them if someone passes in a list of countries.

bletham:
This is great! Merging it in!

python/fbprophet/forecaster.py
```diff
@@ -34,10 +29,18 @@
...
+import holidays as hdays_part2
+
 logging.basicConfig()
 logger = logging.getLogger(__name__)
 warnings.filterwarnings("default", category=DeprecationWarning)
 
+try:
+    import pystan  # noqa F401
+except ImportError:
+    logger.exception('You cannot run fbprophet without pystan installed')
```
orivej: There is no point checking pystan here because if it were not available, `from fbprophet.models` above would fail. This message was meant to anticipate the failure to unpickle the pystan model in `models.py`.

bletham: We needed this for some internal tests which are no longer being used so at this point I think we can just remove the check and raise the ImportError. We could also wrap the unpickling in a try/except if we wanted a more informational message about PyStan not working on that error.

orivej: I'm OK with removing this.

PR #652, opened 2018-08-20
added for mac os

vschs007:
added check for mac os

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -18,7 +18,8 @@ from setuptools.command.test import test as test_command
 PLATFORM = 'unix'
 if platform.platform().startswith('Win'):
     PLATFORM = 'win'
-
+if platform.platform().startswith('Dar'):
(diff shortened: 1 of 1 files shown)


bletham:
That check is because PyStan in Windows does not support matrix multiplication, so we have a separate Stan model for Windows that replaces matrix multiplication with for loops:

https://github.com/facebook/prophet/blob/master/python/stan/win/prophet.stan

This makes the model slower, so we do not want to use this for non-windows platforms. That's why that PLATFORM string defaults to 'unix' and is only changed to 'win' if it is Windows. You can see here where the platform string is used to choose which Stan model to load:
https://github.com/facebook/prophet/blob/master/python/setup.py#L23

There is no "mac"  model (https://github.com/facebook/prophet/tree/master/python/stan, nor does there need to be since PyStan in MacOS supports the default model) so this will actually prevent Prophet from installing correctly.

PR #670, opened 2018-08-29
 Fix bug about holiday year different from input year and others

ziye666:
In hdays.py, output holiday year might differ from input year when there are  different calendars. Fixed this bug.


bletham:
This has all of the original py commits mixed in. Could you re-base on current master? You might have to re-submit the PR.

ziye666:
Hi @bletham, I have rebased and solved conflict problems.

R/R/make_holidays.R
```diff
@@ -0,0 +1,55 @@
...
+  if (country.name == 'India' || country.name == 'IN'){
+    warning.msg = "We only support Asalha Puja holiday from 2006 to 2025 and \
+                    Vassa holiday from 2006 to 2020"
+    warning(warning.msg)
+  }
+  if (max(years) > max.year ||  min(years) < min.year){
+    warning.msg = paste("We only support holidays from year", min.year, 
+                        "to year", max.year)
+    warning(warning.msg)
+  }
```
bletham: I think it's not very ideal to have hard-coded in here the years for which holidays in certain countries are supported, since our holidays csv is auto-generated from the holidays packages. If the holidays packages increase their coverage, then our warning message could be incorrect, and it becomes our job to have to keep on top of what the latest date ranges are from holidays.

I think a better approach is to do something like
```
country_holidays = generated_holidays %>%
    dplyr::filter(country == country.name)
max.year <- max(country_holidays$year)
min.year <- min(country_holidays$year)
if (max(years) > max.year ||  min(years) < min.year){
    warning.msg = paste("Holidays for", country.name, "are only supported from", min.year, 
                        "to", max.year)
    warning(warning.msg)
}
```

This also has the advantage that users could swap out the generated_holidays.csv with their own file, and everything would still work.

R/R/prophet.R
```diff
@@ -177,6 +181,11 @@ validate_inputs <- function(m) {
       validate_column_name(m, h, check_holidays = FALSE)
     }
   }
+  if (!is.null(m$append.holidays)) {
+    if (!m$append.holidays %in% generated_holidays$country){
```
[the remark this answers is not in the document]
seanjtaylor: According to http://r-pkgs.had.co.nz/data.html  if we set `LazyLoad: true` in the package description it will only be loaded at first use. I also don't think this logical check is going to work because of the operator precedence of `!`.

python/scripts/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2017-present, Facebook, Inc.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree. An additional grant
+# of patent rights can be found in the PATENTS file in the same directory.
+
+from generate_holidays_file import generate_holidays_file
+
+generate_holidays_file()
```
seanjtaylor: Can you show me how you run this script from the command line?  I think ideally this is just added to the `generate_holidays_file.py` in a block at the bottom:

```
if __name__ == "__main__":
    # execute only if run as a script
    generate_holidays_file()
```

R/R/prophet.R
```diff
@@ -547,21 +548,23 @@ make_seasonality_features <- function(dates, period, series.order, prefix) {
...
   dates <- set_date(format(dates, "%Y-%m-%d"))
-  all.holidays <- m$holidays %>% 
-    dplyr::mutate(ds=as.character(ds), holiday=as.character(holiday)) 
+  all.holidays <- m$holidays 
   if (!is.null(m$append.holidays)){
     years <- as.numeric(unique(format(dates, "%Y")))
     append.holidays.df <- make_holidays_df(years, m$append.holidays) %>%
       dplyr::mutate(ds=as.character(ds), holiday=as.character(holiday))
-    all.holidays <- all.holidays %>% dplyr::bind_rows(append.holidays.df)
+    all.holidays <- suppressWarnings(dplyr::bind_rows(all.holidays, append.holidays.df))
```
bletham: what warning is being suppressed here?

R/R/prophet.R
```diff
@@ -547,21 +548,23 @@ make_seasonality_features <- function(dates, period, series.order, prefix) {
...
-    all.holidays <- all.holidays %>% dplyr::bind_rows(holidays.to.add)
+  if (!is.null(m$append.holidays) && !is.null(m$train.holiday.names)){
+    row.to.keep <- which(all.holidays$holiday %in% m$train.holiday.names)
+    all.holidays <- all.holidays[row.to.keep,]
+    holidays.to.add <- data.frame(holiday=setdiff(m$train.holiday.names,
+                                                    all.holidays$holiday))
+    all.holidays <- suppressWarnings(dplyr::bind_rows(all.holidays, holidays.to.add))
+  }
+  if (nrow(all.holidays)==0){
+    return(NULL)
```
bletham: I'm not a fan of returning NULL here (I prefer to have a fixed type to the function returns), but more importantly there is no check like this in the current Python function. What made this check necessary, and does it break in the current Python too?

I think the only way this would have no rows is if m$holidays is null, and m$append.holidays is not null, but the country has no holidays for the year in the training data. Is that correct?

R/R/prophet.R
```diff
@@ -616,11 +619,12 @@ make_holiday_features <- function(m, dates) {
   }
   holiday.names <- names(prior.scales.list)
   if (is.null(m$train.holiday.names)){
-    m$train_holiday_names <- holiday.names
+    m$train.holiday.names <- holiday.names
   }
-  return(list(holiday.features = holiday.features,
+  return(list(m = m,
```
bletham: good catch!

R/R/prophet.R
ziye666: Yes, you are right. In the case you described, all.holidays would be NULL, and dplyr cannot operate on NULL.

Issue #401, opened 2018-01-03
Pip install error ModuleNotFoundError: No module named 'pystan'


msoedov:
```shell
pip --version
pip 9.0.1 from /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (python 3.6)
```

```
pip install fbprophet
Collecting fbprophet
  Downloading fbprophet-0.2.1.tar.gz
Requirement already satisfied: matplotlib in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from fbprophet)
Requirement already satisfied: pandas>=0.18.1 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from fbprophet)
Collecting pystan>=2.14 (from fbprophet)
  Downloading pystan-2.17.0.0-cp36-cp36m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl (35.2MB)
    100% |████████████████████████████████| 35.2MB 43kB/s
Requirement already satisfied: python-dateutil>=2.0 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: pyparsing!=2.0.4,!=2.1.2,!=2.1.6,>=2.0.1 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: cycler>=0.10 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: numpy>=1.7.1 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: six>=1.10 in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Requirement already satisfied: pytz in /Users/miasool/pyenvs/xor_ai/lib/python3.6/site-packages (from matplotlib->fbprophet)
Collecting Cython!=0.25.1,>=0.22 (from pystan>=2.14->fbprophet)
  Downloading Cython-0.27.3-cp36-cp36m-macosx_10_6_intel.macosx_10_9_intel.macosx_10_9_x86_64.macosx_10_10_intel.macosx_10_10_x86_64.whl (4.7MB)
[...]

PR #679, opened 2018-09-11
Update setup.py

N3da:
Add `pystan` to setup_requires: 
running the following command on a new virtualenv fails:
`pip install fbprophet --target=/tmp/foo --no-cache-dir`

Error: ImportError: No module named pystan

More context: 
https://stackoverflow.com/questions/52267470/pip-install-options-no-cache-dir-and-target-dont-work-well-together/52275092#52275092

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -106,6 +106,7 @@ setup(
     license='BSD',
     packages=['fbprophet', 'fbprophet.tests'],
     setup_requires=[
+        'pystan'
     ],
     install_requires=install_requires,
     zip_safe=False,


bletham:
There's some discussion of this in #401, see my comment there about the downside of setup_requires. The solution seems to be using a `pyproject.toml` file, but I haven't tested it myself yet.

allisonwang:
I have the exact same issue and is unable to to install fbprophet on any machine. Is it possible to merge this change as a temporary solution?

bletham:
@allisonwang I'm testing this right now. In the meantime, it looks like if you do `pip install fbprophet --no-cache-dir` it will skip building the wheel, and should thus avoid the error.

bletham:
I just merged #685 which fixes this for pip >=10 using a pyproject.toml file, which is the recommended approach by pip.

Issue #643, opened 2018-08-10
Loaded pickle model cannot properly predict new 'future' dataframe


tianchuliang:
I fitted a prophet model (with additional regressors), pickled it, and re-loaded it to predict a future dataframe. However, I am seeing this error: 
A bug occurred in constructing regressors

As I traced the function calls (self.predict -> self.predict_seasonalities -> .. -> self.regressor_column_matrix), the problem is rooted at the final few lines in the "self. regressor_column_matrix" function: 
if self.train_component_cols is not None:
    component_cols = component_cols[self.train_component_cols.columns]
    if not component_cols.equals(self.train_component_cols):
        raise Exception('A bug occurred in constructing regressors.')

I discovered that only a few additional reggressor columns are not equal. Could these additional reggressors cause the problem? 

Thanks,

ispmarin:
I haven't seem any problems on pickling and unpickling using sklearn joblib. Does this only happens with additional regressors?

tianchuliang:
hi @ispmarin , as far as I can tell, it ONLY happens with additional regressors. If I do a simple fbprophet model, no extra aggressors, the pickling wouldn't be a problem.

ispmarin:
Got it. May you try to use joblib and see if the same problem happens? http://scikit-learn.org/stable/modules/model_persistence.html

Gedevan-Aleksizde:
I found similar problem. I didn't encounter this error when I ran the above @bletham 's code. But when I ran the following code, the error was raised (I tried it with Python 3.5). I need to change the regressor names.

```
from fbprophet import Prophet  # 0.3
import pandas as pd  # 0.23.4
from sklearn.externals import joblib  # 0.11
import numpy as np  # 1.15.1

df = pd.DataFrame({
    'ds': pd.date_range('2010-01-01', '2011-01-01'),
    'y': range(366)
})
np.random.seed(42)
extra_variables = [
        'lag_y', 'reg_1', 'reg_2',
        'dummy_00', 'dummy_01', 'dummy_02', 'dummy_03', 'dummy_04',
        'dummy_05', 'dummy_06', 'dummy_07', 'dummy_08', 'dummy_09',
        'dummy_10', 'dummy_11', 'dummy_12', 'dummy_13', 'dummy_14',
        'dummy_15', 'dummy_16', 'dummy_17', 'dummy_18', 'dummy_19',
        'dummy_20', 'dummy_21'
        ]
for v in extra_variables:
[...]

PR #688, opened 2018-09-26
Fix bug that many extra regressors cause an predict error after pickle/unpickling

Gedevan-Aleksizde:
I change the extra_regressors dict object to OrderedDict in the Prophet class.

https://github.com/facebook/prophet/issues/643

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -10,11 +10,12 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-from collections import defaultdict
+from collections import defaultdict, OrderedDict
 from datetime import timedelta
 import logging
 import numpy as np
 import pandas as pd
+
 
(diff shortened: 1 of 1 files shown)


bletham:
Perfect, thanks for the fix!

danielvdende:
@bletham Sorry if this isn't the right place to ask this. I tried to find a release roadmap, couldn't find any. When could we expect this fix to be released? I can work around it with an install from github, but it would be much nicer (of course) if we could have a pypi package :-)

bletham:
We will push this early next week :-)

danielvdende:
@bletham any sign of that release? :-)

bletham:
@danielvdende Sorry for the delay, we wanted to push it to PyPI at the same time as the new R version, but it has been more than two weeks and we still haven't gotten CRAN to push out the new R release so we finally just pushed it to PyPI. So you'll find it now.

PR #706, opened 2018-10-16
Figsize argument

floscha:
This PR adds a `figsize` argument to all plotting methods for the Python version of Prophet, similar to the [plot](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.plot.html) method in Pandas. I am aware that some plotting methods have an `ax` argument, that could also be used to adjust the size of the figure. This is however not the case for the `plot_components` method, so I believe that this PR still adds some value.

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -31,6 +31,7 @@ except ImportError:
 
 def plot(
     m, fcst, ax=None, uncertainty=True, plot_cap=True, xlabel='ds', ylabel='y',
+    figsize=(10, 6)
 ):
     """Plot the Prophet forecast.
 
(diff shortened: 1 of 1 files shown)


bletham:
That seems useful to me, thanks!

python/fbprophet/plot.py
```diff
@@ -44,13 +45,14 @@ def plot(
...
     ylabel: Optional label name on Y-axis
+    figsize: Optional tuple width, height in inches.
 
     Returns
     -------
     A matplotlib figure.
     """
     if ax is None:
-        fig = plt.figure(facecolor='w', figsize=(10, 6))
+        fig = plt.figure(facecolor='w', figsize)
```
mlewis-stripe: FYI, this line introduced a bug that disallows the entire `fbprophet` module from being imported.
`SyntaxError: positional argument follows keyword argument`


Should be:
```python
fig = plt.figure(facecolor='w', figsize=figsize)
```

bletham:
@mlewis-stripe Thanks a bunch, fixed in 9928dc39aeac310d143396595f07b254744ce0bf

Issue #700, opened 2018-10-09
If kwarg algorithm is passed to fit() and stan fails to converge we should strip algorithm from kwarg before invoking newtwon fallback


whitehatty:
In `forecaster.py`, lines 1046-1053:
```
            try:
                params = model.optimizing(
                    dat, init=stan_init, iter=1e4, **kwargs)
            except RuntimeError:
                params = model.optimizing(
                    dat, init=stan_init, iter=1e4, algorithm='Newton',
                    **kwargs
                )
```

if kwarg algorithm is passed to fit and a RuntimeError is triggered the call to optimizing will fail due to duplicated kwarg algorithm. 
algorithm should be removed from kwargs before passing it to the call to optimizing.

bletham:
Agreed. Anyone want to make a PR ? :1st_place_medal:

whitehatty:
I can after work

bletham:
Pushed to PyPI.

whitehatty:
@bletham Can we reopen this? I guess some other PR removed the fix, in v0.4 and master, lines 1096--1104, in forecaster.py:
```
            try:
                params = model.optimizing(**args)
            except RuntimeError:
                if 'algorithm' not in args:
                    # Fall back on Newton
                    args['algorithm'] = 'Newton'
                    params = model.optimizing(**args)
                else:
                    raise
```

bletham:
@whitehatty the behavior is a bit different, but it's now intentional and not a bug.

The bug was that previously if RuntimeError, we would repeat the optimization with the Newton optimizer like
```
model.optimizing(algorithm='Newton', **args)
```
which would throw a SyntaxError error if we had previously specified the algorithm in `args`. This was a bug because we were passing in a repeated keyword argument.

Now, if `algorithm` is specified, then it will be used in the first call to `model.optimizing`. If that errors out with a RuntimeError, then we do not replace the algorithm, rather we let it raise. The thought here is that if the user goes to the effort of specifically setting `algorithm='L-BFGS'`, then I don't think we should override that. If the user wants to have this fallback behavior they can not specify the algorithm, or they could handle the RuntimeError in some different way if they wanted.

Does that seem reasonable, or is there a use case that I'm missing?

whitehatty:
My thought is that we should raise only if we can't fit the model, not if a specific algorithm fails, and this branch of code is invoked as a backup case anyway. 

[...]

PR #708, opened 2018-10-17
Fix duplicate kwarg algorithm when newton fallback is selected

whitehatty:
Address:
If kwarg algorithm is passed to fit() and stan fails to converge we should strip algorithm from kwarg before invoking newtwon fallback #700

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1047,7 +1047,7 @@ class Prophet(object):
                 params = model.optimizing(
                     dat, init=stan_init, iter=1e4, **kwargs)
             except RuntimeError:
-		kwargs.pop('algorithm')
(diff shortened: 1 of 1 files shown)


python/fbprophet/forecaster.py
```diff
@@ -1047,6 +1047,7 @@ def stan_init():
                 params = model.optimizing(
                     dat, init=stan_init, iter=1e4, **kwargs)
             except RuntimeError:
+                kwargs.pop('algorithm')
```
bletham:
```suggestion
                kwargs.pop('algorithm', None)
```
otherwise this will throw a KeyError if algorithm is not in kwargs.

whitehatty:
Sorry, pushed the change instead of approving bletham suggestion

bletham:
Great, thanks!

PR #721, opened 2018-10-27
add symmetric mean absolute percentage error

jope35:
added the symmetric mean absolute error.

over and under classification are weighted equally.
also more robust when the actual and/or forecast values contain zero values


bletham:
Interesting, I wasn't familiar with this metric. Thanks for contributing!

PR #751, opened 2018-11-24
Fix url for docstring in Turkey class

bbuluttekin:
Docstring in Turkey class changed with corresponding url.


bletham:
Thanks for the fix!

PR #759, opened 2018-11-29
Update seasonality,_holiday_effects,_and_regressors.md

shuaiwang88:
fix a grammar issue.


bletham:
Thanks for pointing this out! The documentation md files are actually generated automatically from the notebook examples, so the fix has to be made in the jupyter notebook and not the markdown. I went ahead and did it in https://github.com/facebook/prophet/commit/390af322904cef4d489143806d5dc164a1cdc509.

bletham:
It will show up online when we regenerate the documentation tomorrow.

PR #762, opened 2018-11-30
fixed signature of plot_forecast_component in docs

louisguitton:
fbprophet.plot.plot_forecast_component requires the Prophet object as first argument.

Currently following the documentation brings this error:
```
----> 1     fbprophet.plot.plot_forecast_component(fcst=forecast, name=holiday)

TypeError: plot_forecast_component() missing 1 required positional argument: 'm'
```


bletham:
Thanks for the fix!

PR #774, opened 2018-12-19
Use row_number() instead of seq_len(n())

yutannihilation:
This code will no more work with dplyr 0.8.0 if dplyr is not attached (c.f. https://github.com/tidyverse/dplyr/issues/4030).

```r
    dplyr::mutate(col = seq_len(n())) %>%
```

This can be simply fixed by adding `dplyr::` or moving dplyr from `Imports` to `Depends` to make sure it's attached. But, since `seq_len(n())` is equivalent to `row_number()`, this PR uses the latter, which seems a bit smarter.


yutannihilation:
Since unqualified `n()` is already fixed, I'm closing this, though I still feel `row_number()` is better...

https://github.com/facebook/prophet/commit/151c476030ac334b56a3e8d89d8ba8207a834d94

PR #776, opened 2018-12-19
Suppress pkgbuild messages

jimhester:
Fixes https://github.com/r-lib/pkgbuild/issues/54


seanjtaylor:
Thanks Jim, this is super useful!!

bletham:
It looks like this went in in 703c0731f61044e065b4af1254352a8488312e9a so I'll close the PR.

PR #791, opened 2019-01-03
Rather using Monday const from built-in calendar.py

Gordonei:
Latest version of `python-holidays` (0.9.9) changed the name of its `MONDAY` const to `MON`. Rather using the `MONDAY` const from the same library as the object that it is being used with (`calendar` module). 

Also, given [this constant](https://github.com/python/cpython/blame/3cdb5761c5d779b15e153cf33d52babbf56b5807/Lib/calendar.py) is 19 years old (!), I suspect it might not change in the near future.

### DIFF
--- a/python/fbprophet/hdays.py
+++ b/python/fbprophet/hdays.py
@@ -9,11 +9,11 @@
 from __future__ import absolute_import, division, print_function
 
 import warnings
-from calendar import Calendar
+from calendar import Calendar, MONDAY
 from datetime import date, timedelta
 
 from convertdate.islamic import from_gregorian, to_gregorian
-from holidays import MONDAY, WEEKEND, HolidayBase, easter, rd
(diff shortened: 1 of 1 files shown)


Gordonei:
Hey @ivonnics , @louisguitton suggested a work around in [the issue that he opened](https://github.com/facebook/prophet/issues/792#issuecomment-451371215).

bletham:
This is great, thanks so much for the fix.

Gordonei:
☺️ Doing a bit of Git archeology, I think the authors of `holidays` might
have lifted it from the `calendar` module.

To be honest, I also felt a bit responsible, as I recently made a change to
holidays (not the `MONDAY` one, though) that might have prompted this
release.

bletham:
heh, well all is well that ends well :-)

PR #799, opened 2019-01-08
upgraded requirements following #791

louisguitton:
Noticed while using docker that the requirements for prophet are still using the old holidays compatibility.


louisguitton:
@Gordonei is that correct ?

bletham:
https://github.com/facebook/prophet/pull/791 actually works with the older version of holidays too, and so we preferred to leave the requirement at the older version to not make everyone upgrade unecessarily. So this shouldn't be necessary, I was able to install/use fbprophet 0.4.post1 with holidays 0.9.8 without issue.

PR #815, opened 2019-01-18
Python: instantiate future before predicting

dmvianna:
The original code had
```python
forecast = m.fit(df).predict(future)
```
Well. Where did that `future` come from? A reader will be confused, because the output of that will depend of which variable `future` points to. If following the previous examples, this will point to a different dataset, leading to unexpected results.

This proposal makes the origin of the `future` explicit.

### DIFF
--- a/docs/_docs/uncertainty_intervals.md
+++ b/docs/_docs/uncertainty_intervals.md
@@ -60,7 +60,9 @@ forecast <- predict(m, future)
 ```python
 # Python
 m = Prophet(mcmc_samples=300)
-forecast = m.fit(df).predict(future)
+m.fit(df)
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR! The documentation is generated from IPython notebooks, this particular page here: https://github.com/facebook/prophet/blob/master/notebooks/uncertainty_intervals.ipynb

So you'd need to update the code in the notebook as well (otherwise these changes to the md will be overwritten next time I re-generate it from the notebook). Do note that in the notebook it is
```
future = m.make_future_dataframe(periods=60)
```,
and the R and Py sections should be kept in sync (that is, if it is moved in for Py it should be moved in for R also).

PR #830, opened 2019-01-26
Fix FutureWarning from Pandas 0.24.0.

dargor:
Pandas release 0.24.0 has deprecated Series.nonzero(), apply suggested change.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -877,7 +877,7 @@ class Prophet(object):
         first = self.history['ds'].min()
         last = self.history['ds'].max()
         dt = self.history['ds'].diff()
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR. It looks like `.to_numpy()` is new as of pd 0.24.0, and I'd prefer to maintain support for older versions if possible. I think it should work to just replace `.to_numpy()` with `.values`?

dargor:
You're right, I tried your solution and it works :)

dargor:
CI stuck ? Seems green on Travis https://travis-ci.com/facebook/prophet/builds/100568378

bletham:
This looks great, thanks a bunch for the fix.

Issue #839, opened 2019-02-04
Cross validation metrics should compute one value per day for daily data


fercook:
I wanted to implement my own metrics, so I've been working with the code at [fbprophet.diagnostics.performance_metrics](https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py)
and I am confused at how the rolling average is being used. In particular, performance_metrics first [orders the CV results by horizon](https://github.com/facebook/prophet/blob/ef670eaf6cf87a61bb7ee88008ac954d1ddda972/python/fbprophet/diagnostics.py#L224), but a typical CV dataframe will contain several rows for the same horizon, all for different cutoff dates. This dataframe is then used to compute the metrics, and a right side rolling average is used (by default on 10% of the data). However, this means that the rolling average is computed with data from different CVs, and not from different dates. It seems it doesn't matter except that at the boundary between two horizons, we would me mixing metrics from different CVs and different horizons.
Right now, the sequence is:
* order by horizon regardless of cutoff
* compute metric row by row
* rolling average across index that allows mixing horizons and different cross validations
* take mean of metrics with fixed horizon (but they were mixed in step above)

I think it should be something like
* group by cutoff
* order by horizon
* rolling average (because of grouping, it's for a single cross validation)
* take mean between cross validations for fixed horizons

In this way, the averages are cleanly computed for a given horizon.

EDIT:
I did a quick check with a synthetic cross validation dataframe, and I think I am right:
```
cv=pd.DataFrame()
cv['y']=range(1,41)
cv['yhat']=cv['y']+((cv.y-1)/10).astype(int) #each CV is progressively off by one
cv['yhat_lower']=cv['yhat']-0.5
cv['yhat_upper']=cv['yhat']+0.5
cv['cutoff']= pd.Timestamp('2018-1-1')+((cv.y-1)/10).astype(int)*pd.Timedelta('50d') # cutoff dates for 4 CVs
cv['ds']=cv['cutoff']+(1+(cv.y-1)%10).astype(int)*pd.Timedelta('1d') # 1 day at a time, 100 day horizons
```
This simple example has a staircase MSE, and when ordered by the index values it should look like this:
![image](https://user-images.githubusercontent.com/49636/52244571-8a522200-28de-11e9-9421-908fe01ffe76.png)

However this is the output of the performance metrics calculation:
`performance_metrics(cv).sort_index().rmse.plot()`
[...]

PR #841, opened 2019-02-05
Fixes bug #839 on computation of metrics with multiple cross validations

fercook:
Changed computations to grouping by cutoff, in this way the rolling averages are done within each cross validation set, and we avoid computing a metric that mixes data from different cross validations.


bletham:
As discussed in #839, there is an alternative approach that we'll use that is described there.

Issue #538, opened 2018-05-25
Seasonalities that depend on other seasonal factors


bletham:
Recently there have been a number of issues related to seasonalities that depend on other seasonal factors:
* Weekly seasonality that depends on the time of year (summer vs. winter) (#425)
* Daily seasonality that depends on the day-of-week (weekend vs. weekday) (#434, #497, #524)

@vhpietil provides a solution in #434 for constructing separate daily seasonality features for weekends vs. weekdays, but it's a lot of effort.

We should figure out what a clean interface would look like for a more convenient solution to this problem.

bletham:
A big thanks to @Olof-Hojvall who added this new functionality in #894 and #873.

There is now an example in https://github.com/facebook/prophet/blob/master/notebooks/seasonality%2C_holiday_effects%2C_and_regressors.ipynb where there are different weekly seasonalities for winter vs. rest of the year (scroll 2/3rds down to the section "Seasonalities that depend on other factors")

If you have a time series with this property, please install the latest github version with this new feature, try it out, and report any issues. 

You can get the latest github version like this in R:
```
devtools::install_github('facebookincubator/prophet', subdir='R')
```

In Python, remove any existing fbprophet installation (`pip3 uninstall fbprophet`), clone this repository, cd into the `python` directory, and run
```
python3 setup.py install
```

bletham:
Pushed to PyPI and CRAN

PR #861, opened 2019-02-22
Feature Proposal with Python Draft: More flexible seasonality by specifying all periods

Olof-Hojvall:
I've have created a suggestion for a new implementation of seasonality, by predefining a dataframe with every period similarly to how Holidays Work. In the notebook in this PR the DataFrame contains every week in the data with information if it's a NFL week or an off-season week. My implementation then fit two different seasonalities depending on the period name

There are many additional use-cases such as having different daily seasonality on weekends and weekdays for sub-daily data. Additionally, it lets use use exact days in month and year rather than approximating it as 30.5 and 365.25

The concept of a period DataFrame can in future be further extended to allow Half Fourier Series, for when the underlying seasonality is discontinuous, for example between a restaurant closing and opening. Additionally period breaks could be added, too for example avoid applying seasonality when there's also a holiday on that day. 

I'm curious about your thoughts on this, such as if this is a valid method, if you are interested in implementing it fully, or if there's a better way of dealing with these cases.

image

### DIFF
--- a/notebooks/customized_seasonality.ipynb
+++ b/notebooks/customized_seasonality.ipynb
@@ -1,0 +1,228 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%matplotlib inline\n",
+    "from fbprophet import Prophet\n",
+    "import pandas as pd\n",
+    "from matplotlib import pyplot as plt\n",
+    "import logging\n",
+    "logging.getLogger('fbprophet').setLevel(logging.ERROR)\n",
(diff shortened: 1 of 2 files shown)


bletham:
Thanks for the PR. This looks like it would be a solution for #538, which is a high-pri feature. Let me think a little bit more about this interface and the range of things we'd want to support here and I'll get right back with feedback here.

Olof-Hojvall:
Thanks for having a look, curious to hear what you come up with. I created an alternative/complementary solution in another PR, which more directly attempts to solve 538. This kind of flexibility has some other benefits, but is a bit more complex.

bletham:
superseded by #873

Issue #519, opened 2018-05-16
Random order of seasonality charts


dfaustryjak:
Hello

I have really strange situaation. Sometimes when i generating seasonality charts like trend, weekly, yearly. I suddently i obtain order like -> trend, yearly, weekly. Only chart nr 2 and 3 can be revert.  Is that possible to get chart by name ?

bletham:
Currently the seasonal components are pulled from the keys of `m.seasonalities`, which is a dictionary and so inherently unordered (https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L1299).

It does seem better to sort the seasonalities by name before adding them to that list so that we can guarantee the order they will appear in.

In the meantime, you could either manually construct the plot by following the code in the link above (almost everything is delegated to subfunctions so there really isn't that much code there), or you could figure out which axis is which with something like
```
if ax.get_ylabel() == 'weekly':
...
```

bletham:
This is done in https://github.com/facebook/prophet/commit/635ae5a1a6853619a6e4f4e05702c4f7a54e8795

bletham:
Pushed to PyPI.

PR #865, opened 2019-02-26
Take 1 on stan-based predictions when model is fitted using stan::optimizing

shubhankaray:
I inadvertently introduced many small changes by auto-formatting the code in R studio. Let me know if you would like me to revert these changes. I have also retained some of the older code by adding the prefix "older_" to compare new vs. old predictions. I will remove these once I am done with the remaining work with stan::sampling.

### DIFF
--- a/R/R/prophet.R
+++ b/R/R/prophet.R
@@ -1160,43 +1160,23 @@ fit.prophet <- function(m, df, ...) {
   m <- set_auto_seasonalities(m)
   out2 <- make_all_seasonality_features(m, history)
   m <- out2$m
-  seasonal.features <- out2$seasonal.features
-  prior.scales <- out2$prior.scales
-  component.cols <- out2$component.cols
-  m$train.component.cols <- component.cols
+  m$train.component.cols <- out2$component.cols
   m$component.modes <- out2$modes
   
   m <- set_changepoints(m)
-  
-  # Construct input to stan
-  dat <- list(
-    T = nrow(history),
-    K = ncol(seasonal.features),
-    S = length(m$changepoints.t),
-    y = history$y_scaled,
-    t = history$t,
-    t_change = array(m$changepoints.t),
-    X = as.matrix(seasonal.features),
-    sigmas = array(prior.scales),
-    tau = m$changepoint.prior.scale,
-    trend_indicator = as.numeric(m$growth == 'logistic'),
-    s_a = array(component.cols$additive_terms),
-    s_m = array(component.cols$multiplicative_terms),
-    
-    T_pred = 0,                                      #dummy values
-    t_pred = rep(0, 0),                              #dummy values
-    cap_pred = rep(0, 0),                            #dummy values
-    X_pred = matrix(0, 0, ncol(seasonal.features)),  #dummy values
-    n_samp = 1,                                      #dummy values
-    S_pred = 1                                       #dummy values
+  dat <- data_for_stan(
+    m, 
+    df = history, 
+    fit.seasonal.features = out2$seasonal.features,
+    component.cols = out2$component.cols, 
+    prior.scales = out2$prior.scales, 
+    type = "fit"
   )
   
   # Run stan
   if (m$growth == 'linear') {
-    dat$cap <- rep(0, nrow(history))  # Unused inside Stan
     kinit <- linear_growth_init(history)
   } else {
-    dat$cap <- history$cap_scaled  # Add capacities to the Stan data
     kinit <- logistic_growth_init(history)
   }
   
@@ -1210,7 +1190,7 @@ fit.prophet <- function(m, df, ...) {
     list(k = kinit[1],
          m = kinit[2],
          delta = array(rep(0, length(m$changepoints.t))),
-         beta = array(rep(0, ncol(seasonal.features))),
+         beta = array(rep(0, ncol(out2$seasonal.features))),
          sigma_obs = 1
     )
   }
@@ -1230,7 +1210,6 @@ fit.prophet <- function(m, df, ...) {
     args <- utils::modifyList(args, list(...))
     stan.fit <- do.call(rstan::sampling, args)
     m$params <- rstan::extract(stan.fit)
-    m$stan.fit <- stan.fit
     n.iteration <- length(m$params$k)
   } else {
     args <- list(
@@ -1243,9 +1222,10 @@ fit.prophet <- function(m, df, ...) {
     args <- utils::modifyList(args, list(...))
     stan.fit <- do.call(rstan::optimizing, args)
     m$params <- stan.fit$par
-    m$stan.fit <- stan.fit
     n.iteration <- 1
   }
+  
+  m$stan.fit <- stan.fit
   
   # Remove unwanted variables from params
   for (name in c('y_hat', 'seas_hat', 'trend_hat', 'A_pred', 'trend_samples',
@@ -1289,54 +1269,32 @@ fit.prophet <- function(m, df, ...) {
 #' plot(m, forecast)
 #' }
 #' @export
-predict.prophet <- function(object, df = NULL, ...) {
+predict.prophet <- function(m, df = NULL, ...) {
   
   if (is.null(df)) {
-    m <- object
-    history.and.future <- m$history
+    df <- m$history
   } else {
     if (nrow(df) == 0) {
       stop("Dataframe has no rows.")
     }
-    out <- setup_dataframe(object, df)
-    history.and.future <- out$df
+    out <- setup_dataframe(m, df)
+    df <- out$df
     m <- out$m
   }
   
-  out2 <- make_all_seasonality_features(m, history.and.future)
-  seasonal.features <- out2$seasonal.features
-  component.cols <- object$train.component.cols
-  
   # Construct input to stan
-  dat <- list(
-    T = nrow(object$history),
-    K = ncol(seasonal.features),
-    S = length(object$changepoints.t),
-    y = object$history$y_scaled,                         
-    t = object$history$t,                                
-    cap = rep(0, nrow(object$history)),                         
-    t_change = array(object$changepoints.t),
-    X = as.matrix(seasonal.features[1:nrow(object$history),]),            
-    sigmas = array(out2$prior.scales),
-    tau = object$changepoint.prior.scale,
-    trend_indicator = as.numeric(object$growth == 'logistic'),
-    s_a = array(component.cols$additive_terms),
-    s_m = array(component.cols$multiplicative_terms),
-    
-    T_pred = nrow(history.and.future),
-    t_pred = history.and.future$t,
-    cap_pred = 
-      if(class(history.and.future$cap) == "NULL") { 
-        rep(0, length(history.and.future$t)) 
-      } else { 
-        history.and.future$cap 
-      },
-    X_pred = as.matrix(seasonal.features),
-    n_samp = object$uncertainty.samples
-  )
+  out.fit <- make_all_seasonality_features(m, m$history)
+  out.predict <- make_all_seasonality_features(m, df)
   
-  # Future # of change points assuming a homog. Poisson process
-  dat$S_pred <- stats::qpois(0.95, dat$S * (dat$T_pred - dat$T) / dat$T )
+  dat <- data_for_stan(
+    m, 
+    df = df, 
+    fit.seasonal.features = out.fit$seasonal.features,
+    predict.seasonal.features = out.predict$seasonal.features, 
+    component.cols = m$train.component.cols, 
+    prior.scales = out.fit$prior.scales, 
+    type = "predict"
+  )
   
   # Run stan
   if (exists(".prophet.stan.model")) {
@@ -1347,11 +1305,11 @@ predict.prophet <- function(object, df = NULL, ...) {
   
   stan_init <- 
     list(
-      k = object$params$k,
-      m = object$params$m,
-      delta = as.numeric(object$params$delta),
-      beta = as.numeric(object$params$beta),
-      sigma_obs = object$params$sigma_obs
+      k = m$params$k,
+      m = m$params$m,
+      delta = as.numeric(m$params$delta),
+      beta = as.numeric(m$params$beta),
+      sigma_obs = m$params$sigma_obs
     )
   
   if (m$mcmc.samples > 0) {
@@ -1370,7 +1328,8 @@ predict.prophet <- function(object, df = NULL, ...) {
       init = stan_init,
       iter = 0,
       as_vector = FALSE,
-      verbose = TRUE
+      algorithm = 'Newton',
+      verbose = FALSE
     )
     args <- utils::modifyList(args, list(...))
     stan.fit <- do.call(rstan::optimizing, args)
@@ -1382,14 +1341,12 @@ predict.prophet <- function(object, df = NULL, ...) {
       mutate(yhat = y_hat, trend = trend_hat)
     
     # Extract overall and trend intervals
-    stan.intervals <- predict_uncertainty(object, stan.fit)
+    stan.intervals <- predict_uncertainty(m, stan.fit)
   }
-  
-  
   
   df <- 
     cbind(
-      ds = history.and.future$ds, 
+      ds = df$ds, 
       stan.predictions, 
       stan.intervals
     ) %>% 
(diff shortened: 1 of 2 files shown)


R/R/prophet.R
```diff
@@ -1223,6 +1230,7 @@ fit.prophet <- function(m, df, ...) {
     args <- utils::modifyList(args, list(...))
     stan.fit <- do.call(rstan::sampling, args)
     m$params <- rstan::extract(stan.fit)
+    m$stan.fit <- stan.fit
```
bletham: instead of having this here and in L1246, we can just put it once in L1249

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+    history.and.future <- out$df
+    m <- out$m
+  }
+  
+  out2 <- make_all_seasonality_features(m, history.and.future)
+  seasonal.features <- out2$seasonal.features
+  component.cols <- object$train.component.cols
+  
+  # Construct input to stan
+  dat <- list(
```
bletham: let's have a helper function that computes the model-fitting portion of this, that will be called both here and by fit, and then we can just add in the prediction items.

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+      stop("Dataframe has no rows.")
+    }
+    out <- setup_dataframe(object, df)
+    history.and.future <- out$df
+    m <- out$m
+  }
+  
+  out2 <- make_all_seasonality_features(m, history.and.future)
+  seasonal.features <- out2$seasonal.features
+  component.cols <- object$train.component.cols
```
bletham: since we went ahead and renamed `object` to `m` above, let's use `m` here and everywhere below instead of `object`.

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+    sigmas = array(out2$prior.scales),
+    tau = object$changepoint.prior.scale,
+    trend_indicator = as.numeric(object$growth == 'logistic'),
+    s_a = array(component.cols$additive_terms),
+    s_m = array(component.cols$multiplicative_terms),
+    
+    T_pred = nrow(history.and.future),
+    t_pred = history.and.future$t,
+    cap_pred = 
+      if(class(history.and.future$cap) == "NULL") {
```
bletham: My personal preference is `'cap' %in% colnames(history.and.future)` since I think it reads a little more cleanly (and more closely parallels the python syntax)

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+  component.cols <- object$train.component.cols
+  
+  # Construct input to stan
+  dat <- list(
+    T = nrow(object$history),
+    K = ncol(seasonal.features),
+    S = length(object$changepoints.t),
+    y = object$history$y_scaled,                         
+    t = object$history$t,                                
+    cap = rep(0, nrow(object$history)),
```
bletham: this needs to use the correct cap (can be handled in the helper function mentioned above)

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
 #' forecast <- predict(m, future)
 #' plot(m, forecast)
 #' }
-#'
 #' @export
 predict.prophet <- function(object, df = NULL, ...) {
+  
+  if (is.null(df)) {
+    m <- object
+    history.and.future <- m$history
```
bletham: I think the variable name `history.and.future` isn't quite appropriate because while by default we always generate future dataframes that include the history, it doesn't have to. `df` may contain only future dates.

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+  # Construct input to stan
+  dat <- list(
+    T = nrow(object$history),
+    K = ncol(seasonal.features),
+    S = length(object$changepoints.t),
+    y = object$history$y_scaled,                         
+    t = object$history$t,                                
+    cap = rep(0, nrow(object$history)),                         
+    t_change = array(object$changepoints.t),
+    X = as.matrix(seasonal.features[1:nrow(object$history),]),
```
bletham: this assumes that the input `df` includes the history, which it may not.

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+        rep(0, length(history.and.future$t)) 
+      } else { 
+        history.and.future$cap 
+      },
+    X_pred = as.matrix(seasonal.features),
+    n_samp = object$uncertainty.samples
+  )
+  
+  # Future # of change points assuming a homog. Poisson process
+  dat$S_pred <- stats::qpois(0.95, dat$S * (dat$T_pred - dat$T) / dat$T )
```
bletham: the notation on the Stan side is quite poor, which is my fault because I'm the one who originally defined it :-) But, T_pred and T are the number of rows. What we really want here are the relative lengths of the periods in "real" time, which may be different if the times in the history or future are not regularly spaced. That is, we want little `t` (which is the time) and not the big `T` (which is number of rows). `t` for the history is mapped to [0, 1]. The future begins at 1, and then goes forward from there for however long the future is relative to the history. (e.g., if we have 2 years of data and make a prediction of 1 year, dat$t will be on [0, 1] and dat$t_pred will be on [1, 1.5]).

So what we want here is `dat$S * (max(dat$t_pred) - 1) / (1 - 0) = dat$S * (max(dat$t_pred) - 1)`. Does that make sense?

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+    args <- utils::modifyList(args, list(...))
+    stan.fit <- do.call(rstan::sampling, args)
+  } else {
+    args <- list(
+      object = model,
+      data = dat,
+      init = stan_init,
+      iter = 0,
+      as_vector = FALSE,
+      verbose = TRUE
```
bletham: I expect we'll want `algorithm='Newton'` here too. There are reports that the L-BFGS can hang with very flat likelihood surfaces.

R/R/prophet.R
```diff
@@ -1274,9 +1288,134 @@ fit.prophet <- function(m, df, ...) {
...
+      predict_seasonal_components(object, history.and.future)
+    )
+  
+  
+  return(df)
+}
+
+
+# Older version of predict(...) to be removed 
+old_predict <- function(object, df = NULL, ...) {
```
bletham: go ahead and just remove this now, I'd like to see how things look without the old functionality.

R/R/prophet.R
```diff
@@ -1305,6 +1444,7 @@ predict.prophet <- function(object, df = NULL, ...) {
   return(df)
 }
 
+
```
bletham: we should be able to remove the `piecewise_linear` and the `piecewise_logistic` functions with this change, as well as `predict_trend`.

R/R/prophet.R
shubhankaray: Do you want to set it to 'Newton' and change the current behavior, where the user can choose between the algorithms while calling prophet?

bletham: I meant in predict. In fit we'll keep it like it is (use L-BFGS by default and fallback to Newton), but in predict when we take our 1 step it can be with Newton.

shubhankaray: ok makes sense.

R/R/prophet.R
shubhankaray: Got it, I mixed up upper and lower case t.

R/inst/stan/prophet.stan
```diff
@@ -39,51 +42,60 @@ functions {
...
-    vector t_change,
-    int S
+    real k, real m, vector delta, vector t, vector cap, matrix A,
+    vector t_change, int S
   ) {
     vector[S] gamma;
 
     gamma = logistic_gamma(k, m, delta, t_change, S);
-    return cap .* inv_logit((k + A * delta) .* (t - (m + A * gamma)));
+    return cap ./ (1 + exp(-(k + A * delta) .* (t - (m + A * gamma))));
```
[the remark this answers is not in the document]
shubhankaray: Not sure why I changed this, I am reverting to what it was before.

shubhankaray:
I have made most of the changes and created a new function _data_for_stan_ which is called from _fit.prophet_ and _predict.prophet_ to create the list _dat_. Previously I was assuming that the df used for predictions included historical data, but now this is no longer the case; and I am using two separate dfs namely fit.seasonal.features & predict.seasonal.features in _data_for_stan_. Not sure if this is the best way, but please take look and let me know what you think. Also I am not sure about how _cap_ and _cap_pred_ is initialized in this function, having not used with this part of the model in my forecasting work.

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
 #' forecast <- predict(m, future)
 #' plot(m, forecast)
 #' }
-#'
 #' @export
-predict.prophet <- function(object, df = NULL, ...) {
+predict.prophet <- function(m, df = NULL, ...) {
```
bletham: In the earlier review I had suggested changing everything below from object to m, but I forgot an important detail: having the argument here be named `object` is actually required by R because we are making an implementation of `predict`. So that will unfortunately have to be changed back.

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
...
+    args <- list(
+      object = model,
+      data = dat,
+      init = stan_init,
+      iter = 0,
+      as_vector = FALSE,
+      algorithm = 'Newton',
+      verbose = FALSE
+    )
+    args <- utils::modifyList(args, list(...))
```
bletham: Actually let's don't do this. In `fit`, we accept any args and pass them along to stan to give more control over the fitting (especially if you're doing MCMC). Here in `predict`, `predict.prophet` accepts `...` args because it is required in order to implement `predict`, but we don't want to pass them along to Stan. Since we're not actually doing any fitting, there's no need to give control here.

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
...
+#' @param prior.scales prior scale hyperparameters
+#' @param type 'fit' or 'predict'
+#' 
+#' @return List with all data required for stan fitting or prediction
+#'
+#' @keywords internal
+data_for_stan <- function(
+  m, df, 
+  fit.seasonal.features, 
+  predict.seasonal.features = matrix(0, 0, ncol(fit.seasonal.features)),
```
bletham: Optional arguments (with default values) should be after non-optional arguments.

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
...
+  component.cols, 
+  prior.scales, 
+  type) {
+  
+  if (type == "fit") {
+    
+    dat <- list(
+      T = nrow(df),
```
bletham: this can just be m$history

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
...
+        linear = rep(0, nrow(df)),                     # Unused inside Stan
+        df$cap_scaled                                  # Add capacities to the Stan data
+      ), 
+      t_change = array(m$changepoints.t),
+      X = as.matrix(fit.seasonal.features),
+      sigmas = array(prior.scales),
+      tau = m$changepoint.prior.scale,
+      trend_indicator = as.numeric(m$growth == 'logistic'),
+      s_a = array(component.cols$additive_terms),
+      s_m = array(component.cols$multiplicative_terms),
```
bletham: everything from L1397 through 1413 here is repeated below for prediction. 

How about if we construct a list with these regardless of `type`, and then have the `if type == "fit"...` just to add in `T_pred`, `t_pred`, `cap_pred`, `X_pred`, `n_samp`, and `S_pred`?

R/R/prophet.R
```diff
@@ -1274,37 +1268,203 @@ fit.prophet <- function(m, df, ...) {
...
+      
+      T_pred = nrow(df),
+      t_pred = df$t,
+      cap_pred = 
+        if(!'cap' %in% colnames(df)) {
```
bletham: This won't be necessary. If growth='linear', then we would have already raised an error if df did not have a `cap` column.

shubhankaray:
I've made the changes. Let me know if anything comes up in the testing process. 

For predictions under the sampling option, looping over what I have currently with the optimizing option should be fine I think. Let me go through that part of the code to better understand how its done currently. Do you think we can also give a try to the sampling option in stan for predictions, if it provides any performance benefits?

bletham:
Great, I'll test this next week.

As for doing sampling in prediction: I think there are two ways we can handle MCMC. The first is to do the sampling on fit, and then on predict loop over the samples. This is what I had described earlier.

The second approach is to not do MCMC on fit, but instead just do the full MCMC on predict.

There are pros and cons to each. The second approach (MCMC just on predict) would be a lot cleaner to implement (the pro). There are two cons though. The first is that if you want to make a prediction for a different set of dates, you don't have to re-do the MCMC. Looping over samples for predict is kindof expensive, but still way cheaper than the MCMC. I'm not sure how important that use case is, but it is a nice property to have that you only have to do the MCMC once. The second issue is that we currently do MCMC only on fit, and so moving it to predict would be a pretty significant change and could affect people's workflows. I try to minimize those sorts of changes across versions unless there's a pretty compelling reason to. That's the main reason why I prefer approach (1), even though it's definitely going to be a bit uglier in the implementation. If there's another advantage to moving it to predict that you can think of let me know and we can discuss if it is worth doing.

shubhankaray:
I agree with your points and approach 2 may be too big of a change and lead to instabilities. After the tests, I will go ahead and make the changes to implement approach 1 .

bletham:
OK, I did a head-to-head comparison on two of the documentation examples: the quickstart one here, https://github.com/facebook/prophet/blob/master/notebooks/quick_start.ipynb and the sub-daily data one at the top of here , https://github.com/facebook/prophet/blob/master/notebooks/non-daily_data.ipynb

The good news is that the predictions seem to be the same. The bad news is that it is quite a bit slower.

The blocker for this last summer was an issue in rstan #519 that transferring data from Stan into R was really slow. They pushed a fix for that issue that did make things much faster, but there's still a pretty noticeable lag, and most importantly it's making things quite a bit slower than they currently are. In the quickstart example `predict` was 5x slower (25s vs. 5s), and in the non-daily data (which has a much larger history, and so the matrix of samples being transferred from Stan is much larger) it was 10x slower (4mins vs. 25 seconds).

I really want the flexiblity that we get from moving predictions to Stan and am super excited by how much code was deleted in this PR, but a 5-10x slowdown is just too much.

I think we can do better. The reason it is slow is because we are passing back from Stan these very large T_pred x n_samp matrices. When I make T_pred small (by not including the training data in `future` and forecasting only 2 periods), then this PR is actually 3x _faster_ than current prophet. It is only for T_pred about 80 that it starts to be slower. That is, for an 80x1000 matrix we break even, but bigger than that and it starts to be slower. So I'm thinking that the right solution here would be to not return the T_pred x nsamp matrix unless we really need it. Most of the time (for a usual call to predict), we only care about the percentiles of trend_samples and yhat_samples (that's what predict computes, and then throws away the samples). If we move the percentiles computing into Stan, then it would just have to return a T_pred x 2 matrix, which based on the test above I'd expect to be much faster.

We do still have to have an option to return the full samples for the predictive_samples function, and so that would still be slow, but I think all of the other gains that we get from doing this would be worth the slowdown in predictive_samples, as long as predict stays fast.

This is going to require a bit of a refactor of the GQ block, to move it into a function that would return percentiles if those are requested or the full samples if those are requested. And Stan doesn't seem to have a quantile function yet, but it does have a sort function that we can use, and we could in R compute the indicies of the quantiles that we want and pass those in as ints. Hopefully computing the percentiles row by row like that won't be too slow (I'd expect not given that it will be compiled down).

That's a lot of text, but here's what I would propose:
* Let's move the quantile computing in to Stan, so that we return a (T_pred x 2) matrix rather than a (T_pred x n_samp).
* Everything happening in the GQ block can be moved into a function that returns those things so that all of the other kinda-large stuff in the GQ block (A_pred, A_sim) can be declared in the function rather than in the GQ block and (hopefully I think?) thus not have to be returned to R.
* If that is working and is fast, then we can add in the option to return all of the samples, and will hook `predictive_samples` up to that.

bletham:
A few other minor things that were important for testing:

* You can delete the `predict_trend` function
* In `predict.prophet`, there are a bunch of functions that do not have their package specified. Every function that does not come from base needs to have its package declared, like `dplyr::mutate`, `dplyr::select`, `dplyr::mutate_at`, and `dplyr::funs` I think were all of them.

shubhankaray:
I did notice a slowdown although I didn't test it enough to figure out the reason. I will try to move the quantile calculations to stan and I think a simple sort should be close enough. Let me dig into some examples of the GQ block, I will reach out to you if I have any questions.

seanjtaylor:
This looks useful for computing quantiles: https://discourse.mc-stan.org/t/quantile-function-in-stan/3642/12

Hopefully this is fast and then the I/O costs drop down to a very small number and we're all good! I'd be happy to do the Python implementation (deleting code is my favorite).

shubhankaray:
Thanks, will take a look. I should be able make the changes and push the code in the next few days.

shubhankaray:
Returning the quantiles from stan seems to help. In my test cases the predict function is much faster. Let me know if you see a similar reduction in prediction times in the notebook examples.

bletham:
This looks great. On the quickstart example it was 40% faster, and on the sub-daily data example it was 22% faster. Let's get the missing components in, and then we can merge the PR and see if there's any additional performance tuning that can be done afterwards.

I believe the missing components are:
* The `predictive_samples` function.

I think the best way to handle this will probably be to add an additional flag input to GQ for whether all samples should be returned or just the quantiles. On predict, this will return an empty matrix for samples, on `predictive_samples` it will incur the extra cost to return the full samples matrix.

* MCMC sampling.

Now that I think about it, this will be a bit of a challenge with this approach of computing quantiles in stan. Suppose we have 200 MCMC samples and we want to estimate uncertainty intervals with 1000 samples. What we currently do is we loop through those 200 MCMC samples, and then draw 5 uncertainty samples for each one. We stack all of those to get the total 1000 and then compute quantiles. The only way that would work here would be if for every MCMC sample we extracted the full set of samples, and then computed the quantiles back in R. As we know from above, transferring all of those samples from Stan back to R is very slow.

The alternative is to for MCMC do the sampling at predict time rather than fit time. We can still do MAP optimization at fit time, and then use it to initialize the sampling at predict time. We discussed this a bit above and I was not a fan of this approach because it means you have to repeat the sampling if you change the set of dates at which you make predictions. This makes MCMC not-an-option if you want to do "prophet as a service" type of design. Here, the more common pattern is to make predictions on all the dates you think you'll need and store those (where this wouldn't be an issue), but I don't know what other people externally might be doing. This would, however, make the total fit+predict time no slower for MCMC than it currently is.

@seanjtaylor what do you think? I lean a bit towards the 2nd?

bletham:
With the addition of the cmdstanpy backend I think using GQ for predictions in Stan is going to be a bit more difficult (more places to have to handle two different GQ APIs). I've been thinking now that the best approach for making it easier to customize the model will be to make the model more modular and have a clear API/separation between functions that are evaluating trends/link functons, and the other Prophet pieces like constructing the regressor matrix. So I'm going to go ahead and close this now since I don't think we'll be be pursuing this option right now. But a big thanks @shubhankaray for exploring this approach and assessing its viability. You do great work!

PR #873, opened 2019-03-03
Feature Proposal with Python Draft: Masked seasonality

Olof-Hojvall:
This PR suggest a solutions to some problems relating to https://github.com/facebook/prophet/issues/538. This is a much simpler solution than in my other proposal https://github.com/facebook/prophet/pull/861 PR, where this issue was brought up. 

To get the suggested "summer vs. winter" or "weekend vs. weekday" effects, you would in this proposal add two weekly/daily seasonalities, and then mask the weekend one on weekdays and reverse.

```
m = Prophet(holidays=holidays, weekly_seasonality=False)
m.add_seasonality('nfl_weekly', period=7, fourier_order=3, condition_name='has_nfl_weekly')
m.add_seasonality('offseason_weekly', period=7, fourier_order=3, condition_name='has_offseason_weekly')

df['has_nfl_weekly'] = (df['ds'].dt.month > 8) | (df['ds'].dt.month < 2)
df['has_offseason_weekly'] = ~df['has_nfl_weekly']
m.fit(df)

future = m.make_future_dataframe(periods=366)
future['has_nfl_weekly'] = (future['ds'].dt.month > 8) | (future['ds'].dt.month < 2)
future['has_offseason_weekly'] = ~future['has_nfl_weekly']
forecast = m.predict(future)
```

image


bletham:
Ok I like this and it's definitely along the lines of what I was thinking. I do think this type of interface will be a bit cleaner than the dataframe approach in #861.

My original thought had been something like
```
m = Prophet()
m.add_regressor('is_nfl_week')
m.add_conditional_seasonality(name='nfl_weekly', regressor='is_nfl_week', value=1, fourier_order=6)
```
where basically you would first specify something as an extra regressor, and then you could interact a seasonality with it.

I think I like your interface more. I like that it directly uses the existing `add_seasonality`, and the amount of code to implement is surprisingly low (nice and clean :-) ).

Right now there is some validation in `https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py#L218` that makes sure that all of the columns specified as extra regressors are in the dataframe. We'll have to do the same thing to make sure that these seasonality condition columns are there, and then also validate that they are boolean or binary.

This is really great, thanks for the contribution. Go ahead and finish it up and I'll merge it in once it all looks good there.

If you'd like to make the corresponding change in the R then let me know (it can be a separate PR), otherwise I'll do it as soon as this goes in.

Olof-Hojvall:
Thanks for the positive response and agree that is an easier solution to most problems.
I've done an initial try at preparing all required changes according to your comments.

I additionally had to make some changes for the seasonality plots to work. Saw a few options here, so please check if you prefer it solved differently. plot_components currently results in this plot: 
image

I have less experience with R, but would be happy to make an attempt after design here is finished.

python/fbprophet/forecaster.py
```diff
@@ -634,6 +647,9 @@ def add_seasonality(
             'prior_scale': ps,
             'mode': mode,
         }
+        if condition_name is not None:
```
Olof-Hojvall: I know only add condition_name to seasonality dict if condition_name is not None, but could also refactor the code such that condition_name is always present. Any preference?

python/fbprophet/tests/test_prophet.py
```diff
@@ -580,6 +580,43 @@ def test_custom_seasonality(self):
...
+        )
+        self.assertNotIn('condition_name', m.seasonalities['normal_monthly'])
+        seasonal_features, prior_scales, component_cols, modes = (
+            m.make_all_seasonality_features(m.history)
+        )
+        # Confirm that only values without is_conditional_week has non zero entries
+        conditional_weekly_columns = seasonal_features.columns[
+            seasonal_features.columns.str.startswith('conditional_weekly')]
+        np.testing.assert_array_equal((seasonal_features[conditional_weekly_columns] != 0).any(axis=1).values,
+                                      df['is_conditional_week'].values)
```
Olof-Hojvall: Any more test that would be good, or any test that are useless? Could make this part of the test_custom_seasonality above instead.

python/fbprophet/plot.py
```diff
@@ -136,9 +136,10 @@ def plot_components(
                 m=m, fcst=fcst, name='trend', ax=ax, uncertainty=uncertainty,
                 plot_cap=plot_cap,
             )
-        elif plot_name == 'weekly':
+        elif plot_name == 'weekly' or (plot_name in m.seasonalities and
+                                       m.seasonalities[plot_name]['period'] == 7):
```
Olof-Hojvall: Instead of modifying plot_weekly, we could remove the added condition here and always pass it to  plot_forecast_component instead. It's however a bit confusing in this case since weekdays are not visible in plot_forecast_component.

bletham: Yeah this is nice. We should probably do a little refactoring here and just get rid of the `plot_name == 'weekly'` check altogether and just have it use `plot_weekly` if period==7 as is done here, and the same for yearly `if period ==365.25`.  Then we can delete L106-L111, and replace L115 with `if name in fcst`

python/fbprophet/tests/test_prophet.py
```diff
@@ -580,6 +580,43 @@ def test_custom_seasonality(self):
...
+            },
+        )
+        self.assertNotIn('condition_name', m.seasonalities['normal_monthly'])
+        seasonal_features, prior_scales, component_cols, modes = (
+            m.make_all_seasonality_features(m.history)
+        )
+        # Confirm that only values without is_conditional_week has non zero entries
+        conditional_weekly_columns = seasonal_features.columns[
+            seasonal_features.columns.str.startswith('conditional_weekly')]
+        np.testing.assert_array_equal((seasonal_features[conditional_weekly_columns] != 0).any(axis=1).values,
```
bletham: self.assertTrue(np.array_equal(...))

python/fbprophet/tests/test_prophet.py
bletham: Nah this looks great to me.

python/fbprophet/forecaster.py
bletham: I do have a preference for having it always present, as None if not used. Is there a reason not to? It seems to me like things will be just as clean since we end up doing `is not None` checks anyway.

Olof-Hojvall: I have changed it now, not important for me. If not present you could avoid always having to define it when creating a seasonality. A fun alternative to dict for this situation is dataclasses https://docs.python.org/3/library/dataclasses.html, but not to common in use yet.

bletham:
Beautiful, thanks a bunch! If you do want to give it a stab in the R you'll find that the code there very closely parallels that in Py (intentionally to make it easy to copy features from one to the other), and so this exact set of changes will be what is required over there too.

(But if it does end up being really painful for you let me know and I can put it in later this week).

python/fbprophet/plot.py
Olof-Hojvall: Seems like L106-L111 was added to keep order same, so not sure if I should chnage that. Did some light refactoring here, kept plot_name == 'weekly'  and 'yearly to avoid changing behavior if someone has modified the period, but that's probably quite rare.

gabidan:
Hello all,
Wonder if you could help with adding custom seasonalities - I am new to Prophet and struggle to understand the problem. 
Following the example above I attempt to add a custom seasonality indicating 'weekend' or 'non-weekend' day :


`m = Prophet(weekly_seasonality=False)
m.add_seasonality(name='weekend', period=7, fourier_order=3, condition_name='weekend')
m.add_seasonality(name='non_weekend', period=7, fourier_order=3, condition_name='non_weekend')`

I receive the error below: 

`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-138-a3d749a3b813> in <module>
      1 m = Prophet(weekly_seasonality=False)
----> 2 m.add_seasonality(name='weekend', period=7, fourier_order=3, condition_name='weekend')
      3 m.add_seasonality(name='non_weekend', period=7, fourier_order=3, condition_name='non_weekend')

TypeError: add_seasonality() got an unexpected keyword argument 'condition_name'
`

Not sure what I am doing wrong? Thank you so much for your help

Olof-Hojvall:
Hi @gabidan .
This feature was released with v0.5, could you check if you have it installed?
```
import fbprophet
print(fbprophet.__version__)
```

gabidan:
> Hi @gabidan .
> This feature was released with v0.5, could you check if you have it installed?
> 
> ```
> import fbprophet
> print(fbprophet.__version__)
> ```

@Olof-Hojvall ,

Thanks a million! I have reinstalled v.05 and it has worked. 

Unfortunately it hasn't showed an improvement - sorry to burden with further comments, but maybe you could share some thoughts. 

 Saturdays always reach a peak,  followed by Sunday, but forecast never picks up the peaks. 

I have tried using Saturday as an additional regressor:

`model5 = Prophet(interval_width = 0.80, holidays = holidays, holidays_prior_scale = 100, changepoint_prior_scale=30, seasonality_prior_scale =30, seasonality_mode= 'additive')`
                 
`model5.add_seasonality(name='weekly', period=3, fourier_order=30)`
`model5.add_seasonality(name='monthly', period=12, fourier_order=30)`

`model5.add_regressor('saturday', prior_scale=100 )`

When I remove weekly from the model above, forecast also doesn't reach MIN points. 

Using Saturday, Sunday as custom seasonality:

`m = Prophet(weekly_seasonality=False, holidays=holidays, holidays_prior_scale=60, seasonality_prior_scale =60)`

`m.add_seasonality(name='weekend',  period=7, fourier_order=3, condition_name='weekend')`
`m.add_seasonality(name='non_weekend', period=7, fourier_order=3, condition_name='non_weekend')`

`m.add_seasonality(name='monthly', period=12, fourier_order=30)`

`m.add_seasonality(name='yearly', period=365.25, fourier_order=20)` 

However, neither of these do much impact, and my forecast pretty much always looks like this: 

Screenshot 2019-06-03 at 17 41 55

Tried multiple variations of hyper parameters, mainly playing with prior_scale, but nothing makes much difference.

Would you have any ideas how to reach those peaks? 

My original timeseries looks like this:
Screenshot 2019-06-03 at 17 57 59





Thank you so much again

Olof-Hojvall:
The use-case of using conditional seasonality to separate weekends is mainly if you have hourly data, and applying it on the daily seasonality. For example hourly visitors to a store, which you would expect to be different on weekdays and weekends. Your use-case appears different from a quick glance, with daily data and applying on weekly seasonality. The default weekly seasonality, maybe after some parameter tuning, should be able to catch the peaks if they happen every Saturday, is that the case here?

gabidan:
Thank you Olof, that's very helpful, and makes sense. 
I managed to tune the model so it picks up the peaks on Saturdays during certain months, but some consistently fall behind. 
I might play with custom changepoints for those periods, hoping it will help, thank you again!

MeganBeckett:
Hi there

I've been doing a similar implementation, with hourly data, where I have strong peaks on Friday and Saturday evenings. I'm battling to get the `prophet` model to be sensitive to these strong peaks and it always underestimates.

I did a version of this recommendation where I add the daily seasonality for weekend nights separately to weekday nights based on a condition (https://facebook.github.io/prophet/docs/seasonality,_holiday_effects,_and_regressors.html#seasonalities-that-depend-on-other-factors).

But, it didn't make much difference.

Furthermore, I have tried adjusting the `prior.scale` for the `add_seasonality` and it still hasn't made any difference when visually inspecting the predicted values (even changing up to prior.scale = 1000).

image

And with prior scale changed all the way up to 1000 for the add_seasonality component for Friday and Saturday nights:

image

Any advice on what I can further try?

bletham:
@MeganBeckett could you look at the residuals (y_hat-y) for Fridays and Saturdays? It's definitely missing a bunch of peaks, but I'm wondering if there are other Fridays/Saturdays that have lower values and those are keeping the predictions lower. The fitted seasonality will be something like the average seen across all days.

bletham:
(@MeganBeckett actually #894 is a better place to discuss that, moving my comment there)

Issue #877, opened 2019-03-06
Fall back to Newton optimizer even if algorithm is specified as kwarg


bletham:
Discussion is in #700.
* If optimization fails with default L-BFGS, then it is repeated with Newton.
* If algorithm is explicitly specified as a kwarg, then this isn't done and the optimization failure is raised.

This should be changed to fallback to Newton even if algorithm is specified.

whitehatty:
Open PR:
https://github.com/facebook/prophet/pull/890

bletham:
Pushed

PR #890, opened 2019-03-16
If optimization fails fallback to Newton even if algorithm is specified

whitehatty:
If optimization fails fallback to Newton even if algorithm is specified.

See:
https://github.com/facebook/prophet/issues/877


bletham:
Beautiful, thanks!

PR #894, opened 2019-03-20
Conditional seasonality Feature in R

Olof-Hojvall:
This PR attempts to convert the previously merged Python PR https://github.com/facebook/prophet/pull/873 into R. See the Python PR for more details.

image

### DIFF
--- a/R/R/diagnostics.R
+++ b/R/R/diagnostics.R
@@ -108,6 +108,12 @@ cross_validation <- function(
       }
     }
     columns <- c(columns, names(m$extra_regressors))
+    for (name in names(m$seasonalities)) {
+      condition.name = m$seasonalities[[name]]$condition.name
+      if (!is.null(condition.name)) {
+        columns <- c(columns, condition.name)
+      }
+    }
     future <- df.predict[columns]
     yhat <- stats::predict(m, future)
     # Merge yhat, y, and cutoff.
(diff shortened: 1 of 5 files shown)


bletham:
This is fantastic, thanks again!

I'll add a section to the documentation notebook now that describes this new feature, and I'm planning for a new version release in about a month.

MeganBeckett:
Hi there

Thanks for this notebook and documentation - really useful!

I've been doing a similar implementation, with hourly data, where I have strong peaks on Friday and Saturday evenings. I'm battling to get the `prophet` model to be sensitive to these strong peaks and it always underestimates.

I did a version of this recommendation where I add the daily seasonality for weekend nights separately to weekday nights based on a condition (https://facebook.github.io/prophet/docs/seasonality,_holiday_effects,_and_regressors.html#seasonalities-that-depend-on-other-factors).

But, it didn't make much difference.

Furthermore, I have tried adjusting the prior.scale for the `add_seasonality` and it still hasn't made any difference when visually inspecting the predicted values (even changing up to `prior.scale = 1000`).

image

And with prior scale changed all the way up to 1000 for the `add_seasonality` component for Friday and Saturday nights:

image

Any advice on what I can further try?

bletham:
Could you look at the residuals (y_hat-y) for Fridays and Saturdays? It's definitely missing a bunch of peaks, but I'm wondering if there are other Fridays/Saturdays that have lower values and those are keeping the predictions lower. The fitted seasonality will be something like the average seen across all days.

Issue #14, opened 2017-02-25
Use dygraphs for plotting


jaredlander:
`dygraphs` makes some great interactive plots.  You can zoom in, hover, etc.  I coded it to be the generic `plot` method but replacing `ggplot2` is a bit extreme.  So maybe the generic `plot` should call either  the `ggplot2` or `dygraphs` functions depending on an argument flag, like `plot(m, forecast, dygraphs=TRUE)`?

In terms of compatibility it can be used from the console and in `rmarkdown` documents for rendering to HTML or pdf docs (pdf requires the latest versions of `knitr` and `webshot`).

```R
#' Plot the prophet forecast.
#'
#' @param x Prophet object.
#' @param fcst Data frame returned by predict(m, df).
#' @param uncertainty Boolean indicating if the uncertainty interval for yhat
#'  should be plotted. Must be present in fcst as yhat_lower and yhat_upper.
#' @param actual.color Color of points for actual data
#' @param forecast.color Color of forecast line
#' @param forecast.label Text for legend
#' @param actual.label Text for legend
#' @param forecast.pattern Type of line for forecast. Must be one of c("dashed", "dotted", "dotdash", "solid").
#' @param actual.size Size of points for actual data
#' @param forecast.size Thickness of line for forecast data
#' @param legend.width Width of legend
#' @param group Group to associate this plot with. The x-axis zoom level of plots within a group is automatically synchronized.
#' @param ... additional arguments
#' @importFrom magrittr "%>%"
#' @return A dygraph plot.
#'
#' @examples
#' \dontrun{
#' history <- data.frame(ds = seq(as.Date('2015-01-01'), as.Date('2016-01-01'), by = 'd'),
#'                       y = sin(1:366/200) + rnorm(366)/10)
#' m <- prophet(history)
#' future <- make_future_dataframe(m, periods = 365)
#' forecast <- predict(m, future)
#' plot(m, forecast)
#' }
#'
#' @export
plot.prophet <- function(x, fcst, uncertainty=TRUE, 
                         actual.color='black', forecast.color='#0072B2',
                         forecast.label='Predicted', actual.label='Actual',
                         forecast.pattern=c("dashed", "dotted", "dotdash", "solid"),
                         actual.size=2, forecast.size=1,
                         legend.width=400,
                         ...) 
{
    forecast.pattern <- match.arg(forecast.pattern)
    
    # create data.frame for plotting
    df <- df_for_plotting(x, fcst)

[...]

PR #915, opened 2019-04-03
Interactive Plotly Forecast for Python

Olof-Hojvall:
This PR want to add an interactive Plotly plot of the forecast generated by Prophet. This attempts to solve issues https://github.com/facebook/prophet/issues/14 for python, but by using Plotly instead of dygraphs.

Examples of plots can be seen [here](https://nbviewer.jupyter.org/github/Olof-Hojvall/prophet/blob/plotly_plot_notebook/notebooks/Prophet%20Plotly%20Plot%20Draft.ipynb)

Happy for any suggestions, tried to follow design of existing plots but there are some deviations.

Would be nice to also do interactive components plot if this gets done.

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -25,6 +25,12 @@ try:
 except ImportError:
     logger.error('Importing matplotlib failed. Plotting will not work.')
 
+try:
+    import plotly.graph_objs as go
+    import plotly.offline
+except ImportError:
+    logger.error('Importing plotly failed. Interactive plots will not work.')
+
 
 def plot(
     m, fcst, ax=None, uncertainty=True, plot_cap=True, xlabel='ds', ylabel='y',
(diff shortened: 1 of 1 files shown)


python/fbprophet/plot.py
```diff
@@ -495,3 +501,152 @@ def plot_cross_validation_metric(
...
+                    dict(step='all')
+                ])
+            ),
+            rangeslider=dict(
+                visible=True
+            ),
+        ),
+    )
+    fig = go.Figure(data=data, layout=layout)
+    return plotly.offline.iplot(fig)
```
Olof-Hojvall: A bit tricky what to return here. This is one option which requires plotly.offline.init_notebook_mode() to be run first in notebooks. 
Another option is maybe to do what is done in Plotly Express using a new [ExpressFigure](https://github.com/plotly/plotly_express/blob/68d7ff0bbb53275ab0e0edf014caf8997ecc9c0c/plotly_express/_core.py#L22), but don't want to require a package only to use that (internal?) class and I don't fully get how it works myself. 
A third option is to return the Figure, and let the user do the iplot command. This would let users use plotly online and save to file (not that I know anyone who uses those options much).

bletham:
This looks really nice, thanks a bunch!
As for the return type: this seems to be pretty easy to work with, and we can add to the documentation a reminder to invoke notebook plotting in a notebook. I think it is worth adding this figure to the Quick Start documentation page since people will probably be really interested in using it, especially in a notebook.

Issue #893, opened 2019-03-19
concat() got an unexpected keyword argument 'sort' (pandas<0.23)


lemonsong:
Got concat() got an unexpected keyword argument 'sort' error when use pandas 0.22 with fbprophet 0.4. Should the requirements.txt be updated? 0.23.4 works
The following is the error message:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-23-6798ce6a46da> in <module>()
      1 m, forecast, future = make_forecast(training_data = traing_df,
----> 2                                     changepoint_list = changepoint_list)

<ipython-input-22-850703beebc5> in make_forecast(training_data, changepoint_list)
     31     future = m.make_future_dataframe(periods=200)
     32     print m.train_holiday_names
---> 33     forecast = m.predict(future)
     34 
     35     return m, forecast, future

---------------- moved part -------------
 in predict(self, df)
   1136         df['trend'] = self.predict_trend(df)
   1137         seasonal_components = self.predict_seasonal_components(df)
-> 1138         intervals = self.predict_uncertainty(df)
   1139 
   1140         # Drop columns except ds, cap, floor, and trend

---------------- moved part -------------
 in predict_uncertainty(self, df)
   1338         Dataframe with uncertainty intervals.
   1339         """
-> 1340         sim_values = self.sample_posterior_predictive(df)
   1341 
   1342         lower_p = 100 * (1.0 - self.interval_width) / 2

---------------- moved part -------------

   1291         # Generate seasonality features once so we can re-use them.
   1292         seasonal_features, _, component_cols, _ = (
-> 1293             self.make_all_seasonality_features(df)
   1294         )
   1295 

---------------- moved part -------------

    712 
    713         # Holiday features
--> 714         holidays = self.construct_holiday_dataframe(df['ds'])
    715         if len(holidays) > 0:
    716             features, holiday_priors, holiday_names = (

---------------- moved part -------------
    452                 ]
    453             })
--> 454             all_holidays = pd.concat((all_holidays, holidays_to_add), sort=False)
    455             all_holidays.reset_index(drop=True, inplace=True)
    456         return all_holidays

TypeError: concat() got an unexpected keyword argument 'sort'

bletham:
Thanks for pointing this out. We'll have to either up the requirement to 0.23.4, or see if there is a workaround that would still support the older versions.
(Although 0.23.4 has been out long enough that I don't think it should be a major issue if we have to bump the requirement to that)

[...]

PR #917, opened 2019-04-04
Update requirements.txt

andrealouw:
pandas>=0.23.4 is required to not encounter the following error when passing a holidays argument, since earlier in earlier version pandas the concat does not have a sort argument.

File "/usr/local/lib/python3.6/dist-packages/fbprophet/forecaster.py", line 454, in construct_holiday_dataframe
all_holidays = pd.concat((all_holidays, holidays_to_add), sort=False)
TypeError: concat() got an unexpected keyword argument 'sort'

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,7 +1,7 @@
 Cython>=0.22
 pystan>=2.14
 numpy>=1.10.0
-pandas>=0.20.1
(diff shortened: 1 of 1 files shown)


bletham:
Great, thanks for the fix, this handles #893

PR #924, opened 2019-04-14
DOC: move anaconda section to after pypi section

raybellwaves:
I feel it reads better to have the anaconda section after the pypi section then go into windows and linux


bletham:
Agreed, thanks!

PR #950, opened 2019-04-29
Add logger.warning when falling back to Newton method

jprawiharjo:
This adds logger.warning when we fall back to Newton method.


bletham:
Thanks!

Issue #911, opened 2019-04-02
Feature Request: Add Median Absolute Percent Error to diagnostics


raybellwaves:
Median Absolute Percent Error may be a useful metric for noisy data with multiple holidays.

bletham:
That makes a lot of sense. It would be added to https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py like the others (but would look a little different since it would be a rolling median instead of a rolling mean).

ryankarlos:
@bletham seems that `mdape` is already included in master with tests - should this be closed or is there any further enhancement required ?

bletham:
This is done in Py, but I'll leave it open since for us to translate it into the R too.

ryankarlos:
ok happy to submit a PR

ryankarlos:
take

bletham:
That'd be great!

emmaliden:
It looks to me as if there is some confusion as to if the mdape is an included metric in Python or not. I'm looking in the code in here:

https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py 

In the comments on lines 305-311 and lines 335-339 a list of default performance metrics are given and mdape is not included. On row 345 on the other hand a list called valid_metrics is created and mdape is included in this list. Function mdape is also defined further down (on row 568).

Just from reading the code I'm not sure what the actual behavior would be but when using the package I don't seem to get mdape as a default and I haven't been able to make it work by including it in a list of metrics as an input either.

ryankarlos:
@emmaliden  Are you trying this with prophet v0.6 or the version from master branch ? so `mdape` is definitely included as a metric in python. The docstrings need updating for both the cases which you point out (including the notebook example)- i've pushed a PR including those changes.

```
>>df = pd.read_csv('examples/example_wp_log_peyton_manning.csv')
>>df.head()
 
           ds         y
0  2007-12-10  9.590761
1  2007-12-11  8.519590
2  2007-12-12  8.183677
3  2007-12-13  8.072467
4  2007-12-14  7.893572

>>m = Prophet()
>>m.fit(df)

>>from fbprophet.diagnostics import cross_validation
>>df_cv = cross_validation(m, initial='730 days', period='180 days', horizon = '365 days')
>>df_cv.head()

>>from fbprophet.diagnostics import performance_metrics
>>df_p = performance_metrics(df_cv)
>>df_p.head()
 
[...]

PR #958, opened 2019-05-08
Add median absolute percent error to diagnostics (#911)

louispotok:


bletham:
Thanks for the PR!
Unfortunately I just today pushed a significant change to how CV performance metrics are handled that impacts this PR.

The change is 269133c13342068f7bafcc134c610d085aa52ed6

and basically it makes it so that when computing the rolling average, we first aggregate over horizon and return a single estimate per unique value of horizon. The rolling window is only used for horizons that have <w samples.

I have a comment here that explains why the change was made and how the metrics are computed with the new approach: https://github.com/facebook/prophet/issues/839#issuecomment-462524968

The code that does the aggregation/mean is now quite a bit more complex than it was before, and does have some mean-specific features to it. So you'd need to refactor that to be able to use median.

louispotok:
@bletham Ah! Unfortunate timing :-). 

At a first glance, it looks to me like adding a new function `rolling_median_by_h` would probably be the easiest way to mimic this behavior for medians and would not be too difficult. The `rolling_mean_by_h` function looks to have a lot of mean-specific behavior. Do you agree with that, or would you prefer a refactor to `rolling_agg_by_h` that then takes an argument for `mean` or `median` and shares more functionality?

bletham:
I agree, the function is pretty mean-specific so a separate for the median would make sense.

The thing to be careful of is that all metrics must return results with the same set of horizon values (https://github.com/facebook/prophet/blob/master/python/fbprophet/diagnostics.py#L264). This seemed pretty safe right now since all metrics are computed using the same function, but once we have two functions it would be good to do something a little stricter there to avoid future bugs.

louispotok:
@bletham closed this PR just for git housekeeping on my fork, will reopen once I know what we're doing.

Looking at this code, I don't know that there is a single principled way to do the median. What I understand we do for mean is use a weighted average of the x's with `h == h_i` and the x's with `h ==h_i_-1`, but I don't think that makes sense for median.

For example:

```
x = [1, 2, 4, 5 ,6]
h = [1, 1, 1, 2 ,2]
w = 3
```

To start, if I'm understanding this correctly, the return value should be `[2, 2, 2, a, b]`, right? And a == b?

So now the question is, how do we combine 5.5 (the median for `h=2`) with 2, the median for `h=1`? Does a weighted mean make sense as well? If so, since we need one point from `h=1`, the result should be `(1/3) * (2) + (2/3) * (5.5)`? Is that right?

bletham:
Yeah, what is done for the means is the equivalent of the last, where we are adding a single data point to get it up to 3, but then taking the average over all of the ones we possibly could have taken.

For the mean that is fast because we can work out analytically what that average should be, but for the median it would probably be a bit obnoxious and I don't think is necessary. I think it would be OK to just walk back through `x` and take whatever the first ones are until we reach w, so in this case `[6, 5, 7]`. As an estimator it doesn't change the bias at all, perhaps just has a bit higher variance but I think in most situations there will be enough datapoints in that median that there won't be that much to gain by the avg(median()) approach, and so it would only be worth doing if you found it ended up being easy to implement and fast.

PR #969, opened 2019-05-15
Add figsize parameter to plot methods in Prophet class

pigna90:
`plot` and `plot_components` methods don't have any parameters that can allow you to modify the size of the output plots created by `plot.py` functions.

With this trivial pull request, we can add a really simple and useful functionality :)


bletham:
thanks for the PR, this looks great to me!

PR #970, opened 2019-05-15
Add median absolute percent error to diagnostics (#911)

louispotok:
Follow up to https://github.com/facebook/prophet/pull/958, with the simple implementation discussed there. Haven't added it to the full set of diagnostics yet, wanted @bletham to confirm this makes sense first.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -233,7 +233,7 @@ def performance_metrics(df, metrics=None, rolling_window=0.1):
     -------
     Dataframe with a column for each metric, and column 'horizon'
     """
-    valid_metrics = ['mse', 'rmse', 'mae', 'mape', 'coverage']
+    valid_metrics = ['mse', 'rmse', 'mae', 'mape', 'mdape', 'coverage']
     if metrics is None:
         metrics = valid_metrics
     if len(set(metrics)) != len(metrics):
@@ -343,7 +343,7 @@ def rolling_median_by_h(x, h, w, name):
     df = pd.DataFrame({'x': x, 'h': h})
     grouped = df.groupby('h')
     df2 = grouped.size().reset_index().sort_values('h')
-    hs = df2['h'].values
+    hs = df2['h']
 
     res_h = []
     res_x = []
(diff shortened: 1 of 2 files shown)


python/fbprophet/diagnostics.py
```diff
@@ -319,6 +319,55 @@ def rolling_mean_by_h(x, h, w, name):
...
+    res_h = []
+    res_x = []
+    # Start from the right and work backwards
+    i = len(hs) - 1
+    while i >= 0:
+        h_i = hs[i]
+        xs = grouped.get_group(h_i).x.tolist()
+        j = i - 1
+        while (len(xs) < w) and (j >= 0):
+            # Include points from the previous horizon. All of them if still
```
bletham: The comment here should be updated for this different behavior (include all points from previous horizon)

python/fbprophet/diagnostics.py
```diff
@@ -319,6 +319,55 @@ def rolling_mean_by_h(x, h, w, name):
...
+            # less than w, otherwise just enough to get to w.
+            xs.append(x[j])
```
bletham: Isn't `xs` a list, so this should be `xs.extend(x[j])`?

louispotok: `xs` is a list but `x[j]` is a scalar, so I believe `append` is correct.

bletham: Sorry for the confusion, right after making that comment I realized that I had mistakenly thought x was something different than what it is, so what is here is correct. I wrote up a comment saying that my earlier comment was in error but I guess it didn't post for some reason. Everything looks good here, feel free to proceed!

louispotok:
@bletham Turns out the previous implementation was wrong, as I discovered running the full diagnostics test suite. This one is correct, I believe. Ready for re-review!

bletham:
Sorry for the delay in the review here.
The logic for constructing xs does assume that h is sorted in increasing order. While it is true that they are sorted by h before getting passed in (https://github.com/facebook/prophet/blob/ebb304680d126e5d3a83d4ef9eb1f50cb5965273/python/fbprophet/diagnostics.py#L245), to avoid future bugs in case there are changes there it would be good to explicitly state in the docstring that that function assumes h sorted, or even better it could just sort h (and x correspondingly) at the start. Other than that this looks great and I'll merge.

bletham:
Actually I won't hold this up any longer, I'll merge it and we can add sorting later.

bletham:
Thanks for the contribution!

louispotok:
Oh, didn't mean to go radio silent here! Happy to fix that sorting issue in a follow-up PR. Thanks for your help.

PR #995, opened 2019-05-29
Create dockerfile

joseangel-sc:
This is just a Docker file and a docker-compose, it has the advantage of providing a quick environment to develop.

I ran all the tests and they successfully passed 

Screen Shot 2019-05-28 at 22 51 38

I think a further advance will be to run the test directly from the makefile.


bletham:
Looks good, thanks for the contribution!

PR #1016, opened 2019-06-17
Modifying the way important links are displayed in readme.md

kshithijiyer:


kshithijiyer:
I personally feel that it's better to have hyperlinks instead of copy pasting the URL. If the user is interested he/she'll open it and read it else he/she'll neglect it. And it looks alot clean when you look at the me file.

bletham:
hm I think it is nice to be able to see what the URL is before clicking it to get some sense of where you're going. But when you said "instead of copy pasting the URL" - is it not showing up for you as a hyperlink? At least in my browser I see all of these as hyperlinks with the address as their text. Are you seeing something different?
screenshot

kshithijiyer:
No. I am seeing the same thing, and this looks kind of shabby. I feel if people want to see what's in the link they wouldn't really care where it is stored.

kshithijiyer:
So what should we do finally should we change it or should we keep it as it is?

bletham:
Sorry, I do like to see the URLs and so I'm inclined to leave this out unless it starts to effect usability of the site.

Issue #980, opened 2019-05-22
ValueError: need at least one array to concatenate (custom seasonality)


djrscally:
I am attempting to use random search to optimise hyperparameters. My code is as follows:

```Python
# some preliminary imports and processing, to the effect that 'p' is instantiated as 
# a dataframe of ds and y features and p2 is the last quarter of that dataset sliced off 
# to act as a holdout dataset for validation.

t = 0
limit = 60

m_cps = 0 # changepoint_prior_scale
m_wps = 0 # weekly seasonality prior_scale
m_mps = 0 # monthly seasonality prior_scale
m_qps = 0 # quarterly seasonality prior_scale
m_yps = 0 # yearly seasonality prior_scale

m_wfo = 0 # weekly seasonality fourier_order
m_mfo = 0 # monthly seasonality fourier_order
m_qfo = 0 # quarterly seasonality fourier_order
m_yfo = 0 # yearly seasonality fourier_order

best_score = 9999999999999999


while t < limit:
    print("Onto run {0}. Score to beat: {1}".format(t, best_score))
    cps = np.random.uniform(0,30)
    wps = np.random.uniform(0,30)
    mps = np.random.uniform(0,30)
    qps = np.random.uniform(0,30)
    yps = np.random.uniform(0,30)

    wfo = np.random.randint(0, 50)
    mfo = np.random.randint(0, 50)
    qfo = np.random.randint(0, 50)
    yfo = np.random.randint(0, 50)
    
    m = Prophet(
        growth='linear',
        seasonality_mode='additive',
        daily_seasonality=False,
        weekly_seasonality=False,
        yearly_seasonality=False,
        changepoint_prior_scale=cps
    )

    m.add_country_holidays(country_name='UK')
    m.add_seasonality(name='weekly', period=7, fourier_order=wfo, prior_scale=wps)
    m.add_seasonality(name='monthly', period=30.5, fourier_order=mfo, prior_scale=mps)
    m.add_seasonality(name='quarterly', period=(365.25/4), fourier_order=qfo, prior_scale=qps)
    m.add_seasonality(name='yearly', period=365.25, fourier_order=yfo, prior_scale=yps)
    
    try:
        m.fit(p)
    except ValueError:
        print("""
            m_cps: {0}
            m_wps: {1}
            m_mps: {2}
            m_qps: {3}
            m_yps: {4}
            m_wfo: {5}
            m_mfo: {6}
            m_qfo: {7}
[...]

PR #1020, opened 2019-06-18
[BUG/TST] Add validation to prevent fourier.order <= 0; addresses #980

daikonradish:
In reverse order of importance, this change addresses:
  - #980, by adding validation to ensure that `fourier.order` > 0
  - changes all the tests that `expect_error` using the argument
    `fourier_order`, as the argument is named `fourier.order`, so
    the intent of the test is to validate the holiday names rather
    than to check the use of the correct argument name

Tested these changes by using `devtools::testthat()`.

Resolves: #980

### DIFF
--- a/R/R/prophet.R
+++ b/R/R/prophet.R
@@ -697,7 +697,7 @@ add_regressor <- function(
     mode <- m$seasonality.mode
   }
   if(prior.scale <= 0) {
-    stop("Prior scale must be > 0")
+    stop("Prior scale must be > 0.")
   }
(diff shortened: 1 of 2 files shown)


bletham:
Awesome, thanks a bunch!

bletham:
btw, thanks for all of the help with answering questions too!

bletham:
Great to have another strong member of the community.

daikonradish:
awh no problem! happy to field some of the more... interesting... ones... 😱

PR #1038, opened 2019-06-24
New Plotly Components Plot

Olof-Hojvall:
This PR would add a new Plotly plot of all the components, which intends to match the recently created Plotly plot of the forecast.

To view it in action, check out: https://nbviewer.jupyter.org/github/Olof-Hojvall/prophet/blob/bdc93bd2896488a7cd80a11d66b23b20d3480469/notebooks/Plotly%20Components.ipynb

Suggestion of changes are always appreciated. There are many new features possible using Plotly, but to keep it a bit more simple this first PR mainly seeks to replicated existing matplotlib plot apart from some changes to holidays. 

There are many possible variants of the components plot, making it a bit harder to test. I've not really tested anything else that what's in that notebook, and some help testing other examples would be appreciated.

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -33,6 +33,7 @@ except ImportError:
 
 try:
     import plotly.graph_objs as go
+    from plotly import tools as plotly_tools
 except ImportError:
     logger.error('Importing plotly failed. Interactive plots will not work.')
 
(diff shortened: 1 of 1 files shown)


python/fbprophet/plot.py
```diff
@@ -672,3 +673,303 @@ def plot_plotly(m, fcst, uncertainty=True, plot_cap=True, trend=False, changepoi
...
+    line_width = 2
+
+    range_margin = (fcst['ds'].max() - fcst['ds'].min()) * 0.05
+    range_x = [fcst['ds'].min() - range_margin, fcst['ds'].max() + range_margin]
+
+    text = None
+    mode = 'lines'
+    if name == 'holidays':
+        fcst = fcst[fcst[name] != 0].copy()
+        text = fcst.merge(m.holidays, how='left', on='ds')['holiday']
```
bletham: This runs into issues when there are multiple holidays that share a date, and also has slightly odd behavior with the lower and upper windows on holidays.

You can produce these issues of you generate this plot on the holiday example at the top of https://github.com/facebook/prophet/blob/master/notebooks/seasonality%2C_holiday_effects%2C_and_regressors.ipynb . In this example, each holiday as a +1 holiday window, which means the day after the holiday is also treated as a holiday. What you can see in the first year of the holidays plot is that in these cases "playoff" has the correct text label, but the next day has no text label. This is because it's date is not actually in m.holidays. But the bigger issue is with holidays that share a date. In this case "superbowl" is a subset of "playoff". On these overlapping days, the value being plotted is the sum of the effects. It looks like the text labels are getting mixed up, so that after the first occurrence of "superbowl" some of the "playoff" points are being labeled as "superbowl", and vice versa. 

I think the text labels are _really_ cool and quite useful. So if it is possible to fix them that'd be great, but if you prefer to get this in sooner you could remove them for now and add it back in once it's fixed.

Olof-Hojvall: Well spotted. I've added some new logic that should fix the wrong placements and also include the windows.

bletham:
Awesome, nice work!

PR #1052, opened 2019-07-02
Update hdays.py

seriousmac:
Add holidays in Republic of Korea


bletham:

Issue #1045, opened 2019-06-27
Implement cmdstanpy as optional Python/Stan interface


seanjtaylor:
We'd like to explore making [cmdstanpy](https://github.com/stan-dev/cmdstanpy) (official project from the Stan team) an option for Prophet, making it possible to install/run Prophet without pystan (which is GPLv3). 

Some considerations:

1. **Performance**: will the alternative bindings significantly slow down fitting or prediction, or use additional memory?
2. **Flexibility**: are we currently using any features of PyStan or planning to use features that cmdstanpy will not provide?
3. **Installation**: does having two sets of bindings increase time/effort we spend supporting installation?
4. **Testing**: how do we ensure correctness across two sets of Python bindings?
5. **Code quality**: Does cmdstanpy cause the code to look uglier or become less readable/maintainable?

bletham:
Some additional comments for this:

*During package use*:
The interface between fbprophet and PyStan is fairly small during package use. There are only three things that are done that use PyStan:

(1) We load a pickled `StanModel` that contains the compiled model:
https://github.com/facebook/prophet/blob/e8ddded4fec128191f08be83ae9b5acd7db6f563/python/fbprophet/models.py#L14

(2) We call `StanModel.sampling` and extract the samples:
https://github.com/facebook/prophet/blob/e8ddded4fec128191f08be83ae9b5acd7db6f563/python/fbprophet/forecaster.py#L1116

(3) We call `StanModel.optimizing` and extract the MAP fit:
https://github.com/facebook/prophet/blob/e8ddded4fec128191f08be83ae9b5acd7db6f563/python/fbprophet/forecaster.py#L1131

And that's it!.


*During package install*:

We compile the StanModel and pickle it:
https://github.com/facebook/prophet/blob/e8ddded4fec128191f08be83ae9b5acd7db6f563/python/setup.py#L31

------------

If we want to ultimately swap out PyStan for cmdstanpy then we'll need to be sure that we don't break anyone's setup who is currently using PyStan. I think this means making sure that it will install without issue in an environment in which PyStan installs without issue, and adding it to conda forge.

For the short-term it seems like it might make sense to aim to support both so that if we do want to migrate, that migration can happen a bit more gradually. I think supporting both might not be too terrible since the interface between PyStan and fbprophet is limited to just the few points that I listed above.

daikonradish:
[...]

PR #1083, opened 2019-07-22
Replaced pystan with cmdstanpy

wsuchy:
Related issue: https://github.com/facebook/prophet/issues/1045
As with original implementation we compile a Stan model during the setup phrase. 
CmdStanPy calls the executable each time we fit/sample the model.

The function `stan_to_dict_numpy` is needed to convert output from Stan (fields from arrays as separate columns) into a dense format.

As discussed in another PR it is hard to have stable results across different machines and thus I ended up adding tests with high range of accepted value.


yamamotoseiji:
@seanjtaylor , @bletham , do you have any comments on this PR from Chris?

seanjtaylor:
Sorry for the delay @wsuchy! Had a bunch of travel. I'll give a review today.

python/fbprophet/forecaster.py
```diff
@@ -1097,6 +1097,31 @@ def stan_init():
                 'sigma_obs': 1,
             }
 
+        cmdstanpy_data = {
```
seanjtaylor: Can we either subclass `Forecaster` and add this as an alternative implementation, or delegate this new code path to a different method.

# Option 1

```
class CmdStanPyForecaster(Forecaster):
    def fit(): ...
        # new fit implementation
```
# Option 2

```
class Forecaster: # existing class 
    def __init__(self, ..., stan_backend='pystan'):
        self.stan_backend = stan_backend
    def fit(...):
        if self.stan_backend == 'cmdstanpy':
            self.fit_cmdstanpy(...)
        else:
            self.fit_pystan(...)
```

# Option 3

``` 
class Forecaster: # existing class
    def fit(..., stan_backend='pystan'):
        # similar to option 2
```

python/fbprophet/tests/test_prophet.py
```diff
@@ -28,14 +28,59 @@
...
+    @staticmethod
+    def rmse(predictions, targets):
+        return np.sqrt(np.mean((predictions - targets) ** 2))
+
     def test_fit_predict(self):
+        days = 30
         N = DATA.shape[0]
-        train = DATA.head(N // 2)
-        future = DATA.tail(N // 2)
+        train = DATA.head(N - days)
```
seanjtaylor: Is this mainly to make the tests more reliable?

python/fbprophet/tests/test_prophet.py
```diff
@@ -28,14 +28,59 @@
...
+        forecaster = Prophet()
+        forecaster.fit(train, seed=1237861298, algorithm='Newton')
+        np.random.seed(876543987)
+        future = forecaster.make_future_dataframe(days, include_history=False)
+        future = forecaster.predict(future)
+        # this gives ~ 10.64
+        res = self.rmse(future['yhat'], test['y'])
+        self.assertAlmostEqual(res, 23.44, places=2)
+
+    def test_fit_sampling_predict(self):
```
seanjtaylor: How slow is this test? Is there a way to denote that a test is slow and show only be run optionally?

python/requirements.txt
```diff
@@ -1,5 +1,5 @@
 Cython>=0.22
-pystan>=2.14
+cmdstanpy==0.4
```
seanjtaylor: Not sure we can completely rip out pystan just yet. I see we need to figure out how to make both backends optional so that pystan is optional, though. I haven't seen a Python package with an either/or optional dependency before.

python/setup.py
```diff
@@ -29,15 +29,14 @@
 
 
 def build_stan_model(target_dir, model_dir=MODEL_DIR):
```
seanjtaylor: Just like above, we'll need to keep two versions of this function around.

python/requirements.txt
wsuchy: @seanjtaylor  how about we parametrized install command like this:
`pip install fbprophet --install-option='--with-cmdstanpy'` 
(it seems doable: https://pip.pypa.io/en/stable/reference/pip_install/#per-requirement-overrides)
And then based on the parameter we would create either pystan or cmdstan model.

python/fbprophet/tests/test_prophet.py
wsuchy: Yes, indeed.

python/fbprophet/tests/test_prophet.py
wsuchy: Yes, there is a way of marking the test as slow, yet I am not sure how to pass that around via  setup.py `test` command. I need to look into it.

wsuchy: Sampling test will now run only when `--test-slow` argument present when running tests

python/requirements.txt
seanjtaylor: That looks great to me!

yamamotoseiji:
@seanjtaylor there seem to be two things for us to resolve.  (1) The build system needs to be upgraded to get things to pass (2) cmdstanpy only supports python 3 rather than python 2, is that ok?  What are your thoughts?

python/fbprophet/forecaster.py
```diff
@@ -16,11 +16,12 @@
 
 from fbprophet.diagnostics import prophet_copy
 from fbprophet.make_holidays import get_holiday_names, make_holidays_df
-from fbprophet.models import prophet_stan_model
+from fbprophet.models import *
```
seanjtaylor: Can we import the actual functions instead of `*`?

python/fbprophet/forecaster.py
```diff
@@ -149,9 +159,9 @@ def validate_inputs(self):
             raise ValueError("Parameter 'changepoint_range' must be in [0, 1]")
         if self.holidays is not None:
             if not (
-                isinstance(self.holidays, pd.DataFrame)
-                and 'ds' in self.holidays  # noqa W503
-                and 'holiday' in self.holidays  # noqa W503
+                    isinstance(self.holidays, pd.DataFrame)
```
seanjtaylor: Is this just a style change?

seanjtaylor:
Looks like there's a build issue from Travis, I'm going to looking into that before merging.

bgedik:
@seanjtaylor What happened to this PR? Is the GLP dependency removed?

sirpaddy:
I guess there are quite a few people waiting for this to happen.. Any idea when that could be?

SalmonTimo:
The outstanding travis issues are the same ones mentioned by yamamotoseiji. For (1), it's the maintainers' call if they want to supporting python 2. For (2), the build system needs to have cmdstan installed to use the cmdstanpy backend.

On my fork, I added a call to cmdstanpy's command `install_cmdstan` in the install stage the `.travis.yml` file, but building cmdstan from binary takes over 10 minutes on the agent, so travis to drops the job. I'm guessing we need to include a pre-built cmdstan binary in the build environment or pull in a built binary at runtime.

bletham:
@seanjtaylor do you have any other comments on this? Everything looks good to me and it looks like the right way to be able to run these in parallel and have a gradual transition. It'd be great to get this in for the next release.

It looks like the only remaining issues are build issues. I agree that it'd be ideal to have a parameterized install that allows selecting pytsan or cmdstanpy. For our travis tests here we'll want to test with both. But on platforms like conda forge we could limit it to pystan so that getting cmdstanpy working in conda forge isn't a blocker.

akshayah3:
Hey Guys. Was wondering if there is any progress on this PR. This would really help reduce the time during build time for us as Pystan is taking a lot of resources and time.

bletham:
Can you rebase this PR?
I believe the only outstanding issue is getting unit tests to run in Travis. I'd be fine with merging this now as long as it is able to prevent Travis from trying to run tests that require cmdstanpy. We can run those tests locally before pushing each any new versions to pypi.
The downside is that new PRs might break cmdstanpy from functioning and we wouldn't necessarily know. So master here could end up broken in between releases. But if that's OK with you, then CI isn't strictly necessary to me (we don't have it for the R version) so I'm OK with getting this in.

wsuchy:
@bletham I've rebased the PR. It seems that it should be fairly easy to make Travis job working by doing the following:

1. Dropping Python 2 support
2. Installing CmdStan as a part of the job

Would be happy to do that if it is somehow possible to grant me access to Travis.

bletham:
Great!
As far as I know all of the travis configuration happens here: https://github.com/facebook/prophet/blob/master/.travis.yml

Certainly it looks like Py2 support can be dropped there. My familiarity with travis is pretty superficial so I'm not sure what is involved in installing CmdStan but if there is something that has to happen outside this PR let me know.

bletham:
One other comment - It'd be great if we can get CI tests for cmdstanpy, but regardless I do think it important for now for it to be possible to use fbprophet without having cmdstanpy installed. Just so we don't have to deal with a wave of people who can't get CmdStan working on whatever platform they are currently using fbprophet, and if we ultimately want to make a hard switch that can happen a bit more gradually.

I think it should be pretty easy - in `CmdStanPyBackend.build_model` we can put a try/except around the import of cmdstanpy, and if it fails, just log a warning that the cmdstanpy backend is unavailable. Same thing in load. Sound good? The issue now is that `build_models` in setup.py will run `CmdStanPyBackend.build_model` and so would error out if the import fails.

python/fbprophet/forecaster.py
```diff
@@ -98,7 +94,7 @@ def __init__(
             mcmc_samples=0,
             interval_width=0.80,
             uncertainty_samples=1000,
-            stan_backend='pystan'
+            stan_backend=StanBackendEnum.PYSTAN
```
bletham: Perhaps the better default here is None, and then if None we can select the first available backend from
```
get_backends_from_env()[0]
```
Then if you only have cmdstanpy backend setup, it will work without having to specify it here.
That would also clean up all of the edits to the tests below since they would work without having to specify the backend.

python/setup.py
```diff
@@ -29,8 +28,14 @@
 
 
 def build_models(target_dir):
-    CmdStanPyBackend.build_model(target_dir, MODEL_DIR)
-    PyStanBackend.build_model(target_dir, MODEL_DIR)
+    backends = get_backends_from_env()
```
bletham: To maintain current functionality, we want this to default to pystan if no env variables are specified.

wsuchy: This is exactly how it behaves https://github.com/facebook/prophet/pull/1083/files#diff-85ff8370a138a633c3303428fb5eb0fcR12

wsuchy:
@bletham  I added dependency on the env variable `STAN_BACKEND` that behaves as follows:
- Default `PYSTAN` (if not set)
- Can be set to `CMDSTANPY` or `PYSTAN` or both during the installation phase (setup.py) so the user can select preferred backend.
- Please refer to the changes in `.travis.yml` - the build works fine without `cmdstan` installed (default) 
- TODO: make list of dependencies dynamic based on the choice.

python/fbprophet/forecaster.py
wsuchy: The original intention behind that env var is to serve only as a command switch for installation (setup.py).
How about we stored the default value in a file along with models' binaries instead and thus make the runtime more hermetical?

I would not users' code depend on some hidden env variable.

bletham: Yeah that sounds great to require setting the env variable for cmdstanpy only for install. Ideally after that it would work seamlessly the same as with pystan without having to specify the cmdstanpy backend on model init. What do you think about a strategy where (if no backend is specified) here in __init__ we could try loading both of them, and then if only one is available (catch the IOError when it tries to load the model file) that is the one we use. If both are installed, we could default to pystan. With this approach, if you have only cmdstanpy set up, then it will use cmdstanpy without having to do any extra work. You would only have to specify cmdstanpy if you have both it and pystan installed (and built with fbprophet).

bletham:
The handling for selecting the backend looks great. I think my only remaining question is if it should be `STAN_BACKEND=PYSTAN,CMDSTANPY` in the travis.yml.

wsuchy:
@bletham two questions:
1) While I can add `STAN_BACKEND=PYSTAN,CMDSTANPY` line to the build it will *not* preform tests for both backends (as tests check the default, first backend) so it will actually only confirm that two backends can be installed ad the same time
2) Do you think it would be valuable to modify the content of `install_requires` (to remove either pystan or cmdstanpy dependency) based on selected backend?

bletham:
1. Thanks, I hadn't looked closely enough at the travis.yaml and missed that it was first installing and testing with pystan. It looks great.

2. I guess that would allow one to install without pystan via pip if the env variable was set? That seems like it could be really useful and would make a script for cmdstanpy install cleaner.

In any case, this looks great. A *huge* thanks for your persistence in getting this put together, I know a lot of people will be really excited about this!

Issue #1082, opened 2019-07-22
Speed up prophet by disabling uncertainty estimation (R)


alexgawrilow:
This issue is related to issue #957 by @scasault \
I tried to speed up the calculation of the prophet model. Since I do not care about uncertainties, disabling their calculation is the easiest way to speed up prophet for me.\
Setting the default parameter  `uncertainty_samples=1000` to `uncertainty_samples=0` reduces the calculation time from 
> 8.04 s ± 310 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

 by a factor of almost 2 to: 

> 4.56 s ± 30.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

on the Peyton Manning data set from the [Quick Start](https://facebook.github.io/prophet/docs/quick_start.html#python-api).\
When setting `uncertainty_samples=0` prophet's `predict` method still calls the method `predict_uncertainty` and `predict_seasonal_components` which calculates some upper and lower uncertainty boundaries. Disabling the calculation of the uncertainties reduces the calculation time by another factor of almost 3 to:
> 1.64 s ± 8.44 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

I included some if statements in my prophet source code that check if the boolean value of `uncertainty_samples` is `False`. If so, the calculation of the uncertainties is disabled which reduces the calculation time by a factor of almost 3, otherwise the uncertainties are calculated normally. Do you think this feature is useful for other users? I could open a pull request.

alexgawrilow:
I opened a pull request #1085 that takes care of this issue.

bletham:
This is done for Python. Now #1085 just needs to be translated over to R to add this feature there.

ryankarlos:
@bletham Can i have a go at adding this to R if still required ? I normally use fbprophet with Python but happy to give this a go with R ?

bletham:
@ryankarlos that'd be great! You'll find that the R code closely parallels the Py code, and so it should require the same changes in the same equivalent places.

csetzkorn:
Can you please tell me how to do this R? Thanks.

bletham:
@csetzkorn `m <- prophet(uncertainty.samples = 0)`

PR #1085, opened 2019-07-23
Feature/disable uncertainty

alexgawrilow:
This pull request is related to issue #1082. <br/>
When the user does not care about uncertainties, it is possible to set `uncertainty_samples=0` to speed up the calculations by a factor of about 2 compared to `uncertainty_samples=1000`. Doing this, prophet's `predict` method still calls some methods that calculate an uncertainty. <br/>
I added some `if`-statements that check if the model's `uncertainty_samples` is set to `0` or `False`. In that case, the methods that calculate an uncertainty are not called. This speeds up the calculation by another factor of about 3 making it all in all 6 times faster compared to `uncertainty_samples=1000` when there is no need for uncertainty intervals. <br/>
Since prophet's `plot` module expects uncertainty intervals, I changed that module as well, that no uncertainty intervals are plotted when the model's `uncertainty_samples` equals `0` or `False`.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -101,6 +101,10 @@ def cross_validation(model, horizon, period=None, initial=None):
             msg += 'Consider increasing initial.'
             logger.warning(msg)
 
+    predict_columns = ['ds', 'yhat']
+    if model.uncertainty_samples:
+        predict_columns.extend(['yhat_lower', 'yhat_upper'])
+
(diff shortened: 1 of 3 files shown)


alexgawrilow:
Unfortunately I never worked with R and therefore I am not able to implement the changes in R.

bletham:
This looks good to me too, thanks!
We can just leave #1082 open for now and I or someone else can add the R implementation before the next push.

PR #1087, opened 2019-07-25
Add python 2/3 trove classifiers

proinsias:
Clearly identify this project as supporting python 2 and 3.
This is useful for utility programs like [caniusepython3](https://github.com/brettcannon/caniusepython3#how-do-you-tell-if-a-project-has-been-ported-to-python-3).


bletham:
Looks good, thanks!

PR #1090, opened 2019-07-27
removed unnecessary comma

shahules786:
I removed an unnecessary  ',' at the end of a list.

### DIFF
--- a/python/README.md
+++ b/python/README.md
@@ -29,7 +29,7 @@ On Windows, PyStan requires a compiler so you'll need to [follow the instruction
 
 ## Installation using Docker and docker-compose (via Makefile)
 
-Simply type `make build` and if everything is fine you should be able to `make shell` or alternative jump directly to `make py-shell`. 
+Simply type `make build` and if everything is fine you should be able to `make shell` or alternative jump directly to `make py-shell`.
 
 To run the tests, inside the container `cd python/fbprophet` and then `python -m unittest`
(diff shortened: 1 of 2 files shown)


python/fbprophet/forecaster.py
```diff
@@ -188,7 +188,7 @@ def validate_column_name(self, name, check_holidays=True,
         reserved_names = [
             'trend', 'additive_terms', 'daily', 'weekly', 'yearly',
             'holidays', 'zeros', 'extra_regressors_additive', 'yhat',
-            'extra_regressors_multiplicative', 'multiplicative_terms',
+            'extra_regressors_multiplicative', 'multiplicative_terms'
```
seanjtaylor: We tend to leave these commas in so that if we add new rows it doesn't create a line diff.

python/README.md
```diff
@@ -38,7 +38,7 @@ To run the tests, inside the container `cd python/fbprophet` and then `python -m
 ```python
   >>> from fbprophet import Prophet
   >>> m = Prophet()
-  >>> m.fit(df)  # df is a pandas.DataFrame with 'y' and 'ds' columns
+  >>> m.fit(df)  # df is a pandas.DataFrame with 'Y' and 'DS' as columns
```
seanjtaylor: The columns names must be lower case so this is an incorrect comment.

Issue #1069, opened 2019-07-12
Possible licensing issue: lunardate is GPL


gsbowlin:
Similar to the pystan library being GPLv3 (https://github.com/facebook/prophet/issues/1045), it would be nice if lunardate was replaced with something also not GPLv3, so that prophet could be run without it. 

https://github.com/facebook/prophet/blob/master/python/requirements.txt
https://pypi.org/project/lunardate/0.1.5/

bletham:
I can't seem to find an alternative, but if there is we'd be happy to make a swap. In the meantime, that package is used only in the file https://github.com/facebook/prophet/blob/master/python/fbprophet/hdays.py for built-in country holidays of countries with lunar holidays. If one needed, it would not be too much effort to modify that file to remove those countries and thus remove the dependence on lunardate.

bletham:
#1091 replaces lunardate with LunarCalendar (MIT license)

bletham:
This has been pushed to PyPI.

However, in the process of putting this on conda forge I discovered that LunarCalendar depends on ephem which is LGPL so I'm not sure how much was actually gained here.

PR #1091, opened 2019-07-28
Replace lunardate with LunarCalendar

SalmonTimo:
Related issue: #1069 

Replace instances of `lunardate.LunarDate(year, month, day).toSolar()` with `LunarCalendar.Converter.Lunar2Solar(Lunar(year, month, day)).to_date()`. Remove dependency on lunardate and add dependency on LunarCalendar>=0.0.9.

Switch dependencies from lunardate (GPL) to LunarCalendar (MIT License).


bletham:
Thanks!! This looks good to me too. We will though have to add LunarCalendar to conda forge before this can be pushed (so that the fbprophet package on conda forge will be able to build). I opened an issue in #1104 for it.

bgedik:
Would there be an fbprophet release with the new LunarCalendar?

Issue #891, opened 2019-03-16
FutureWarning: 'ds' is both an index level and a column label


whitehatty:
In `setup_dataframe` line 250 of `forecaster.py`:
```                
        df = df.sort_values('ds')
        df.reset_index(inplace=True, drop=True)
```

If `df` contains index and column with same name the following warning is generated:
```
fbprophet/forecaster.py:250: FutureWarning: 'ds' is both an index level and a column label.
 Defaulting to column, but this will raise an ambiguity error in a future version
```

Warning (and future issue) can be solved by putting either of the following lines before the sorting:
- `df.index.name = None`
- `df = df.rename_axis(None)`
or by calling `reset_index` before and after the call to `sort_values`

bletham:
Thanks for identifying this, there are a few FutureWarnings popping up with the latest pandas that we'll have to go through and clean up for the next version.

whitehatty:
I can open a PR later today if you tell me which approach you prefer better (out of those I listed or something else)

bletham:
That'd be great, I don't have a preference for the approach.

bletham:
Actually I haven't been able to produce this, how did you run into it?

whitehatty:
If your pandas dataframe df has a timestamp column ds (which is what prophet expect) and you call df.resample(..., on=‘ds’) before passing the df to prophet ds becomes an index too.

So the problem is how data is manipulated before passing it to prophet, because prophet does not clean it up when setting up the df for internal use.

t3ja:
is there a work around for this currently?

bletham:
@t3ja does `df.index.name = None` before passing the data into prophet do it?

maverick100:
Can I work on this issue,if its not closed yet ?

whitehatty:
I am sorry I didn’t follow up on this sooner.

Setting the index name to None works, but maybe you want to wait to see if Facebook devs want the change within prophet or let it be on the user.

IMO it is alright as it is literally a one line change.

bletham:
A change inside prophet I think would make sense.

fdroessler:
I hope it was ok to pick this up @maverick100 @whitehatty

whitehatty:
Absolutely, I had completely forgot about it

Nithindinesh24:
df.index.name = None, works fine .thanks @bletham bletham

bletham:
Fix is now pushed to PyPI

PR #1126, opened 2019-08-29
Add test and fix for issue #891

fdroessler:
Picked up issue #891 and implemented a quick fix as discussed in the issue. Also made the reset_index compatible with future pandas version as I understand that they will drop the `inplace` kwarg.


bletham:
Looks great, thanks so much for the contribution!

PR #1163, opened 2019-10-02
plot individual components forecaster.py

akleefel:
Not sure if this already existed somehow but I wanted to make plotting of individual components possible.


bletham:
This was actually left off intentionally. the `Prophet` object has a very large number of methods so we tried to refactor some of them out, just so that file is not so long and the code for the class is easier to understand. We moved most of the plotting to be functions instead of methods, which is only a small amount of extra effort to call vs. a method. Does it seem reasonable for your use case to call the function instead of a method?

PR #1165, opened 2019-10-03
Disable-uncertainty-estimation-in-R 

ryankarlos:
- [x] Fixes issue #1082 
- [x] Tests pass in `testthat` folder

### DIFF
--- a/R/R/diagnostics.R
+++ b/R/R/diagnostics.R
@@ -99,6 +99,11 @@ cross_validation <- function(
         'is larger than initial window. Consider increasing initial.'))
     }
   }
+  
+  predict_columns <- c('ds', 'yhat')
+  if (model$uncertainty_samples){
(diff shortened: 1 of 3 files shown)


ryankarlos:
@bletham Do we need extra tests for these - the python additions don't seem to have any additional ones to test when uncertainty is set to 0 of False

bletham:
This looks great!
Yeah we should definitely add a couple tests. One that does fit/predict and verifies that the column set is correct would be sufficiently to cover all of the changes in the main class, and then a second that does a cross validation would cover the changes there. As you note this should be done for Py as well so I'll merge this now and that can be a separate PR/ pair of PRs.

ryankarlos:
@bletham thanks will create issues and separate PRs for this

bletham:
Awesome, we can use #1082 still for this still if you'd like.

ryankarlos:
@bletham sure, fine by me

Issue #1040, opened 2019-06-25
In cross_validation, STAN arguments aren't passed.


Alexsbuchanan:
For example, say I want to cross validate with 1000 mcmc samples, but only want it on a single process, the n_jobs=1 parameter for STAN is discarded (as far as I can tell) in the fit in the cross_validation function. Can we have a kwarg parameter in cross_validation that lets us pass arguments to STAN in this function?

daikonradish:
do you have a code sample along with desired/undesired behavior?

bletham:
This is a good suggestion. In cross validation, an unfit copy of the model is first made here:
https://github.com/facebook/prophet/blob/df8e39ae15a07ed7635a94a837247582225548fa/python/fbprophet/diagnostics.py#L108

And then it is fit here:
https://github.com/facebook/prophet/blob/df8e39ae15a07ed7635a94a837247582225548fa/python/fbprophet/diagnostics.py#L116

Any arguments to `Prophet()` are stored on the object and are appropriately copied over. However, as you note, kwargs to `fit` (which are passed along to Stan) are not being passed along.

Since `fit` can only be called once, as opposed to passing in the kwargs to cross_validation, I think my preferred solution would be to actually store the fit kwargs on the Prophet object. The copy utility can copy them over, and then in cross validation we would call
```
m.fit(history_c, **m.fit_kwargs)
```
Does that sound reasonable?

Alexsbuchanan:
Hi @bletham 
Thanks for responding!
You're solution is certainly an improvement upon my initial request, I like it.

daikonradish:
Took a stab at this. Seems reasonable, but *gotcha!* The `kwargs` allowed for `optimizing()` and `sampling()` are _not the same. Therefore this approach doesn't work: `Prophet.fit()` passes arguments merrily along into `sampling()`, `Prophet.cross_validation()` passes arguments merrily along into `optimizing()`.

https://pystan.readthedocs.io/en/latest/_modules/pystan/model.html#StanModel.optimizing

Actually I dunno why the two functions have different signatures 🤷‍♂

bletham:
@daikonradish I don't think that should be an issue - cross_validation just calls `fit`, and whether or not `optimizing()` or `sampling()` is called will depend on `m.mcmc_samples`, as specified on model initialization. When the model is copied to make the new cross validation model, that attribute is copied as well:
[...]

PR #1166, opened 2019-10-06
Issue #1040: Adds fit_kwargs

frederickayala:
This pull request adds the functionality requested in #1040 by adding a `fit_kwargs` parameter to the cross_validation method and to the Prophet class. `fit_kwargs` can be used to pass additional parameters for the optimizing and sampling functions of StanModel. I checked https://pystan.readthedocs.io/en/latest/_modules/pystan/model.html to see what parameters can be used.

I am new to contributing in open source projects. I appreciate your patience and guidance in advance.

Thanks!

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -77,7 +77,7 @@ def cross_validation(model, horizon, period=None, initial=None, fit_kwargs=None)
     initial: string with pd.Timedelta compatible style. The first training
         period will begin here. If not provided, 3 * horizon is used.
     fit_kwargs: Additional arguments passed to the optimizing or sampling functions in Stan.
-        If it is not 'None' it replaces the fit_kwargs if the model.
+        If it is not 'None' it replaces the fit_kwargs of the model.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


bletham:
Thanks for the PR! There are a couple things that can be done more simply here.

* We don't need to pass in fit_kwargs on `__init__`, so go ahead and remove that. The only thing we want to do on `__init__` is to specify that it exists, with `self.fit_kwargs = {}`. That can go right in that big chunk of varaibles being set to None or {}.
* In `fit`, the kwargs are passed in as `**kwargs`. The only thing we need to do in `fit` is to store them exactly as they're passed in. So somewhere in `fit` put `self.fit_kwargs = kwargs`. I don't think we need to worry about validating the arguments - Stan will raise an error if they are invalid and the model will not be fit.
* In `prophet_copy`, we can modify the `fit_kwargs` attribute directly instead of passing it in to `__init__`. So just like is done for extra_regressors and seasonalities, you would do `m2.fit_kwargs = deepcopy(m.fit_kwargs)`.
* We don't want `fit_kwargs` to be an input to `cross_validation`. The goal of `cross_validation` is to use the same settings as in `model`, so it should read `fit_kwargs` directly off from `model`. So the only change needed to `cross_validation` is the line `m.fit(history_c)`, which should be changed to `m.fit(history_c, **m.fit_kwargs)`.

Let me know if that all makes sense, and thanks again for tackling this!

bletham:
This is done in https://github.com/facebook/prophet/commit/6e51130f28092099b4bc3b1c7c6349359c60be22

PR #1171, opened 2019-10-08
Option to limit the number of folds in cross validation

matiasbattocchia:
I am willing to forecast thousand of series for a project and I would like to have validation scores for them. It would be enough for me to use the `k` most recent folds in cross validation instead of all available cutoffs, which may increase the number of models to train considerably.

#### TODO

1. [ ] Write a test for this feature. **I need advice on running `test_diagnostics.py`.**
2. [ ] Do the R part.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -16,7 +16,7 @@ import pandas as pd
 logger = logging.getLogger('fbprophet')
 
 
-def generate_cutoffs(df, horizon, initial, period):
+def generate_cutoffs(df, horizon, initial, period, folds=None):
     """Generate cutoff dates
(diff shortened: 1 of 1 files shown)


bletham:
You can get this same behavior by setting `initial` large enough to only allow for `folds` folds. Is the issue here basically that it is obnoxious to have to compute `initial` to get a desired number of folds, so you want to be able to override the number indicated by `initial`?

bletham:
If you install locally your modified version of the package, then you can run the unit tests with 
```
python3 -m unittest test_diagnostics.py
```

matiasbattocchia:
I misunderstood the `initial` argument. Thanks for your response.

PR #1172, opened 2019-10-09
Fix usage of uncertainty.samples for R version

mitchelloharawild:


bletham:
Thanks for the fix! We need to get PR integration testing for R too...

Issue #1158, opened 2019-09-30
[r] dyplot.prophet should pass ... to dygraphs::dySeries(...)


adrfantini:
As per title. Currently, `dyplot.prophet`  does not pass `...` onward to the plotting function, so you cannot set the title, axis labels, etc.
Ideally instead one could be able to do `dyplot.prophet(m, fc, main = "my own title")`.

bletham:
Seems like a useful improvement, go ahead and make a PR and I'll merge!

bletham:
This is pushed to CRAN in v0.6

PR #1177, opened 2019-10-12
pass ... from dyplot.prophet to dygraph::dygraph, addresses #1158

adrfantini:
This PR addresses #1158, passing `...` from `dyplot.prophet` to `dyplot.prophet` to allow for additional parameters such as `main` to be set.

Apparently my text editor removed a bunch of spaces at the end of the lines, hence the large number of deletions. Let me know if this constitutes a problem.

This is one of my first PRs, please be understanding if I screwed something up!

### DIFF
--- a/R/R/plot.R
+++ b/R/R/plot.R
@@ -25,8 +25,8 @@ df_for_plotting <- function(m, fcst) {
 #' @param x Prophet object.
 #' @param fcst Data frame returned by predict(m, df).
 #' @param uncertainty Optional boolean indicating if the uncertainty interval for yhat
-#'  should be plotted, which will only be done if x$uncertainty.samples > 0. 
-#'  Must be present in fcst as yhat_lower and yhat_upper. 
+#'  should be plotted, which will only be done if x$uncertainty.samples > 0.
+#'  Must be present in fcst as yhat_lower and yhat_upper.
 #' @param plot_cap Boolean indicating if the capacity should be shown in the
 #'  figure, if available.
 #' @param xlabel Optional label for x-axis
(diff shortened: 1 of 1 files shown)


bletham:
This LGTM! Were you able to test that it works? You can install the local version with
```
devtools::install('.')
```
from inside the R directory. And then if you could post some code that demonstrates it working that'd be great!

adrfantini:
Mh, after the local install funnily enough I can't use prophet:
```R
# From dyplot.prophet's manual page
library(prophet)
history <- data.frame(
      ds = seq(as.Date('2015-01-01'), as.Date('2016-01-01'), by = 'd'),
      y = sin(1:366/200) + rnorm(366)/10)
m <- prophet(history)
future <- make_future_dataframe(m, periods = 365)
forecast <- predict(m, future)
```
Returns this error:
```
Error in predict_seasonal_components(object, df) :
  object 'uncertainty.samples' not found
```
This also happens with other tests.

bletham:
Ah, sorry about that, the R version on github was briefly broken. We landed the fix on Wednesday (https://github.com/facebook/prophet/commit/f4d997e0acb286507182cb9f5aefeaaa17ff675b). You'll have to rebase your commit on top of that for it to work.

adrfantini:
Mh, I'm still getting (example from `dyplot.prophet`'s docs):
```R
> dyplot.prophet(m, forecast)
Error in uncertainty && x$uncertainty.sampes :
  invalid 'y' type in 'x && y'
```

bletham:
mm, thanks for pointing this out, it looks like there is a typo in that line, `x$uncertainty.sampes` should be `x$uncertainty.samples`. If you make that substitution it should work, and sorry again for the bugs here; that PR seems to have had a few and wasn't tested well before it went in.

janithwanni:
Did anyone merge this PR yet? Because I had the same issue of adding a title into a dygraph.

PR #1178, opened 2019-10-12
Attempt to solve ABI issue with build requirement for stan(and numpy/Cython??)

golf-player:
According to https://github.com/facebook/prophet/issues/401#issuecomment-452484397, it seems like the largest issue with using pyproject.toml is the ABI incompatibility issue. This ensures that the build version of pystan will match the run version.

### DIFF
--- a/python/MANIFEST.in
+++ b/python/MANIFEST.in
@@ -2,6 +2,7 @@ include stan/unix/*.stan
 include stan/win/*.stan
 include LICENSE
 include requirements.txt
+include pyproject.toml
 
 # Ensure in-place built models do not get included in the source dist.
 prune fbprophet/stan_model
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["Cython>=0.22", "numpy>=1.10.0", "pystan>=2.14", "setuptools", "wheel"]
+
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,6 +1,3 @@
-Cython>=0.22
-pystan>=2.14
-numpy>=1.10.0
 pandas>=0.23.4
 matplotlib>=2.0.0
 LunarCalendar>=0.0.9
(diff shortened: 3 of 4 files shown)


golf-player:
I really don't want to deal with the CLA, so someone will have to copy whatever I wrote into another PR assuming this is something you want to do. 

IMO it solves the problem adequately, is overall better UX than requiring pystan to be installed first.

golf-player:
@bletham any thoughts?

bletham:
Thanks for taking a stab at this! Though I'm not sure that it fixes the core issue. Here's how I'm testing:
```
pip3 uninstall pystan
pip3 install pystan==2.18.1
```
in prophet/python directory (after patching this PR),
```
pip3 install -v .
```
In the build environment, it does install the latest version of pystan:
```
Successfully installed Cython-0.29.13 numpy-1.17.2 pystan-2.19.0.0 setuptools-41.4.0 wheel-0.33.6
```
It successfully builds the wheel, and tries to install it. It prints to stderr that the current installed versions don't match those required by the wheel:
```
Successfully built fbprophet
fbprophet 0.5 has requirement Cython==0.29.13, but you'll have cython 0.29.6 which is incompatible.
fbprophet 0.5 has requirement numpy==1.17.2, but you'll have numpy 1.16.4 which is incompatible.
fbprophet 0.5 has requirement pystan==2.19.0.0, but you'll have pystan 2.18.1.0 which is incompatible.
```
But it installs the wheel anyway, without updating the installed version of pystan:
```
Installing collected packages: fbprophet

Successfully installed fbprophet-0.5
Cleaning up...
```
and seems to work.

So from what I can tell the two issues with pyproject.toml remain:
* The build environment downloads and creates a fresh pystan install. When @seanjtaylor did this on his OSX machine, he found that pystan on his system install worked fine but pystan in the build environment did not work because for whatever reason it did not hook up with the C++ compiler properly.
* The wheel is built with a different version of pystan than is installed on the system. In this case it wasn't an issue because pystan 2.18.1 can load models compiled in 2.19, but if I had an older version of pystan installed it would be an issue.

PR #1181, opened 2019-10-14
fixed issue where plotting components with plotly was not working

magnawhale:
I changed only a few lines of code (904-906 & 988-989). In both instances the problematic code was:

`title=go.layout.yaxis.Title(text=name),`

Because `go.layout.yaxis` does not have a method `Title(text=name)`, I had to replace it entirely with the variable `name`. The properly functioning code reads as:

`title=name`

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -902,7 +902,7 @@ def get_forecast_component_plotly_props(m, fcst, name, uncertainty=True, plot_ca
         type='date',
         range=range_x)
     yaxis = go.layout.YAxis(rangemode='normal' if name == 'trend' else 'tozero',
-                            title=go.layout.yaxis.Title(text=name),
+                            title=name,
                             zerolinecolor=zeroline_color)
     if name in m.component_modes['multiplicative']:
         yaxis.update(tickformat='%', hoverformat='.2%')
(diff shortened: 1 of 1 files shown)


yns88:
@magnawhale we had a bug in our CLA form just now, so apologies if you hit errors trying to sign it. You should be able to try again now.

bletham:
I actually haven't run into any error in the components plot. This code works for me:
```
import pandas as pd
from fbprophet import Prophet

df = pd.read_csv('../examples/example_wp_log_peyton_manning.csv')

m = Prophet()
m.fit(df)

future = m.make_future_dataframe(periods=365)
forecast = m.predict(future)

from fbprophet.plot import plot_plotly
import plotly.offline as py
py.init_notebook_mode()

fig = plot_plotly(m, forecast)  # This returns a plotly Figure
py.iplot(fig)

from fbprophet.plot import plot_components_plotly
fig = plot_components_plotly(m, forecast)  # This returns a plotly Figure
py.iplot(fig)
```

Does it work for you or are you doing something different? What version of plotly do you have? Mine is 4.1.1.

magnawhale:
Hi Ben,

I am apparently using plotly 3.1.0, which I suppose might explain why the
original code was not working for me.

I just installed plotly 4.1.1, but now I'm encountering a whole bunch of
other errors (involving Pyximporter and `from fbprophet import Prophet`)
that I will need some time to figure out. I'm relatively new to coding, so
apologies if this has caused any inconvenience. If I figure out what's
going on and am able to test your code with plotly 4.1.1, I'll let you know
if it works for me.

At a bare minimum, at least now I know how to make the methods work with
plotly 3.1!

bletham:
I see, thanks for the heads up that it doesn't work in 3.1. I made a new issue in #1186 to identify what the minimum is. I did previously use this in a plotly 3.x but I don't recall what the actual version was at this point so we'll have to figure out what the minimum is here.

bletham:
I'm going to go ahead and keep things as-is since it is working for plotly 4.x

PR #1191, opened 2019-10-24
Update hdays.py

almeida-marcos:
In Brazil, Republic Proclamation Day is celebrated in November 15th.


bletham:
Thanks for the fix! Must have been a typo.

PR #1203, opened 2019-11-02
TST: Add python test for disabling uncertainty 

### LINKED ISSUE #1082: Speed up prophet by disabling uncertainty estimation (R)

ryankarlos:
- [x] Adds python test for #1085  which fixed #1082

### DIFF
--- a/python/fbprophet/tests/test_diagnostics.py
+++ b/python/fbprophet/tests/test_diagnostics.py
@@ -107,11 +107,12 @@ class TestDiagnostics(TestCase):
 
     def test_cross_validation_uncertainty_disabled(self):
         df = self.__df.copy()
-        m = Prophet(uncertainty_samples=0)
-        m.fit(df)
-        df_cv = diagnostics.cross_validation(
-            m, horizon='4 days', period='4 days', initial='115 days')
-        self.assertListEqual(['ds', 'yhat', 'y', 'cutoff'], df_cv.columns.tolist())
+        for uncertainty in [0, False]:
+            m = Prophet(uncertainty_samples=uncertainty)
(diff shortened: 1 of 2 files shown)


ryankarlos:
As discussed in #1165 adding python test in this PR. Will add R test in this PR or open a new one.

bletham:
Looks great!

PR #1204, opened 2019-11-03
add a little code for make sure last_day is format of datetime with unit is day

binhmuc:
When i use prophet in superset source code for predicting from csv file. The format of datetime of csv file is change from YY-MM-DD to YY-MM-DD HH-MM-SS. So i think the code :

> last_date = self.history_dates.max()

 is so vulnerable.
I change the code and with new code is make sure that the last_date is always get true format for 
"pd.date_range" function.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1494,7 +1494,7 @@ class Prophet(object):
         """
         if self.history_dates is None:
             raise Exception('Model must be fit before this can be used.')
-        last_date = self.history_dates.max()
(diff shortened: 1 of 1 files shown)


bletham:
Could you provide some extra context for this?

While the name of the variable is "last_date", it is actually meant to be a timestamp. With hourly data for instance if you specify freq to be 'H', you'd want this to start actually 1 hour after the end of the history, and this rounding to date would be problematic in that case.

Is there a particular issue you've run into that this fixes? I'm wondering if there might be an alternative approach to fix it.

PR #1212, opened 2019-11-13
Add hyperlinks for easier reference to Quick start

susan-shu-c:


bletham:
Thanks for the PR!
I think adding hyperlinks here is great. However the documentation markdown files are automatically generated from ipython notebooks. That particular documentation file is generated from here: https://github.com/facebook/prophet/blob/master/notebooks/seasonality%2C_holiday_effects%2C_and_regressors.ipynb
This means that this change needs to be made in the ipynb file, not in the md file. Then when the documentation is re-generated next (which either you or I could do as described in https://facebook.github.io/prophet/docs/contributing.html#documentation), the md file will be updated to match the ipynb.

susan-shu-c:
Thanks for the helpful comment, closing and following the correct steps in PR #1299

PR #1215, opened 2019-11-15
Modified text in saturating_forecasts.md

togobingi:
Changed forecast period on line 60 from '3' years to '5' years. 5 years is actually what is being calculated in the example.

### DIFF
--- a/docs/_docs/saturating_forecasts.md
+++ b/docs/_docs/saturating_forecasts.md
@@ -57,7 +57,7 @@ m <- prophet(df, growth = 'logistic')
 m = Prophet(growth='logistic')
 m.fit(df)
 ```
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for fixing this!
However the documentation md files are automatically generated from ipython notebooks (for that page, it's this one: https://github.com/facebook/prophet/blob/master/notebooks/saturating_forecasts.ipynb) so changes to the text like that have to be done in the ipynb file and not the md file. Then next time the documentation is built (https://facebook.github.io/prophet/docs/contributing.html#documentation, but I could do that part) the new md file will be generated from that.

bletham:
notebook fix in https://github.com/facebook/prophet/commit/aeaf0ad2e178092a59d34d9fbe29787eae4b9b57

PR #1217, opened 2019-11-19
Minor style updates to Python implementation (esp forecaster.py)

baogorek:
During my study of Prophet's Python implementation, I've been making made minor changes to style as I would my own code. That's the basis of this pull request; it won't hurt my feelings if you pass on it!

A summary of the changes are:
- Used f-strings in place of .format()...**Update** (2019-11-26): changed to named formats for
   Python 2.7. Should make future conversion to f-strings quicker. 
- Brought all lines in forecaster.py to live within 79 characters
- Made minor modifications to comments, deleted one that I felt was obvious from the code
- Made single quote (') on the outside and double quote on the inside (") the standard. There was
   some inconsistency here so I replicated the version I saw most often.
- Standardized the 'Model has not been fit' exception

Happy to change anything back.

**Note on tests**: Locally on Windows 10, Python 3.6.8, every test passes *except* for `test_logistic_floor` which fails because the translated forecast differs by up to 3 in absolute value, exceeding the threshold of 1. I ran this with the conda installed version of fb_prophet and it passed. After staring at my changes, I don't see how I could have changed the functionality in order to break this one test. My best guess it that `pip install .` is not equivalent to the conda install, but I don't know for sure. **Update** (2019-11-26): repeated same tests on Ubuntu and did not see the same issue with `test_logistic_floor`.

Hope this is useful,
Ben

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -198,24 +198,30 @@ class Prophet(object):
         reserved_names.extend([
             'ds', 'y', 'cap', 'floor', 'y_scaled', 'cap_scaled'])
         if name in reserved_names:
-            raise ValueError('Name "{name}" is reserved.'.format(name=name)
+            raise ValueError(
+                'Name "{name}" is reserved.'.format(name=name)
+            )
         if (check_holidays and self.holidays is not None and
                 name in self.holidays['holiday'].unique()):
             raise ValueError(
                 'Name "{name}" already used for a holiday.'.format(name=name)
+            )
         if (check_holidays and self.country_holidays is not None and
                 name in get_holiday_names(self.country_holidays)):
             raise ValueError(
                 'Name "{name}" is a holiday name in {country_holidays}.'
-                .format(name=name, country_holidays=self.country_holidays))
+                .format(name=name, country_holidays=self.country_holidays)
+            )
         if check_seasonalities and name in self.seasonalities:
             raise ValueError(
                 'Name "{name}" already used for a seasonality.'
-                .format(name=name))
+                .format(name=name)
+            )
         if check_regressors and name in self.extra_regressors:
             raise ValueError(
                 'Name "{name}" already used for an added regressor.'
-                .format(name=name))
+                .format(name=name)
+            )
 
     def setup_dataframe(self, df, initialize_scales=False):
         """Prepare dataframe for fitting or predicting.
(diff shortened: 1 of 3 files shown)


baogorek:
The build's still going but I just realized Python 2.7 isn't going to like those f-strings.

baogorek:
Yeah, the 2.7 build failed on the f strings.  I was thinking about changing all of them into named formatted strings, both for readability and easier conversion to f-strings in the future. Happy to make this change, but please let me know if I should abort! (@bletham, mind weighting in?)

At least the 3.7 build passed (I guess I didn't break `test_logistic_floor`).

python/fbprophet/forecaster.py
```diff
@@ -198,21 +198,30 @@ def validate_column_name(self, name, check_holidays=True,
         reserved_names.extend([
             'ds', 'y', 'cap', 'floor', 'y_scaled', 'cap_scaled'])
         if name in reserved_names:
-            raise ValueError('Name "{}" is reserved.'.format(name))
+            raise ValueError(
+                'Name "{name}" is reserved.'.format(name=name)
```
mmphego: Perhaps rewrite as:
```suggestion
                'Name {name!r} is reserved.'.format(name=name)
```

python/fbprophet/forecaster.py
```diff
@@ -198,21 +198,30 @@ def validate_column_name(self, name, check_holidays=True,
...
-                'Name "{}" is a holiday name in {}.'.format(name, self.country_holidays))
+                'Name "{name}" is a holiday name in {country_holidays}.'
+                .format(name=name, country_holidays=self.country_holidays)
+            )
         if check_seasonalities and name in self.seasonalities:
             raise ValueError(
-                'Name "{}" already used for a seasonality.'.format(name))
+                'Name "{name}" already used for a seasonality.'
+                .format(name=name)
+            )
```
mmphego: See above comment.

bletham:
Thanks for cleaning these up! We haven't been especially rigorous on style to date, and this puts things in a much nicer state.

python/fbprophet/forecaster.py
baogorek: I did not know about this option! Changing the formats now....

mmphego: We learn new things each day. Im glad you learned that today.

python/fbprophet/forecaster.py
baogorek: Sorry it took me so long. Addressed these in [PR 1301](https://github.com/facebook/prophet/pull/1301)

PR #1238, opened 2019-12-02
fix output of valid metrics in performance_metrics

ben-schwen:
If the user inputs metrics which are not valid, he gets to see his **own input** metrics, however I guess its intended for the user to see **real valid metrics**.


bletham:
Yep, that was a bug, thanks for the fix!

PR #1266, opened 2020-01-03
add missing pystan setup dependency

jonbuffington:
During setup, build_stan_model() requires pystan. Build tools that
enforce strict dependency trees fail to install fbprophet. This change
makes the implicit dependency explicit.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -110,6 +110,7 @@ setup(
     license='MIT',
     packages=find_packages(),
     setup_requires=[
+        'pystan>=2.14',
     ],
     install_requires=install_requires,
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR!
The right way to handle the pystan build dependency has been a subject of much discussion (#401). It turns out that adding pystan to setup_requires causes issues with pip that I describe here: https://github.com/facebook/prophet/issues/401#issuecomment-392615740. So while it is correct that pystan is required for setup there doesn't seem to be a great way to express that (other than the current of letting wheel building fail, and then everything in install_requires will be installed).

PR #1274, opened 2020-01-09
remove repeated sentence in docstring

sarikayamerts:
The deleted sentence was just a replicate of the above sentence.


bletham:
Thanks for the fix, looks like a merge-gone-wrong maybe.

PR #1295, opened 2020-01-27
update holidays dependency according to the new structure

sarikayamerts:


bletham:
Thanks for this update! It's unfortunate to have the not-backwards-compatible change in the holidays package. It seems best to make this change along with updating the requirement to `holidays>=0.10.1` and we'll try and get this new version on pip stat.

sarikayamerts:
> Thanks for this update! It's unfortunate to have the not-backwards-compatible change in the holidays package. It seems best to make this change along with updating the requirement to `holidays>=0.10.1` and we'll try and get this new version on pip stat.

What about https://github.com/facebook/prophet/pull/1295/commits/e326372ba8cd47ce5040814d84fa45af5dfc0949 in order to by-pass the version problem? @bletham

bletham:
My feeling is that since things are broken and it will require a version bump on our end to fix it anyway, we might as well just have everyone upgrade to holidays 0.10.1 while we're at it. I can appreciate that for some packages (like pandas) some people may be pinned to an older version for a variety of reasons, but that doesn't seem very likely to me for a package like holidays. Does that seem reasonable to you? If there is a reason to need to continue supporting old versions of holidays then e326372 would definitely be the way to go.

Issue #1304, opened 2020-01-31
AttributeError: 'DatetimeIndex' object has no attribute 'weekday_name'


iplayneon:
model.plot_components(forecast)
It is giving me an attribute error.
pandas.__version__ = '1.0.0'
There is no such method called weekday_name in pandas version 1.0.0

bletham:
Thanks for raising this! It looks like this is a change in pandas 1.0.0, which was released yesterday. We'll have to fix this, but in a backwards compatible way that still works with pandas 0.23.4 (the current min requirement).

dhernandezgt:
it happens to me too.
pandas           : 1.0.0

bletham:
This was fixed in #1304, so you can also install from github to get an immediate fix.

camiloromers:
When you install with conda, the issue still persists.

luxoracle:
> When you install it with conda, the issue still persists.

Indeed, it seems that the 'conda install' method fails to create a proper env for prophet.

ghuname:
Is there a solution for conda install?

camiloromers:
@ghuname what I did is downgrade panda version to pandas==0.25.3 and then I works without problems.

bletham:
As of yesterday, the fix has been pushed to both PyPI and conda forge in the latest version (0.6). So fbprophet==0.6 will work with pandas>=1.0.0, and you shouldn't see this error after upgrading.

If you do see this error, please make sure that you're using the latest version (0.6):
```
import fbprophet
print(fbprophet.__version__)
```

aylinjr:
I am using pandas : 1.0.3
I fixed this problem by replacing the weekday_name to weekday in plot.py file line#288.
I hope this helps.

AVplusTV:
I use in pd version 1.1.3 - **day_name()**  instead **weekday_name**

Laminf2ray:
I used in pd version 1.4.2 **day_name()** to display weekday_name and **month_name()** for months name

PR #1308, opened 2020-02-03
Correction for issue #1304\n\nChange use of weekday_name to day_name …

romulomadu:
…to add consistency with pandas 1.0.


bletham:
Thanks for the fix!

Issue #1310, opened 2020-02-04
2.5x speed-up when using np.percentile rather than np.nanpercentile


jackdent:
I profiled the Prophet `predict` method and noticed that the vast majority of CPU cycles are spent on the four invocations of `np.nanpercentile` function inside `predict_seasonal_components`: 
https://github.com/facebook/prophet/blob/0a270b4a19cafb084e5ed6ee0cf942198d4c913b/python/fbprophet/forecaster.py#L1329-L1334
and `predict_uncertainty`:
https://github.com/facebook/prophet/blob/0a270b4a19cafb084e5ed6ee0cf942198d4c913b/python/fbprophet/forecaster.py#L1410-L1413

Modifying `np.nanpercentile` to `np.percentile` does not change my results, but is 2.5x faster according to the [following notebook](https://gist.github.com/jackdent/11671445054486df17851a5f7997b3a8) I put together (7.2 seconds vs. 2.8 seconds). It looks like the implementation of `np.nanpercentile` is [particularly slow](https://krstn.eu/np.nanpercentile()-there-has-to-be-a-faster-way/).

What do you think of changing the `np.nanpercentile` function to `np.percentile`? Is there ever a situation where Prophet returns `nan`s in a forecast and where we'd want to ignore nan values, rather than failing noisily?

bletham:
The reason it is nanpercentile is because there are rare instances where with MCMC there can be a small number of bad samples that produce NaNs (especially with something like the logistic trend that has an exp() in it). So NaN handling would be nice to keep.

Is it faster to do a check for `isnan` and then only use nanpercentile if so?

jackdent:
@bletham it looks like checking `isnan` and conditionally dispatching to either `np.percentile` or `np.nanpercentile` is much faster: see the following [notebook](https://gist.github.com/jackdent/c119388454539416c009dc60cf639140).

bletham:
That's great, would you be up for making a PR?

jackdent:
Sure thing: https://github.com/facebook/prophet/pull/1311

bletham:
This fix has been pushed to PyPI in the latest version (0.6)

BHIMARAO1984:
could you please help me out to get prediction for next week or a given date instead of head or tail?

PR #1311, opened 2020-02-05
Use np.percentile if array does not contain NaNs

jackdent:
See https://github.com/facebook/prophet/issues/1310 for more details.


jackdent:
@bletham I validated this change by monkey patching numpy in the Jupyter notebooks thet I listed in the original ticket, but didn't test this PR. Any guidance on how to do so would be much appreciated.

bletham:
That's a great improvement, thanks for diagnosing and for the fix!

PR #1331, opened 2020-02-16
Remove deprecated usage of Series.nonzero()

charignon:
#### Summary
Series.nonzero() was deprecated in pandas-dev/pandas#24048.
Using the latest pandas breaks fbprophet.
We can replace .nonzero() use with .to_numpy().nonzero().

#### Test Plan 
- [ ] CI

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -942,7 +942,11 @@ class Prophet(object):
         first = self.history['ds'].min()
         last = self.history['ds'].max()
         dt = self.history['ds'].diff()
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for the PR!
The latest pandas does break fbprophet (to be fixed real soon), but I don't think this is an issue. `dt.values` is a numpy array, and so shouldn't be affected by the deprecation of nonzero() in pandas. Is there an error you're seeing related to this or another reason this is needed?

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.values.html

charignon:
You are right, I tested on an older version and this is no longer relevant after your last commit touching this code, closing.

Issue #1242, opened 2019-12-03
Suppressing `logger.error('Importing plotly failed. Interactive plots will not work.')`


DBCerigo:
I'm not using interactive plots, but the`fbprophet` logger logs the above error at the time `prophet` is imported. This creates unnecessary noise in application logging. 

Is there anyway I can suppress that error? (I still want to propagate other logs from `fbprophet`.)

Link to line for ref: https://github.com/facebook/prophet/blob/83f66dadfd8f127687aa1d6e88cf153fff963fc7/python/fbprophet/plot.py#L38

Rockgecko:
I'd like to do this too.

bletham:
I don't think there's any way to do this right now.

What we can do to make this possible is to change the logger for the plotting to be a different logging. That is, change
https://github.com/facebook/prophet/blob/0a270b4a19cafb084e5ed6ee0cf942198d4c913b/python/fbprophet/plot.py#L16
to be something like
```
logger = logging.getLogger('fbprophet.plot')
```

That would allow for setting different logging levels for the `fbprophet` logger (could stay at default) and the `fbprophet.plot` logger (could be set to critical).

bletham:
This is now pushed to PyPI in v0.6. See the PR in #1332 for an example of how to disable logging for plotting but not other logging.

mthmtc:
the following strange solution helped me

```
class suppress_stdout_stderr(object):
    def __init__(self):
        self.null_fds = [os.open(os.devnull, os.O_RDWR) for x in range(2)]
        self.save_fds = [os.dup(1), os.dup(2)]

    def __enter__(self):
        os.dup2(self.null_fds[0], 1)
        os.dup2(self.null_fds[1], 2)

    def __exit__(self, *_):
        os.dup2(self.save_fds[0], 1)
        os.dup2(self.save_fds[1], 2)
        for fd in self.null_fds + self.save_fds:
            os.close(fd)

with suppress_stdout_stderr(): 
    from fbprophet import Prophet

```

PR #1332, opened 2020-02-16
suppress plotly error logging 

ryankarlos:
closes issue #1242

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -13,7 +13,9 @@ import pandas as pd
 
 from fbprophet.diagnostics import performance_metrics
 
-logger = logging.getLogger('fbprophet')
+logger = logging.getLogger('fbprophet.plot')
+logger.setLevel(logging.CRITICAL)
+
 
 
 try:
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -14,7 +14,6 @@ import pandas as pd
 from fbprophet.diagnostics import performance_metrics
 
 logger = logging.getLogger('fbprophet.plot')
-logger.setLevel(logging.CRITICAL)
 
 
 


bletham:
Thanks for the PR!
Changing the logger to be different than that of the main class is great, but my preference would be to not set the logging level for it inside the package. That would be better left to the user, who can add that line to their script if they want to suppress the logging. Does that work for your use?

bletham:
See my comment in #1336 for how one can set the logging level when using the package.

I do like having these messages by default to avoid confusion with users over why the plot isn't working (especially since plotly is not included in the package requirements and so will not be automatically installed by pip).

ryankarlos:
Thanks @bletham - sure yes that works for me

bletham:
To be clear: the renamed logger I think is great and allows for setting levels independently, so it's just a matter of removing the setLevel command

ryankarlos:
ok just removed the setLevel command

bletham:
Perfect, this works great.
If you do
```
from fbprophet import Prophet
```
without plotly installed, you see the message `Importing plotly failed. Interactive plots will not work.`. If you do
```
import logging
logger = logging.getLogger('fbprophet.plot')
logger.setLevel(logging.CRITICAL)
from fbprophet import Prophet
```
then that message is suppressed, but the logging from the Prophet class is not.

ryankarlos:
Thanks @bletham !

extrememike:
> import logging
> logger = logging.getLogger('fbprophet.plot')
> logger.setLevel(logging.CRITICAL)
> from fbprophet import Prophet

@bletham do you have to rename the logger every time? or what this supposed to be addressed for lates version. Just installed 'fbprophet=0.7.1=py38h7ae7562_0'  but the plotly message is still showing up. Thanks

bletham:
@extrememike you would need to use this code blurb every time (or install plotly):
```
import logging
logger = logging.getLogger('fbprophet.plot')
logger.setLevel(logging.CRITICAL)
from fbprophet import Prophet
```
The only other option would be to locally edit the code to remove the warning.

PR #1336, opened 2020-02-16
Remove logging.root basicConfig on fbprophet import

### LINKED ISSUE #1006: Importing fbprophet sets root logger level to `INFO`
When using the Python version of the library, importing `fbprophet` can sometimes cause the  logging level of the root logger to become `logging.INFO`.

Reproducible example:
```python
import logging
import sys

# notice that the default logging level for the root logger is `WARNING`
assert logging.getLogger().getEffectiveLevel() == logging.WARNING

custom_logger = logging.getLogger('herp-e-derp')
custom_logger.addHandler(logging.StreamHandler(sys.stdout))
custom_logger.info('This message should not show up')

import fbprophet

# after importing fbprophet, the root logger will have level `INFO`
custom_logger.info('We would not expect this message to show up, but it does anyway')
```

This issue has caused some problems for [people using `fbprophet` on DataBricks](https://forums.databricks.com/questions/17799/-infopy4jjava-gatewayreceived-command-c-on-object.html).

I think this is caused by the following lines in [`python/fbprophet/forecaster.py](https://github.com/facebook/prophet/blob/master/python/fbprophet/forecaster.py):
```python
# lines 24-27 in python/fbprophet/forecaster.py
logger = logging.getLogger('fbprophet')
logger.addHandler(logging.NullHandler())
if len(logger.handlers) == 1:
    logging.basicConfig(level=logging.INFO)
```

Perhaps it would be better to set the logging level for the `fbprophet` logger instead of the root logger?
```python
logger = logging.getLogger('fbprophet')
logger.addHandler(logging.NullHandler())
if len(logger.handlers) == 1:
    logger.setLevel(logging.INFO)
```

cdeil:
This PR fixes #1006 . Like the people there, we are struggling with logging configuration and duplicate log messages, because in our ML pipeline we import and use `fbprophet`, and `import fbprophet` does mess with the root logger, and causes duplicate logging messages, and also means that if we call `logging.basicConfig` after `import fbprophet`, it has no effect any more, which is very confusing to new users.

This PR changes the logging to follow the standard recommendation for Python libraries (see [here](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)).  Note that since Python 3.2, there is [logging.lastResort](https://docs.python.org/3/library/logging.html#logging.lastResort) in place, a handler of last resort, which I think to a large degree removes the motivation libraries had in Python 2 to add a handler so that library users will see logging messages by default - this is now the case anyways.

I'll have a look at PyStan now - probably will make the same change request there.

@bletham - Could you please try with this logging setup, and see if behaviour is acceptable for the use cases you have in mind? If no, what exactly is the issue? - Would like to try and resolve it without using the root logger, just the fbprophet logger.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -19,9 +19,7 @@ from fbprophet.models import StanBackendEnum
 from fbprophet.plot import (plot, plot_components)
 
 logger = logging.getLogger('fbprophet')
-logger.addHandler(logging.NullHandler())
-if len(logger.handlers) == 1:
-    logging.basicConfig(level=logging.INFO)
+logger.setLevel(logging.INFO)
 
 
 class Prophet(object):


bletham:
@cdeil Thanks for the PR, this works great.

With typical usage, we get INFO level logging:
```
import pandas as pd
from fbprophet import Prophet

df = pd.read_csv('examples/example_wp_log_peyton_manning.csv')
m = Prophet().fit(df)
```

If we add in
```
import logging
logger = logging.getLogger('fbprophet')
logger.setLevel(logging.ERROR)
# or, logger.setLevel(logging.DEBUG)
```
before fitting the model, then it changes the logging level appropriately.

So this seem to be doing exactly what we want!

PR #1338, opened 2020-02-17
Add progress bar to cross_validation

NikeNano:
The goal with this PR is to add a progress bar to the crossvalidation main for loop. It is not uncommon to run relatively long cross validations runs and therefore it would be great to also have a progress bar showing how much work is left/done.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
+import tqdm 
 from copy import deepcopy
 from functools import reduce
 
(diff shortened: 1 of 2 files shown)


python/fbprophet/diagnostics.py
```diff
@@ -7,6 +7,7 @@
 from __future__ import absolute_import, division, print_function
 
 import logging
+import tqdm
```
NikeNano:
```suggestion
```

NikeNano:
Will test it further and fix issue!

NikeNano:
Example on how it looks here: https://imgur.com/a/ZnFR13F

NikeNano:
> Will test it further and fix issue!

Fixed!

NikeNano:
friendly ping @bletham :)

bletham:
Sorry for the delay, we're trying to get the R version updated on CRAN and running into issues on that, I'll get to these as soon as we can get the R issues worked out.

python/requirements.txt
```diff
@@ -9,3 +9,4 @@ convertdate>=2.1.2
 holidays>=0.9.5
 setuptools-git>=1.2
 python-dateutil>=2.8.0
+tqdm==4.42.1
```
bletham: It'd be best to not pin the version in the requirements here, since some people may not have that level of control over the environment in which they are using the package.
4.42.1 is also a very recent release and so may not be available for all users.

This seems like pretty standard usage, so would this work with tqdm >= 4.36.1? That would be a better requirement on my end certainly.

NikeNano: Sounds good will update!

bletham:
Looks good, thanks!

PR #1353, opened 2020-02-21
Extract stanfit from model object in Py following cmdstanpy refactoring

### LINKED ISSUE #482: Extract stan fit from 'm' object in prophet()
Hi,

First off, thank you for sharing this convenient and valuable tool.

I noticed in prophet() that the return object 'm' has a parameter stan.fit that's default to NULL.

Will it be possible to allow a boolean argument so user can choose to retrieve fitted stan 'fit' (if TRUE)?

Thank you.
Ivan

ryankarlos:
- [x] closes #482

### DIFF
--- a/python/fbprophet/models.py
+++ b/python/fbprophet/models.py
@@ -74,7 +74,7 @@ class CmdStanPyBackend(IStanBackend):
             kwargs['algorithm'] = 'Newton' if stan_data['T'] < 100 else 'LBFGS'
         iterations = int(1e4)
         try:
-            stan_fit = self.model.optimize(data=stan_data,
+            self.stan_fit = self.model.optimize(data=stan_data,
                                            inits=stan_init,
                                            iter=iterations,
                                            **kwargs)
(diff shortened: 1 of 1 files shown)


bletham:
Thanks again for the PR!

This adds the stan fit object as an attribute to the model backend. It would then be accessed like `m.stan_backend.stan_fit`. That seems OK to me, but I have a couple of comments:

- This add `params` as an attribute to bot PyStanBackend and CmdStanBackend by adding the `self.params` logic. Was there a particular reason for this? The params are already being stored on the Prophet class, here, so I'd prefer to not store them on the backend classes:
https://github.com/facebook/prophet/blob/8c48f5b0429d127f6243110ff13f61a322cbf227/python/fbprophet/forecaster.py#L1138-L1141

- Right now the stan_fit is being stored only during a call to `XStanBackend.fit` , but we'll want it for the `sampling` method also. (Actually I think it will be most useful for the sampling method, where the stan fit object contains a bunch of extra information about the sampling).

- All of the current attributes of the stan backends are defined in the base class init here: https://github.com/facebook/prophet/blob/8c48f5b0429d127f6243110ff13f61a322cbf227/python/fbprophet/models.py#L20-L21 . It'd be good to also define the `self.stan_fit` attribute, which could be initialized to None.


Does that all make sense / seem reasonable?

bletham:
Perfect, thank you!

PR #1356, opened 2020-02-22
Running Cross-Validation with Multiprocessing Option

### LINKED ISSUE #1250: Is there a way to run cross validation in parallel (multiprocessing)?
I have quite a few loops to run and running the cross validation in parallel would help quite a bit.
Thanks.

ryankarlos:
- [x] closes #1250 
- [x] added Unittests

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -13,6 +13,7 @@ from functools import reduce
 
 import numpy as np
 import pandas as pd
+from multiprocessing import Pool
 
 logger = logging.getLogger('fbprophet')
 
@@ -58,7 +59,7 @@ def generate_cutoffs(df, horizon, initial, period):
     return list(reversed(result))
 
 
-def cross_validation(model, horizon, period=None, initial=None):
+def cross_validation(model, horizon, period=None, initial=None, multiprocessing=None):
     """Cross-Validation for time series.
 
     Computes forecasts from historical cutoff points. Beginning from
@@ -77,6 +78,8 @@ def cross_validation(model, horizon, period=None, initial=None):
         be done at every this period. If not provided, 0.5 * horizon is used.
     initial: string with pd.Timedelta compatible style. The first training
         period will begin here. If not provided, 3 * horizon is used.
+    multiprocessing: Boolean True. Turns on parallel processing. If not provided,
+    defaults to sequential processing.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


ryankarlos:
Screenshot 2020-02-22 at 16 08 18

NikeNano:
Awesome feature add!

python/fbprophet/diagnostics.py
```diff
@@ -105,43 +110,84 @@ def cross_validation(model, horizon, period=None, initial=None):
...
+            logger.info('Running cross validation in multiprocessing mode')
+            input_df = [[df, model, cutoff, horizon, predict_columns] for cutoff in cutoffs]
+            predicts = pool.starmap(single_cutoff_forecast, input_df)
+    elif multiprocess is None:
+        predicts = []
+        for cutoff in cutoffs:
+            predicts.append(single_cutoff_forecast(df, model, cutoff, horizon, predict_columns))
+    else:
+        raise ValueError(f"{multiprocess} is not a valid assignment to multiprocess argument.Valid "
+                         f"options are 'None' or 'True'")
```
bletham: I think this wouldn't be necessary if it was True/False options for multiprocessing, it'd be more apparent

python/fbprophet/diagnostics.py
```diff
@@ -105,43 +110,84 @@ def cross_validation(model, horizon, period=None, initial=None):
...
+    Parameters
+    ----------
+    df: pd.DataFrame
+        DataFrame with history to be used for single
+        cutoff forecast
+    m: Prophet model.
+    cutoff: string with pd.Timedelta compatible style.
+        Simulated Forecast will start from this date.
+    horizon: string with pd.Timedelta compatible style, e.g., '5 days',
+        '3 hours', '10 seconds'.
```
bletham: At this point this is a pd.Timedelta, not string (it is converted in L90 above)

python/fbprophet/diagnostics.py
```diff
@@ -105,43 +110,84 @@ def cross_validation(model, horizon, period=None, initial=None):
...
+    when evaluating for multiple cutoffs either sequentially or in parallel .
+
+    Parameters
+    ----------
+    df: pd.DataFrame
+        DataFrame with history to be used for single
+        cutoff forecast
+    m: Prophet model.
+    cutoff: string with pd.Timedelta compatible style.
+        Simulated Forecast will start from this date.
```
bletham: This is a pd.Timestamp (that's what generate_cutoffs returns)

python/fbprophet/diagnostics.py
```diff
@@ -105,43 +110,84 @@ def cross_validation(model, horizon, period=None, initial=None):
...
+    # Generate new object with copying fitting options
+    m = prophet_copy(m, cutoff)
+    # Train model
+    history_c = df[df['ds'] <= cutoff]
+    if history_c.shape[0] < 2:
+        raise Exception(
+            'Less than two datapoints before cutoff. '
+            'Increase initial window.'
+        )
+    m.fit(history_c, **m.fit_kwargs)
```
bletham: This actually isn't going to work - you'll notice in the original it is doing
```
m.fit(history_c, **model.fit_kwargs)
```
where `model` is the original model (`m` is the copy).
`fit_kwargs` are set during fitting, and aren't copied over from `model` to `m` by `prophet_copy`.

The purpose of this is that if a particular kwarg was used for fitting the original model (like `algorithm=Newton`) we want to use that same kwarg when fitting the CV model.

That can easily be done here by just changing the name of the arg to this function to be `model` so that we keep track of which was the original model vs. the copy, and using `model.fit_kwargs` here.

ryankarlos: thanks, fixed now

ryankarlos:
@bletham  i've pushed some commits to address the changes.  Also, rebased with master and fixed a conflict - comment here https://github.com/facebook/prophet/pull/1356#discussion_r388675822

bletham:
Looks great, thanks a lot!

bhavanareddy92:
Hi @ryankarlos and @bletham,

I want to use the multiprocessing cv method for testing bu I saw that only the diagnostics file was updated. I was unable to use the mulitprocessing feature. My guess is that it has not been added to the set up file.
How does one make use of this feature?

Thanks

bletham:
@bhavanareddy92 it hasn't been pushed to PyPI yet, so to use this immediately you'll need to download the code from github (master branch) and then install from that. Then you can use it like
```
cross_validation(model, horizon, multiprocess=True)
```
Otherwise it will be available once we push out the next version.

nrchade:
This has been super useful and looking forward to seeing how this will work with dask as well.  I am currently using the version that uses multiprocess but have an issue I'm not sure how to resolve.  How does one safely call this function (cross_validation(model, horizon, multiprocess=True)) inside of a loop without spawning way too many processes?  It took me several computer crashes to figure out that the outer loop wasn't waiting for cross_validation to finish :-(

ryankarlos:
> This has been super useful and looking forward to seeing how this will work with dask as well. I am currently using the version that uses multiprocess but have an issue I'm not sure how to resolve. How does one safely call this function (cross_validation(model, horizon, multiprocess=True)) inside of a loop without spawning way too many processes? It took me several computer crashes to figure out that the outer loop wasn't waiting for cross_validation to finish :-(

Hello, this has now been superseded by PR #1434 which has been merged into master - could you update to most recent version of master and try again

The details on how to run the new version is here where you will have to pass in `parallel=process` to run in multiprocess mode https://github.com/facebook/prophet/blob/master/notebooks/diagnostics.ipynb

NikeNano:
I would be happy to look inte how to utilize dask for utisation if it would be of interest?Have done some work with it.

PR #1380, opened 2020-03-05
Holidays in Belarus + codestyle

dakarpov:
Added official holidays of Belarus

### DIFF
--- a/python/fbprophet/hdays.py
+++ b/python/fbprophet/hdays.py
@@ -1185,7 +1185,7 @@ class Egypt(HolidayBase):
         # The Monday following Orthodox Easter
         name = "Sham El Nessim"
         for offset in range(-1, 2, 1):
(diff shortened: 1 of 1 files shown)


python/fbprophet/hdays.py
```diff
@@ -1430,3 +1432,62 @@ def _populate(self, year):
...
+        name = "Spring and Labour Day"
+        self[date(year, 5, 1)] = name
+
+        # Victory Day
+        name = "Victory Day"
+        self[date(year, 5, 9)] = name
+
+        # Independence Day
+        name = "Independence Day"
+        self[date(year, 6, 3)] = name
```
bletham: It looks like this should be July?

dakarpov: Sorry, fixed + hardcode replaced with an imported variable

bletham:
Great, thank you for the improvement!

Issue #1388, opened 2020-03-12
cmdstanpy version 0.9.5 sampling arg name changes


mitzimorris:
Sampling argument names in CmdStanPy (and CmdStanR) which specify sampling and warmup iterations have been standardized across interfaces - https://github.com/stan-dev/cmdstanpy/issues/221.  Changes which affect fbprophet are:

- `sampling_iters` change to `iter_sampling`
- `warmup_iters` change to `iter_warmup`

bletham:
@mitzimorris Thanks for the heads up!

This will impact us here: https://github.com/facebook/prophet/blob/d9bea6196981bfcedae40525e702d424d26e7cd0/python/fbprophet/models.py#L111

Right now on PyPI it is pinned to 0.4 (https://github.com/facebook/prophet/blob/d9bea6196981bfcedae40525e702d424d26e7cd0/python/requirements.txt#L2), so this won't break any pip installs, but we should fix to the new argument name and then advance the pin to 0.9.5.

PR #1389, opened 2020-03-12
sampling iters arg name, logic changes

mitzimorris:
Changed call to CmdStanPy `sample` method fo CmdStanPy backend to match new argument names, per issue https://github.com/facebook/prophet/issues/1388

Also changed number of sampling iterations - in CmdStan, unlike PyStan and RStan, you specify the number of warmup and sampling iterations as args `num_warmup` and `num_samples`, respectively.  In PyStan and RStan, argument `iter` is the total number of iterations.
It looks like as currently written, the CmdStanPy backend will run twice as many sampling iters as intended.

### DIFF
--- a/python/fbprophet/models.py
+++ b/python/fbprophet/models.py
@@ -57,7 +57,7 @@ class CmdStanPyBackend(IStanBackend):
         model_name = 'prophet.stan'
         target_name = 'prophet_model.bin'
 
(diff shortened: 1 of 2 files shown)


mitzimorris:
this fix is only needed if you upgrade to CmdStanPy 0.9.5 or above.
current build is using CmdStanPy 0.4.0.

PR #1390, opened 2020-03-12
Holidays in United Arab Emirates

sheikheddy:
Added official holidays of the UAE. Our dates for Eid are slightly longer than other countries. I tried to keep the style consistent with what's already there.  We need to add this and Belarus to the docs here: https://facebook.github.io/prophet/docs/seasonality,_holiday_effects,_and_regressors.html#built-in-country-holidays


bletham:
Looks great, thanks!

PR #1391, opened 2020-03-12
Update seasonality,_holiday_effects,_and_regressors.md

sheikheddy:
Korea and Belarus 🇧🇾 were missing. I added UAE 🇦🇪 in my previous pull request so I included that too.

### DIFF
--- a/docs/_docs/seasonality,_holiday_effects,_and_regressors.md
+++ b/docs/_docs/seasonality,_holiday_effects,_and_regressors.md
@@ -285,7 +285,7 @@ m.train_holiday_names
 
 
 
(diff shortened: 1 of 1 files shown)


bletham:
So these documentation markdowns are actually generated from ipython notebooks. This means that the change needs to be made in the notebook, otherwise the change you just made in the markdown will be wiped next time they are re-generated from notebooks. 

The particular notebook is here: https://github.com/facebook/prophet/blob/master/notebooks/seasonality%2C_holiday_effects%2C_and_regressors.ipynb
and you can just make the same change to the corresponding paragraph in the notebook.

sheikheddy:
Thanks, I'll make the changes later today and send them in then. Thanks for pointing out the particular notebook!

bletham:
Updated in the nb in #1440.

PR #1397, opened 2020-03-17
optimize setup.py

### LINKED ISSUE #1396: using setup_requires  at setup.py
when we test `prophet` library using `pip install prophet`,there is always many exception
```
  File "/root/python/prophet/python/fbprophet/models.py", line 207, in build_model
    import pystan
ModuleNotFoundError: No module named 'pystan'
```

the `setup.py` at prophet is not best practices, there is no setup_requires,but `build_models` function need all prophet dependence
I'll fix it in my PR

zjjott:
fix #1396

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 
 # Setuptools distribution folder.
 python/dist/
+python/.eggs/
 
 # Python egg metadata, regenerated from source files by setuptools.
 python/*.egg-info
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,6 +1,6 @@
 Cython>=0.22
 cmdstanpy==0.4
-pystan>=2.14
+pystan>=2.14,<3.0
 numpy>=1.10.0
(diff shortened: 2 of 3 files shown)


zjjott:
:( is there anyone?

bletham:
Thanks for the PR!
As you note in the issue, there is the unfortunate error if one attempts to install fbprophet prior to pystan being installed, since pystan is not specified as a setup requirement.

There is a great deal of discussion on this issue in #401. Unfortunately this approach of adding pytsan to setup_requires does not work because it has some issues with pip. See my comment here for one: https://github.com/facebook/prophet/issues/401#issuecomment-392615740 . What pip recommends doing is using a pyproject.toml file, but that has issues too which I describe here: https://github.com/facebook/prophet/issues/401#issuecomment-452484397 . More generally, it's very important that the setup packages are installed in the same environment as fbprophet, and as far as I am aware there is currently no way with pip to ensure that happens.

PR #1402, opened 2020-03-20
modified cross_validation to allow custom cutoffs

MarcoFusi:
Modified the cross_validation function in diagnostics.py in order to allow users to pass custom cutoff dates to be used instead of auto-generated ones.

A new parameter called cutffs has been added of type: list of pd.Timestamp.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -93,30 +93,33 @@ def cross_validation(model, horizon, period=None, initial=None, multiprocess=Fal
 
     df = model.history.copy().reset_index(drop=True)
     horizon = pd.Timedelta(horizon)
-    # Set period
-    period = 0.5 * horizon if period is None else pd.Timedelta(period)
-    # Identify largest seasonality period
-    period_max = 0.
-    for s in model.seasonalities.values():
-        period_max = max(period_max, s['period'])
-    seasonality_dt = pd.Timedelta(str(period_max) + ' days')
-    # Set initial
-    if initial is None:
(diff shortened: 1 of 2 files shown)


bletham:
Overall this looks great. A few comments on how things can be improved a bit:

1. All of this code for computing the period and the initial does not need to be run if we are specifying cutoffs: https://github.com/facebook/prophet/blob/e9771c862da7f2f834a08808750c85a90b3d3439/python/fbprophet/diagnostics.py#L92-L103
So it can be put inside the `if cutoffs is None` block.

2. This code here is raising a warning if the initial window (that is, the amount of time between the start of the history and the first cutoff) is less than the maximum seasonality period: https://github.com/facebook/prophet/blob/e9771c862da7f2f834a08808750c85a90b3d3439/python/fbprophet/diagnostics.py#L104-L108
It'd be great to display that warning in the case where a user specifies cutoffs that leave the initial block with less length than the maximum seasonality period.

3. We should definitely add a unit test, along the lines of https://github.com/facebook/prophet/blob/master/python/fbprophet/tests/test_diagnostics.py#L121-L126
where we specify a list of cutoffs, and then just make sure that the cutoff dates in df_cv are those specified.

MarcoFusi:
@bletham thank you for your advices. I'm sorry but I haven't been able to actually install the cloned repository on neither AWS Sagemaker nor my laptop so I had to blindly develop and test it against the current build.

bletham:
This is perfect, thanks for the iteration and for pushing this through! This will be very useful.

Issue #1300, opened 2020-01-28
ImportError: cannot import name 'easter' from 'holidays'


kevindesai777:
ImportError: cannot import name 'easter' from 'holidays' (/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/site-packages/holidays/__init__.py)

Install fine using pip3 install fbprophet

nescobar:
This is discussed here: [https://github.com/facebook/prophet/issues/1292](url)

dr-prodigy:
Complete explanation and fix here:
https://github.com/dr-prodigy/python-holidays/issues/277

EmptyG:
dr-prodigy's recommendation (linked above) corrected the issue for me.

kevindesai777:
Thanks guys. Downgrading holidays to 9.1.2 fixed it.

dr-prodigy:
This is not the right way of fixing it. 
Please consider fixing prophet code instead (as explained in my linked issue), in order to be future-proof (holidays supports new countries and features on every release).

kevindesai777:
Thanks @dr-prodigy. Made fixes in the Prophet code. Changes work fine

RSwarnkar:
Is this fixed? I am still getting error with `holidays==0.10.1` and `fbprophet==0.5`.
Please suggest.

RSwarnkar:
> Thanks guys. Downgrading holidays to 9.1.2 fixed it.

There is no holiday==9.1.2 version? 

```
ERROR: Could not find a version that satisfies the requirement holidays==9.1.2 (from versions: 0.1, 0.2, 0.3, 0.3.1, 0.4, 0.4.1, 0.5, 0.6, 0.7, 0.8, 0.8.1, 0.9.3, 0.9.4, 0.9.5, 0.9.6, 0.9.7, 0.9.8, 0.9.9, 0.9.10, 0.9.11, 0.9.12, 0.10.1)
ERROR: No matching distribution found for holidays==9.1.2
```

RSwarnkar:
Ok, below actually worked, at-least for importing fbprophet: 
` pip install holidays==0.9.12`

mpodlasin:
Wait, what do you mean "fix it in prophet code"? Like fork it and maintain project ourselves? 

I don't want to be mean but... that's just absurd! @dr-prodigy

dr-prodigy:
> Wait, what do you mean "fix it in prophet code"? Like fork it and maintain project ourselves?
> 
[...]

PR #1407, opened 2020-03-27
Limit holidays version to 0.9.12

eero-tuhka:
As per #1300, Python prophet breaks with `holidays` version >=0.10.0. The proposed workaround is to manually downgrade the dependency. Ideally we would however set dependencies such that the install works out-of-the-box.

This PR simply limits the `holidays` dependency version to do just that.

Why not do this: maybe just migrate fbprophet to work with 0.10.x of holidays?

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -6,7 +6,7 @@ pandas>=0.23.4
 matplotlib>=2.0.0
 LunarCalendar>=0.0.9
 convertdate>=2.1.2
-holidays>=0.9.5
+holidays>=0.9.5,<=0.9.12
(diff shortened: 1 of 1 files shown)


bletham:
This was actually fixed in 6070b5a3092a6408a6130f3e74f589a6c3dc7460 and d4c460e3dd066093eef0f4767295976ad2a36880, and the current version of fbprophet on pypi (version 0.6) works fine with holidays >=0.10.0. If you're still having issues, could you check what version of fbprophet you are on?

eero-tuhka:
Ok you're right. Turns out we had a stale version in use.

Issue #1381, opened 2020-03-06
Hyperparameter Tuning Snippet


HelioNeves:
I'm happy to share with you my hyperparameter tuning proccess:

Requirements: _fbprophet and tqdm_

**Imports**
```
import logging 
logging.getLogger('fbprophet').setLevel(logging.ERROR)

from itertools import product
from fbprophet import Prophet

from fbprophet.diagnostics import cross_validation
from fbprophet.diagnostics import performance_metrics

from tqdm import tqdm
```

**Holidays**
```
def holidays():
    especial = pd.DataFrame({
    'holiday': 'especial',
    'ds': pd.to_datetime([
                            '2019-01-01', '2019-12-25', 
                            '2020-01-01', '2020-12-25'
                            ]),
    'lower_window': 0,
    'upper_window': 0,
    })

    alta = pd.DataFrame({
    'holiday': 'alta',
    'ds': pd.to_datetime([
                            '2019-03-01', '2019-03-06', '2019-03-07', 
                            '2019-05-11', '2019-06-12', '2019-08-10',
                            '2019-11-29'
                            ]),
    'lower_window': 0,
    'upper_window': 0,
    })

    holidays = pd.concat((especial, alta))
    return holidays
```

**List of Params**
Create a cartesian product based on these key-values with every parameter combination and produces a list for iteration.

```
param_grid = {  'growth': ["linear"], 
                'changepoints': [None], 
                'n_changepoints': [25, 50, 75], 
                'changepoint_range': [0.25, 0.5, 0.75],
                'yearly_seasonality': ["auto"],
                'weekly_seasonality': ["auto"],
                'daily_seasonality': [False],
                'holidays': [holidays],
                'seasonality_mode': ["additive"],
                'seasonality_prior_scale': [10, 50, 100],
                'holidays_prior_scale': [10, 50, 100],
                'changepoint_prior_scale': [0.1, 0.33, 0.66],
                'mcmc_samples': [0],
                'interval_width': [0.25, 0.5, 0.75],
                'uncertainty_samples': [0]
              }

args = list(product(*param_grid.values()))
args
```

**Cross-Validation and Performance**
It produces a perfomance report based on each parameter combination.
```
df_ps = pd.DataFrame()

for arg in tqdm(args):
    m = Prophet(*arg[:7], arg[7](), *arg[8:]).fit(df)
    df_cv = cross_validation(m, initial='1000 days', period='30 days', horizon = '30 days')
[...]

Issue #549, opened 2018-05-31
Creating an "auto" selection mode using AIC/Cross-validation


denizn:
Hello,

I have been trying to choose a model manually currently, however, I would like to do this in a more automated fashion in order to remove hyper care supervision I am currently making for each and every forecast.
I want to try and get scores for all of below combinations and then choose accordingly:
multiplicative seasonality vs. additive seasonality
additive promotions (Exreg) vs multiplicative promotions (Exreg)
additive christmas (Exreg) vs multiplicative christmas (Exreg)
additive easter (Exreg) vs multiplicative easter (Exreg)
linear trend vs. logistic trend

Can you please help me how I could get a "Score" for the fit of the data on the historical data?

AIC,BIC,DIC any of these would be good for me. If you have another suggestion or concerns with above that would be even better.

Deniz

bletham:
There are tools for computing various metrics of predictive power (see `performance_metrics` https://github.com/facebook/prophet/blob/v0.3/notebooks/diagnostics.ipynb), so you could do this with the cross validation by choosing the metric you care about (MAPE perhaps) on the forecast horizon that you care about, and then running the cross_validation and calling performance_metrics (in this case with rolling_window=1 to get a single number to use to compare performance of various settings.)

It would be nice to have some built-in tools for doing this type of selection, although writing the loop for the cross_validation described above should not be too onerous.

deniznoah:
Hi Bletham,

Thank you very much for all the information.

I am testing the library to create 100 models/forecasts at a time with different data. For some of them I would like to use log transformation and for some others I do not.

I am trying to use cross_validation results, but with my understanding is that with log transformation and without log transformation the results would be incomparable.

Can you please kindly tell how I would do this with prophet?

Deniz

shaidams64:
Hi, I was wondering if anyone has worked on adding AIC metric for model selection, and if not, can some one help me out where I should start to implement that? Thanks a lot for your help!

jpmallette:
I believe The following suggestion of @bletham is a good start

> There are tools for computing various metrics of predictive power (see performance_metrics https://github.com/facebook/prophet/blob/v0.3/notebooks/diagnostics.ipynb), so you could do this with the cross validation by choosing the metric you care about (MAPE perhaps) on the forecast horizon that you care about, and then running the cross_validation and calling performance_metrics (in this case with rolling_window=1 to get a single number to use to compare performance of various settings.)

[...]

PR #1434, opened 2020-04-15
Parallel Cross Validation

TomAugspurger:
This changes how parallel cross validation works. It's an alternative to https://github.com/facebook/prophet/pull/1356. I think that a release hasn't been made with #1356, so hopefully this isn't breaking an existing API.

My primary motivation is to allow distributing the work on a cluster with Dask. As a precursor, https://github.com/facebook/prophet/commit/ed3639cfa204728486629f71b1c7ae025bf2c387 refactors things to use `concurrent.futures` for either thread- or process-based parallelism. Users choose parallelism with `parallel=None / "processes" / "threads"`.

https://github.com/facebook/prophet/commit/ed3639cfa204728486629f71b1c7ae025bf2c387 adds in Dask support. It provides a similar API to `concurrent.futures`, with a couple differences I'll annotate inline.

cc @ryankarlos (from #1356) if you have thoughts and @bletham.

---

Some rough timings

```
# parallel=None
CPU times: user 3min 17s, sys: 33.5 s, total: 3min 51s
Wall time: 43.1 s

parallel="threads"
CPU times: user 1min 5s, sys: 3min 27s, total: 4min 33s
Wall time: 45.4 s

# parallel="processes"
CPU times: user 1.18 s, sys: 318 ms, total: 1.49 s
Wall time: 12.2 s

# parallel="dask"
CPU times: user 954 ms, sys: 159 ms, total: 1.11 s
Wall time: 13.3 s
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 
 install:
   - pip install --upgrade pip
-  - pip install -U -r python/requirements.txt
+  - pip install -U -r python/requirements.txt dask dask[dataframe]
 
 script:
   - cd python && python setup.py develop test
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 
 install:
   - pip install --upgrade pip
-  - pip install -U -r python/requirements.txt dask dask[dataframe]
+  - pip install -U -r python/requirements.txt dask[dataframe] distributed
 
 script:
   - cd python && python setup.py develop test
--- a/notebooks/diagnostics.ipynb
+++ b/notebooks/diagnostics.ipynb
@@ -263,7 +263,22 @@
    "source": [
     "In R, the argument `units` must be a type accepted by `as.difftime`, which is weeks or shorter. In Python, the string for `initial`, `period`, and `horizon` should be in the format used by Pandas Timedelta, which accepts units of days or shorter.\n",
     "\n",
-    "Cross-validation can also be run in multiprocessing mode in Python, by setting the `multiprocess` argument to `True`\n",
+    "Cross-validation can also be run in parallel mode in Python, by setting specifying the `parallel` keyword. Three modes are supported\n",
+    "\n",
+    "* `parallel=\"processes\"`\n",
+    "* `parallel=\"threads\"`\n",
+    "* `parallel=\"dask\"`\n",
+    "\n",
+    "For problems that aren't too big, we recommend using `parallel=\"processes\"`. It will achieve the highest performance when the parallel cross validation can be done on a single machine. For large problems, a [Dask](https://dask.org) cluster can be used to do the cross validation on many machines.\n",
+    "\n",
+    "\n",
+    "```python\n",
+    "from dask.distributed import Client\n",
+    "\n",
+    "client = Client()  # connect to the cluster\n",
+    "df_cv = cross_validation(m, initial='730 days', period='180 days', horizon='365 days',\n",
+    "                         parallel=\"dask\")\n",
+    "```\n",
     "\n",
     "The `performance_metrics` utility can be used to compute some useful statistics of the prediction performance (`yhat`, `yhat_lower`, and `yhat_upper` compared to `y`), as a function of the distance from the cutoff (how far into the future the prediction was). The statistics computed are mean squared error (MSE), root mean squared error (RMSE), mean absolute error (MAE), mean absolute percent error (MAPE), and coverage of the `yhat_lower` and `yhat_upper` estimates. These are computed on a rolling window of the predictions in `df_cv` after sorting by horizon (`ds` minus `cutoff`). By default 10% of the predictions will be included in each window, but this can be changed with the `rolling_window` argument."
    ]
@@ -475,7 +490,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.0"
+   "version": "3.7.6"
   }
  },
  "nbformat": 4,
--- a/notebooks/diagnostics.ipynb
+++ b/notebooks/diagnostics.ipynb
@@ -269,7 +269,7 @@
     "* `parallel=\"threads\"`\n",
     "* `parallel=\"dask\"`\n",
     "\n",
-    "For problems that aren't too big, we recommend using `parallel=\"processes\"`. It will achieve the highest performance when the parallel cross validation can be done on a single machine. For large problems, a [Dask](https://dask.org) cluster can be used to do the cross validation on many machines.\n",
+    "For problems that aren't too big, we recommend using `parallel=\"processes\"`. It will achieve the highest performance when the parallel cross validation can be done on a single machine. For large problems, a [Dask](https://dask.org) cluster can be used to do the cross validation on many machines. You will need to [install Dask](https://docs.dask.org/en/latest/install.html) separately, as it will not be installed with `fbprophet`.\n",
     "\n",
     "\n",
     "```python\n",
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -176,8 +176,11 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
             raise ValueError(msg)
 
         iterables = [
-            (df, model, cutoff, horizon, predict_columns)
-            for cutoff in cutoffs
+            itertools.cycle([df]),
+            itertools.cycle([model]),
+            cutoffs,
+            itertools.cycle([horizon]),
+            itertools.cycle([predict_columns]),
         ]
 
         logger.info("Applying in parallel with %s", pool)
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -162,9 +162,9 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
         elif parallel == "dask":
             try:
                 from dask.distributed import get_client
-            except ImportError:
+            except ImportError as e:
                 raise ImportError("parallel='dask' requies the optional "
-                                  "dependency dask.")
+                                  "dependency dask.") as e
             pool = get_client()
             # delay df and model to avoid large objects in task graph.
             df, model = pool.scatter([df, model])
@@ -175,13 +175,10 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
                    "'map' method".format(', '.join(valid)))
             raise ValueError(msg)
 
-        iterables = [
-            itertools.cycle([df]),
-            itertools.cycle([model]),
-            cutoffs,
-            itertools.cycle([horizon]),
-            itertools.cycle([predict_columns]),
-        ]
+        iterables = (
+            (df, model, cutoff, horizon, predict_columns)
+            for cutoff in cutoffs
+        )
 
         logger.info("Applying in parallel with %s", pool)
         predicts = pool.map(single_cutoff_forecast, *iterables)
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -83,7 +83,7 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
         cross-validtation. If not provided works beginning from
         (end - horizon), works backwards making cutoffs with a spacing of period
         until initial is reached.
-    parallel : {None, 'processes', 'threads'}
+    parallel : {None, 'processes', 'threads', 'dask'}
 
         How to parallelize the forecast computation. By default no parallelism
         is used.
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -94,6 +94,8 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
             Note that some operations currently hold Python's Global Interpreter
             Lock, so parallelizing with threads may be slower than training
             sequentially.
+        * 'dask': Parallelize with Dask.
+           This requires that a dask.distributed Client be created.
 
     Returns
     -------
@@ -137,14 +139,24 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
             logger.warning(msg)
 
     if parallel:
-        if parallel not in {"threads", "processes"}:
-            raise ValueError("'parallel' must be one of 'threads' or "
-                             "'processes'. Got '{}' instead".format(parallel))
+        if parallel not in {"threads", "processes", "dask"}:
+            raise ValueError("'parallel' must be one of 'threads', 'processes',"
+                             " or 'dask'. Got '{}' instead".format(parallel))
 
         if parallel == "threads":
             pool = concurrent.futures.ThreadPoolExecutor()
         elif parallel == "processes":
             pool = concurrent.futures.ProcessPoolExecutor()
+        elif parallel == "dask":
+            try:
+                import dask
+                from dask.distributed import get_client
+            except ImportError:
+                raise ImportError("parallel='dask' requies the optional "
+                                  "dependency dask.")
+            pool = get_client()
+            # delay df and model to avoid large objects in task graph.
+            df, model = pool.scatter([df, model])
 
         iterables = [
             itertools.cycle([df]),
@@ -155,6 +167,9 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
         ]
 
         predicts = pool.map(single_cutoff_forecast, *iterables)
+        if parallel == "dask":
+            # convert Futures to DataFrames
+            predicts = pool.gather(predicts)
 
     else:
         predicts = [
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ python:
 
 install:
   - pip install --upgrade pip
-  - pip install -U -r python/requirements.txt
+  - pip install -U -r python/requirements.txt dask[dataframe] distributed
```
TomAugspurger, line 8: I suspect you don't want Dask as a required dependency, so I didn't add it to requirements.txt.

python/fbprophet/diagnostics.py
```diff
@@ -82,9 +83,19 @@ def cross_validation(model, horizon, period=None, initial=None, multiprocess=Fal
...
-        core.
+    parallel : {None, 'processes', 'threads'}
+
+        How to parallelize the forecast computation. By default no parallelism
+        is used.
+
+        * None : No parallelism.
+        * 'processes' : Parallelize with concurrent.futures.ProcessPoolExectuor.
+        * 'threads' : Parallelize with concurrent.futures.ThreadPoolExecutor.
+            Note that some operations currently hold Python's Global Interpreter
```
TomAugspurger, line 94: This may be worth looking into. In theory, thread-based parallelism can be faster since you don't need to serialize the (largeish) DataFrames between processes. But it seems like a decent amount of operations in `single_cutoff_forecast` and what calls seems to hold the GIL.

I'm not sure how feasible it is to improve this. Things like manipulating a DataFrame (slicing / adding columns, etc.) hold the GIL. Things like linear algebra don't.

python/fbprophet/diagnostics.py
```diff
@@ -127,11 +138,39 @@ def cross_validation(model, horizon, period=None, initial=None, multiprocess=Fal
...
+            itertools.cycle([model]),
+            cutoffs,
+            itertools.cycle([horizon]),
+            itertools.cycle([predict_columns]),
+        ]
+
+        predicts = pool.map(single_cutoff_forecast, *iterables)
+        if parallel == "dask":
+            # convert Futures to DataFrames
+            predicts = pool.gather(predicts)
```
TomAugspurger, line 172: Dask's Client.map is a bit different. It returns a List[Future] objects, while concurrent.futures returns the a List of the concrete results.

So we need this extra `gather` here to convert from futures to concrete results (DataFrames in this case)

python/fbprophet/diagnostics.py
```diff
@@ -127,11 +138,39 @@ def cross_validation(model, horizon, period=None, initial=None, multiprocess=Fal
...
+        elif parallel == "dask":
+            try:
+                import dask
+                from dask.distributed import get_client
+            except ImportError:
+                raise ImportError("parallel='dask' requies the optional "
+                                  "dependency dask.")
+            pool = get_client()
+            # delay df and model to avoid large objects in task graph.
+            df, model = pool.scatter([df, model])
```
TomAugspurger, line 159: These objects are a bit large. Doing this pre-scattering helps improve the performance for Dask, and avoids a warning bubbling up to the user about large objects in the task graph.

TomAugspurger:
I'll look into the CI failure. It's passing for me locally with `python -m unittest fbprophet/tests/test_diagnostics.py`, but failing with `python setup.py test`.

ryankarlos:
@TomAugspurger  this looks great ! Now the user has an option for  multi-thread, multiprocessing and on a dask cluster. Yes the multiprocess bit wasnt pushed to pypy yet - so should break anything but @bletham  can confirm. 

One small question - how do the timings vary - between using `parallel=processes` and `parallel=dask` for large number of cutoffs (say if the horizon and period were reduced) - i suppose thread based parallelism would  be slightly faster with smaller dataframes in this case ?
Also, do you think it would be useful for user to choose the number of chunks/chunk size - assuming at the moment the chunk size is determined automatically by number of cores in machine and available memory ?
Im not very familiar with dask distributed (ive only used dask df and dask delayed previously)  - i gather   the url for the performance dashboard is automatically generated for the user ?

notebooks/diagnostics.ipynb
```diff
@@ -263,7 +263,22 @@
...
     "In R, the argument `units` must be a type accepted by `as.difftime`, which is weeks or shorter. In Python, the string for `initial`, `period`, and `horizon` should be in the format used by Pandas Timedelta, which accepts units of days or shorter.\n",
     "\n",
-    "Cross-validation can also be run in multiprocessing mode in Python, by setting the `multiprocess` argument to `True`\n",
+    "Cross-validation can also be run in parallel mode in Python, by setting specifying the `parallel` keyword. Three modes are supported\n",
+    "\n",
+    "* `parallel=\"processes\"`\n",
+    "* `parallel=\"threads\"`\n",
+    "* `parallel=\"dask\"`\n",
+    "\n",
+    "For problems that aren't too big, we recommend using `parallel=\"processes\"`. It will achieve the highest performance when the parallel cross validation can be done on a single machine. For large problems, a [Dask](https://dask.org) cluster can be used to do the cross validation on many machines.\n",
```
ryankarlos, line 272: nit - could we add this line in -  'You will need to install dask 2.14.0 or above separately, as it will not by default be installed with fbprophet'.

python/fbprophet/tests/test_diagnostics.py
```diff
@@ -44,7 +44,7 @@ def test_cross_validation(self):
 
         try:
             from dask.distributed import Client
-            client = Client()
+            client = Client(processes=False)  # noqa
```
ryankarlos, line 47: Is this an issue just with `setup.py test` or does it also impact if the user runs it with `processes=True` - assuming the timings you listed in  are for multiple processes with single threaded workers ?

TomAugspurger:
> Few questions - how do the timings vary...

I'll get some additional timings later today, including some on a real cluster. But the high-level overview is that

1. *Currently* parallel="threads"` will be slower than `parallel="processes"` in essentially every case. In theory thread-based parallelism should be faster when
  - a: The time to serialize data between processes is a large portion of the runtime. So something like `# of forecasts == # of CPU cores`, and a somewhat large DataFrame?
  - b: The GIL is not held too long inside `single_cutoff_forecast`.
2. Dask has no real benefit on a single-machine. It's really only useful on a cluster
3. Dask will be faster when the number of forecasts is much larger than the number of (local) CPU cores. Then the cost of serializing the data will be offset by the increased parallelism.


I've spent a bit of time profiling just what happens in `single_cutoff_forecast`. Nothing concrete yet though. It seems like quite a bit of time is spent in pandas, when we really just need NumPy arrays, but I need to do a bit more digging. Since NumPy generally releases the GIL, we might be able to improve the `threads` case (probably as a followup PR if that's OK).

> do you think it would be useful to provide an option to the user for choosing the number of chunks/chunk size

Yes, that could be useful. What do you think about accepting an initialized Pool?

```python
pool = concurrent.futures.ProcessPoolExecutor(4)
df_cv = cross_validation(..., pool=pool)
```

> but i gather that the url for the performance dashboard is automatically generated for the user ?

Yes. I'm going to get an example at `examples.dask.org`. I have one at https://mybinder.org/v2/gh/TomAugspurger/dask-examples/prophet?filepath=applications%2Fforecasting-with-prophet.ipynb. It's not using the new API proposed here yet, but things should look similar.

python/fbprophet/tests/test_diagnostics.py
TomAugspurger: Yes, it's only failing with `setup.py test`. For context, this typically happens when have a callable that creates processes at the top-level of a file.

```
# file.py
pool = multiprocess.Pool()  # or dask.distributed.Client()
```

I don't really see why it would be happening in a test function.

And those timings were all for one-thread per worker.

ryankarlos:
> Since NumPy generally releases the GIL, we might be able to improve the `threads` case (probably as a followup PR if that's OK).

That sounds reasonable - i may have some time later next week once you have got this PR in, to modify `single_cutoff_forecast` to use primarily Numpy to see if we get any improvement for threads case  - as this is some of the more interesting things I get to do these days. 

> Yes, that could be useful. What do you think about accepting an initialized Pool?
> 
> ```python
> pool = concurrent.futures.ProcessPoolExecutor(4)
> df_cv = cross_validation(..., pool=pool)
> ```

yeh that sounds good -  guess the `parallel` keyword would be redundant then ? 

Im also working on running some HP tuning with this - and maybe can add that to the 
docs as well. Will post some stuff later.

TomAugspurger:
On the topic of accepting callables for custom parallelism

> [@blethem] I do think it'd be good to allow for callables so that people in the future could try new functions without having to push them into the package first, but for the built-in ones having a string alias seems good.

> [@ryankarlos] yeh that sounds good - guess the parallel keyword would be redundant then ?

What do you think about accepting anything with a `.map(Callable, *iterables)`? This isn't quite as simple as just accepting a callable, but I think it's not too bad. I'm trying out an example with a [joblib's](https://joblib.readthedocs.io/en/latest/) loky backend now:

python/fbprophet/diagnostics.py
```diff
@@ -127,11 +138,39 @@ def cross_validation(model, horizon, period=None, initial=None, multiprocess=Fal
...
+
+        iterables = [
+            itertools.cycle([df]),
+            itertools.cycle([model]),
+            cutoffs,
+            itertools.cycle([horizon]),
+            itertools.cycle([predict_columns]),
+        ]
```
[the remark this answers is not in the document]
TomAugspurger: Mmm that doesn't quite work for the `concurrent.futures.map`-style interface. It expects its `*iterables*` to be "grouped" by argument, rather than unpacked.

```python
iterables = [[a1_v1, a1_v2, a1_v3], [a2_v1, a2_v2, a2_v3]]  # want
iterables [(a1_v1, a2_v2), (a1_v2, a2_v2), (a1_v3, a2_v3)]  # unpacked
```

I've pushed up a change that's maybe a bit more readable, assuming you're familiar with how `zip(*iterables)` works.

ryankarlos:
@bletham  @TomAugspurger  ive just experimented with some hyperparameter tuning as an example for users to implement in the docs - using normal for loop and dask.  Like to know your opinions - and how/if it would be useful to go on top of the parallel cross validation being implemented here. (im limited by a not so powerful machine so timings may seem slow but i think with a bit of optimisation in the code here we could get significant speedup espcially trying out a larger number of param combinations)

```
def single_param_cv(history_df, metrics, param_dict):
    m = fbprophet.Prophet(**param_dict)
    m.fit(history_df)
    df_cv = cross_validation(m, initial='2600 days', period='100 days', horizon = '200 days')
    df_p = performance_metrics(df_cv, rolling_window=1)
    df_p['params'] = str(param_dict)
    df_p = df_p.loc[:, metrics]
    return df_p

def param_grid_to_df(**param_dict):
    param_iter = itertools.product(*param_dict.values())
    params =[]
    for param in param_iter:
        params.append(param) 
    params_df = pd.DataFrame(params, columns=list(param_dict.keys()))
    return params_df


def hyperparameter_cv(history_df, params_df, single_cv_callable, pool):
    results = []
    for param in params_df.values:
        param_dict = dict(zip(params_df.keys(), param))
        if pool is None:
            predict = single_cv_callable(history_df, param_dict=param_dict)
            results.append(predict)
        elif isinstance(pool, dask.distributed.client.Client):
            remote_df = pool.scatter(history_df)
            future = pool.submit(single_cv_callable, remote_df, param_dict=param_dict)
            results.append(future)
        else:
            raise ValueError(f'Pool needs to be an instantiated dask distributed client object or None')
    if isinstance(pool, dask.distributed.client.Client):
        results = pool.gather(results)
    results_df = pd.concat(results)
    
    return results_df
```
**Parameter grid settings**
```
param_dict = {  
                'changepoint_prior_scale': [0.1, 1, 10],
                'changepoint_range': [0.8, 0.9]
              }

metrics = ['horizon', 'rmse', 'mape', 'params'] 

params_df = param_grid_to_df(**param_dict)
single_cv_callable = functools.partial(single_param_cv, metrics=metrics)
```
**Dask Distributed Futures API**
```
pool = Client()
pool

%time result_df  = hyperparameter_cv(df, params_df, single_cv_callable, pool=pool)
result_df
best_param = result_df.loc[result_df['rmse'] == min(result_df['rmse']), ['params']]
print(f'The best param combination is {best_param.values[0][0]}')
result_df
```
*Output*

The best param combination is {'changepoint_prior_scale': 0.1, 'changepoint_range': 0.8}
CPU times: user 10.8 s, sys: 1.41 s, total: 12.2 s
Wall time: 1min 36s

horizon | rmse | mape | params
-- | -- | -- | --
200 days | 0.453529 | 0.034922 | {'changepoint_prior_scale': 0.1, 'changepoint_...
200 days | 0.464545 | 0.036444 | {'changepoint_prior_scale': 0.1, 'changepoint_...
200 days | 0.460679 | 0.035948 | {'changepoint_prior_scale': 1.0, 'changepoint_...
200 days | 0.503636 | 0.040299 | {'changepoint_prior_scale': 1.0, 'changepoint_...
200 days | 0.463724 | 0.036518 | {'changepoint_prior_scale': 10.0, 'changepoint...
200 days | 0.505390 | 0.039331 | {'changepoint_prior_scale': 10.0, 'changepoint...

**Normal for loop**

```
%time result_df  = hyperparameter_cv(df, params_df, single_cv_callable, pool=None)
best_param = result_df.loc[result_df['rmse'] == min(result_df['rmse']), ['params']]
print(f'The best param combination is {best_param.values[0][0]}')
result_df
```
*Output*

The best param combination is {'changepoint_prior_scale': 0.1, 'changepoint_range': 0.8}
CPU times: user 2min 14s, sys: 6.43 s, total: 2min 21s
Wall time: 2min 25s

horizon | rmse | mape | params
-- | -- | -- | --
200 days | 0.453529 | 0.034922 | {'changepoint_prior_scale': 0.1, 'changepoint_...
200 days | 0.464545 | 0.036444 | {'changepoint_prior_scale': 0.1, 'changepoint_...
200 days | 0.460679 | 0.035948 | {'changepoint_prior_scale': 1.0, 'changepoint_...
200 days | 0.503636 | 0.040299 | {'changepoint_prior_scale': 1.0, 'changepoint_...
200 days | 0.463724 | 0.036518 | {'changepoint_prior_scale': 10.0, 'changepoint...
200 days | 0.505390 | 0.039331 | {'changepoint_prior_scale': 10.0, 'changepoint...

ryankarlos:
Also the SearchCV in Dask mirrors scikit learn API for cross sectional data so couldnt really diretcly use those methods above https://ml.dask.org/hyper-parameter-search.html

TomAugspurger:
https://github.com/facebook/prophet/issues/549 would be neat!

Those look reasonable. I'll try out your example with more parameters on https://binder.pangeo.io/.

ryankarlos:
> #549 would be neat!
> 
> Those look reasonable. I'll try out your example with more parameters on https://binder.pangeo.io/.

Cool ! Other param settings would be  'seasonality_prior_scale':[0.1, 1, 10.0] and  'holidays_prior_scale':[0.1, 1, 10.0]. @bletham has listed a useful set of recommendations here https://github.com/facebook/prophet/issues/1381#issuecomment-597301902

TomAugspurger:
Just ran those CV snippets on a grid with 108 hyper-parameter combinations. Each single one took ~30 - 60s, depending on the params.

I ran it on a cluster with 40 workers, one core each. It finished in 3.5 minutes (compared with 54 minutes, assuming 30s hyperparam combination). So the scaling seems to be pretty good.

https://nbviewer.jupyter.org/gist/TomAugspurger/8619e223ea42c922cd655d18d6c96424 has the rough notebook.

ryankarlos:
> Just ran those CV snippets on a grid with 108 hyper-parameter combinations. Each single one took ~30 - 60s, depending on the params.
> 
> I ran it on a cluster with 40 workers, one core each. It finished in 3.5 minutes (compared with 54 minutes, assuming 30s hyperparam combination). So the scaling seems to be pretty good.
> 
> https://nbviewer.jupyter.org/gist/TomAugspurger/8619e223ea42c922cd655d18d6c96424 has the rough notebook.

oh wow nice - thats some massive time improvement !  You OK if I submit this as a separate PR for the docs once you get this PR in ? 
One more thing - is there any point in using dask delayed as an example as well - ive found it to be slightly slower https://examples.dask.org/applications/embarrassingly-parallel.html

TomAugspurger:
Yeah, I think keeping the hyperparameter stuff as a followup makes sense.

When you used `dask.delayed` did you also have a distributed `Client()`
active? If not, `delayed` uses the `threaded` scheduler by default which
will suffer from GIL issues.

Otherwise I'm not sure why it would be slower. It shouldn't have much
overhead over the `Client.submit` style, especially since these are such
long-running tasks (10s of seconds).

ryankarlos:
> When you used `dask.delayed` did you also have a distributed `Client()` active? If not, `delayed` uses the `threaded` scheduler by default which will suffer from GIL issues. 

Ahh yes - i had run it before `Client()` was active - so thats probably why !

bletham:
This is great work, thanks for adding it in!

TomAugspurger:
Thanks for the help! Please do ping me when people report issues :)

Issue #1437, opened 2020-04-17
Drop Korea class from hdays.py


chongdae:
Curent lunar caleder for Korea class is based on the "Chinese" lunar calender, which is differ from the "Korean" lunar calender.

Recently the python-holidays package is updated and fixes the problem.
https://github.com/dr-prodigy/python-holidays/pull/294

Please drop Korea class and use the class in python-holidays instead.

cc @seriousran @1kko

seriousran:
@chongdae 

Good point! I'm going to replace Korea class with python-holidays.
Which **released** version of python-holidays includes Korea class?

ref: https://github.com/dr-prodigy/python-holidays/releases

seriousran:
I found it! :)

seriousran:
@chongdae @1kko

I got an error like below. Should I install and import the **korean_lunar_calendar** package?
Would you mind if you add this dependency to python-holidays? It looks better.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-10-838628385e95> in <module>()
      1 import holidays as hdays_part1
      2 
----> 3 holiday_names = getattr(hdays_part1, 'KR')(years=2019).values()

3 frames
/usr/local/lib/python3.6/dist-packages/holidays/countries/korea.py in get_solar_date(self, year, month, day)
    187     # convert lunar calendar date to solar
    188     def get_solar_date(self, year, month, day):
--> 189         self.korean_cal.setLunarDate(year, month, day, False)
    190         return date(self.korean_cal.solarYear, self.korean_cal.solarMonth,
    191                     self.korean_cal.solarDay)

AttributeError: 'KR' object has no attribute 'korean_cal'
```

chongdae:
@seriousran 

The holidays package already has dependency to korean_lunar_calendar.

https://github.com/dr-prodigy/python-holidays/blob/master/setup.py#L40

seriousran:
Thanks! I got it! python-holidays uses 'korea' word and prophet uses 'Korea' and 'KR'. I will try.

bletham:
This fix has been pushed.

PR #1438, opened 2020-04-17
Drop Korea class from hdays.py from issue #1437

seriousran:
To solve this [issue #1437](https://github.com/facebook/prophet/issues/1437) by @chongdae

Korean lunar calendar which is different from Chinese lunar calendar is added python-holidays package recently.
This commit is about dropping the Korean and KR class in `hdays.py` file and `hdays_part1` (=`holidays`) will work for Korean holidays.

```
import fbprophet.hdays as hdays_part2
import holidays as hdays_part1
```


bletham:
It looks like this will also require bumping the version in requirements.txt to holidays>=0.10.2.

Can you post some code that you run to verify that this is actually working? `holidays.KR` isn't working for me:
```
import holidays
holidays.Korea(years=np.arange(1995, 2045))
```
gives
```
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/holidays/countries/korea.py", line 35, in __init__
    HolidayBase.__init__(self, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/holidays/holiday_base.py", line 34, in __init__
    self._populate(year)
  File "/usr/local/lib/python3.7/site-packages/holidays/countries/korea.py", line 61, in _populate
    dt = self.get_solar_date(year, 1, 1)
  File "/usr/local/lib/python3.7/site-packages/holidays/countries/korea.py", line 189, in get_solar_date
    self.korean_cal.setLunarDate(year, month, day, False)
AttributeError: 'Korea' object has no attribute 'korean_cal'
```

That's the call that fbprophet makes:
https://github.com/facebook/prophet/blob/3c69ce3312fcf91d15fd7e56812aa28cabf6f9f2/python/fbprophet/make_holidays.py#L36

seriousran:
@bletham 
Would you mind if you check it again with newer version? I've found an error in `python-holiday` and fixed it. 

https://github.com/dr-prodigy/python-holidays/pull/306

bletham:
Nice work on the upstream fix. It looks like KR holidays will be working then with the next holidays push so we can go ahead and use them. Thanks for the contribution!

PR #1440, opened 2020-04-18
Update seasonality,_holiday_effects,_and_regressors.ipynb

seriousran:
To solve the #1391 Problem


bletham:
Great, thanks! and then once #1438 goes in KR can come off the list.

PR #1452, opened 2020-04-25
Fix Issue with Pickling Models in Python3.6 because of Logger Attribute

### LINKED ISSUE #1361: Error pickling Prophet objects in Python 3.6.
Attempting to pickle a Prophet object in Python 3.6 (3.6.10 to be precise) results in the following error:

`TypeError: can't pickle _thread.RLock objects`

Snippet to reproduce the error:
```python
from fbprophet import Prophet
import pickle

with open('test.pickle', 'wb') as f:
   pickle.dump(Prophet(), f)
```
Trying to save with dill also does not help. The error does not occur in Python 3.7.

chrisbenson32:
Removed logger attribute from IStanBackend to allow pickling in python3.6. Use generic logger like done in plot.py.

This should fix #1361. 

Testing by running python -m unittest

Also tested the below code without error in python3.6
```
from fbprophet import Prophet
import pickle

with open('test.pickle', 'wb') as f:
   pickle.dump(Prophet(), f)
```

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -150,7 +150,7 @@ class Prophet(object):
                 except Exception as e:
                     logger.debug("Unable to load backend %s (%s), trying the next one", i.name, e)
         else:
-            self.stan_backend = StanBackendEnum.get_backend_class(stan_backend)(logger)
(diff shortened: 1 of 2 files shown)


bletham:
This is perfect, thanks for the fix!

PR #1458, opened 2020-04-28
{rstantools}-based compilation

bearloga:
+ cleans up Stan model compilation by switching over to the [{rstantools}-based workflow ](https://mc-stan.org/rstantools/articles/minimal-rstan-package.html)for #1448
+ minor documentation change: {devtools} -> {remotes}, which is better for end-users
+ minor README formatting changes
+ adds RStudio project file which makes it easier for community to get started with contributing to the package

PR #1461, opened 2020-04-28
dplyr 1.0.0 compatibility

hadley:
I'm not sure why a data frame ends up with names and a tibble does not, but it doesn't seem like an important enough problem to investigate in detail, and this is a simple fix.

Branched of off #1458


seanjtaylor:
Thanks @hadley!

PR #1464, opened 2020-04-29
Add example of auto parameter selection to notebook diagnostics

### LINKED ISSUE #549: Creating an "auto" selection mode using AIC/Cross-validation

ryankarlos:
Hello ! 

@bletham  @seanjtaylor  Now that we have dask implementation in - I've added an example to the end of `diagnostics.ipynb` to provide guidance to users for running hyperparameter tuning in parallel based on a previous implementation i had posted in https://github.com/facebook/prophet/pull/1434#issuecomment-619085460

This should close #549  - although from the thread in that issue, not sure if you are still looking for a feature implementation or just a doc example that users can implement themselves. 

Results of timings run on a cluster by @TomAugspurger here https://nbviewer.jupyter.org/gist/TomAugspurger/8619e223ea42c922cd655d18d6c96424

Also mentioned some recommended ranges for parameters suggested by @bletham  here https://github.com/facebook/prophet/issues/1381#issuecomment-597301902
cc @TomAugspurger  if you have anything to add as well regarding the example.


ryankarlos:
The actual ouput is only fot dask mode but Ive also run this with all the different combinations `pool=concurrent.futures.ThreadPoolExecutor()`,`pool=concurrent.futures.ProcessPoolExecutor()` and pool=None and they seemed ok (i thought it would be a bit redundant to run it with those settings as well)  but added instructions for users to run it themselves. 
Maybe worth just running that cell with these combinations as a further confirmation.

notebooks/diagnostics.ipynb
```diff
@@ -468,17 +477,211 @@
...
+    "\n",
+    "\n",
+    "Auto parameter tuning can also be carried out by creating a number of helper functions which can be adapted to run in one of the three parallel modes available: \"processes\", \"threads\" or \"dask\". This can give significant speedup for a large number of paramater combinations. An example implementation in Python is shown below which can be adapted to suit different cases. The `ParameterGrid` class in `sklearn.model_selection`, also provides a nice alternative for creating parameter combinations from the parameter grid. This would need to be installaed and imported separately if required, as it is not included with fbprophet."
```
TomAugspurger, line 483: Might want to add a subsection header just before this.

bletham:
This is really nice, thanks for adding it! I recently typed up a big description of all of the Prophet hyperparameters with thoughts on the tunability of each, and I can add it to the bottom of this too after it merges and then we can have all of the HPO information in one place.

I do have one question / thought though. The parallelization being done here is over parameter values. For each parameter value, the cross validation then uses a number of cutoffs which are evaluated in serial. Alternatively, one could evaluate the parameters in serial and cutoffs in parallel (with the built-in parallelization). Obviously which is better would depend on whether there are more cutoffs or parameter values and how large the pool is. But one upside of the latter (parallelize over cutoffs) is that it is built-in and so would be simpler code. And if you're doing this on a personal computer with 4-8 cores, it probably would be just as fast even with a large number of parameters (because cutoffs would fill up the cores). So maybe it'd be worth describing that first, and then having this example for parallelization over parameters (which could share some code and build on the other). What do you think about that? So basically we'd have two examples for cross validation,
(1) Use built-in parallelization with multiprocessing (simplest / least code example)
(2) Use dask to parallelize over parameter values (best when have access to a large pool)
and the 2nd would build on the 1st and could use a bunch of the same functions.

ryankarlos:
> This is really nice, thanks for adding it! I recently typed up a big description of all of the Prophet hyperparameters with thoughts on the tunability of each, and I can add it to the bottom of this too after it merges and then we can have all of the HPO information in one place.
> 
> I do have one question / thought though. The parallelization being done here is over parameter values. For each parameter value, the cross validation then uses a number of cutoffs which are evaluated in serial. Alternatively, one could evaluate the parameters in serial and cutoffs in parallel (with the built-in parallelization). Obviously which is better would depend on whether there are more cutoffs or parameter values and how large the pool is. But one upside of the latter (parallelize over cutoffs) is that it is built-in and so would be simpler code. And if you're doing this on a personal computer with 4-8 cores, it probably would be just as fast even with a large number of parameters (because cutoffs would fill up the cores). So maybe it'd be worth describing that first, and then having this example for parallelization over parameters (which could share some code and build on the other). What do you think about that? So basically we'd have two examples for cross validation,
> (1) Use built-in parallelization with multiprocessing (simplest / least code example)
> (2) Use dask to parallelize over parameter values (best when have access to a large pool)
> and the 2nd would build on the 1st and could use a bunch of the same functions.

Thats a good point - Ive now split them into two examples one with built-in parallelization and other with parallelization over parameter values. Ive updated the descriptions accordingly - but let me know if anything does not make sense.
Also, created a section at the end for the recommended parameter ranges - if you need to add to it later.

bletham:
This looks great! I'll merge it now, and then at the next pypi push I'll re-build the md so it will update on the documentation page.

PR #1466, opened 2020-04-30
add implementation for constant trend in Python

### LINKED ISSUE #614: Allow constant trend
Hi,

Is there a way to force the trend to be flat? My data has strong daily and weekly seasonality, rather than a strong trend change pattern. Is there a way to make a flat trend on model?

example:
screen shot 2018-07-16 at 11 34 02 am

right now the trend change is too strong, so if I use the current model to predict, prediction will go down (because of the trend). I want to avoid this by making the trend flat. What really matters in my data should only be the seasonalities.

ryankarlos:
Just trying to get a working solution for implementation of constant trend without having to make any edits to stan.

- [x] closes #614
- [x] example outputs
- [x] tests added
- [x] doc updated


ryankarlos:
@bletham I have got a working solution for forcing the trend constant if the user sets `growth=constant`. I havent added an additonal `function in stan` - rather I have just forced the linear trend to return m during the predict steps and setting the k and delta parameters to 0 during the fitting. Let me know what you think - then i can do the same for R.
 Just writing some tests as well.
 
**Linear Trend**

```
m = Prophet(growth='constant')
m.fit(df)
future = m.make_future_dataframe(periods=365)
forecast = m.predict(future)
fig1 = m.plot(forecast)
forecast[['ds', 'yhat', 'yhat_lower', 'yhat_upper']].tail()
```
ds | yhat | yhat_lower | yhat_upper
-- | -- | -- | --
2017-01-15 | 8.206497 | 7.463361 | 8.869975
2017-01-16 | 8.531523 | 7.838554 | 9.263192
2017-01-17 | 8.318930 | 7.644358 | 9.032637
2017-01-18 | 8.151543 | 7.380200 | 8.853538
2017-01-19 | 8.163477 | 7.405988 | 8.890852

image

**Constant Trend**
```
m = Prophet(growth='constant')
m.fit(df)
future = m.make_future_dataframe(periods=365)
forecast = m.predict(future)
fig1 = m.plot(forecast)
forecast[['ds', 'yhat', 'yhat_lower', 'yhat_upper']].tail()
```


ds | yhat | yhat_lower | yhat_upper
-- | -- | -- | --
2017-01-15 | 9.059176 | 8.454865 | 9.673352
2017-01-16 | 9.385230 | 8.768510 | 9.993758
2017-01-17 | 9.173664 | 8.610415 | 9.790425
2017-01-18 | 9.007305 | 8.385269 | 9.618456
2017-01-19 | 9.020267 | 8.440935 | 9.673791



image


```
fig2 = m.plot_components(forecast)
```

image


```
print(f'k {m.params["k"]} \n delta:{m.params["delta"]} \n m:{m.params["m"]} \n beta:{m.params["beta"]} \n sigma:{m.params["sigma_obs"]}')
```
```
k [[0.]] 
 delta:[[0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]
 [0.]] 
 m:[[0.6259358]] 
 beta:[[-0.03000343  0.0482256   0.01001021 -0.00228196  0.01253666  0.01558693
   0.00950642  0.000754    0.00391267 -0.00587152  0.00754138 -0.00524394
   0.00208043 -0.00477661 -0.00410567 -0.00777141 -0.00081501  0.00125614
   0.0018705   0.00698347 -0.01233719 -0.01057211  0.00938554  0.00847058
   0.00088033 -0.00352276]] 
 sigma:[[0.03765628]]
```

bletham:
Thanks for the PR and for tackling this issue. And for the great description and tests!

So this is fitting the usual piecewise linear trend, and then just replacing it with a flat trend that has the same initial offset. Here's an example of what it looks like to use `growth='flat'` on a time series that actually has a lot of trend (this is the retail sales example from the non-daily_data notebook):

flat_trend

What I had originally been describing in the issue was that actually fitting the `m` while keeping `k` fixed at 0. The difference in a time series like the one above is that the best-fit flat trend for the above time series would have the intercept around 325k and would have a high noise term (since all of the change in trend would have to be fit with the variance).

If the time series actually has a flat trend then this will be more or less the same thing. But in cases where the trend isn't flat, I think that what we'd get if we did fit the constant trend model (high predictive uncertainty due to model misspecification) would be much preferred to return a confident but very incorrect prediction. That will require making a change in the Stan code.

ryankarlos:
@bletham  Ahh ok - i had misunderstood from the discussion in the original issue. Ok in that case would you be ok with setting a switch for `trend_indicator` to 2 for the flat trend ?  

The only tricky bit maybe this in `def fit(self, df, **kwargs)`  which I will have to adapt for case with flat trend.
https://github.com/facebook/prophet/blob/4472c2c379af504d633f91e57b6fdcf8fb5f19ae/python/fbprophet/forecaster.py#L1112

bletham:
Yeah that seems good. To avoid repeating the code
https://github.com/facebook/prophet/blob/caaaee8d34f66fd4e8ad140b20c4d26b59a90e0a/python/stan/unix/prophet.stan#L119-L122
three times, it'd probably be better to create a `trend` variable that will be the output of whichever the appropriate trend function is according to `trend_indicator`, and then there can be a single sampling statement for `y` that just uses `trend`. `trend` can be defined in a transformed parameters block.
The particular code you posted for `flat_trend` probably won't work because `m` is a float and that function needs to return a vector, but `rep_row_vector(m, T)` should do the job.

ryankarlos:
> Yeah that seems good. To avoid repeating the code
> 
>   
>     
>       prophet/python/stan/unix/prophet.stan
>     
>     
>         Lines 119 to 122
>       in
>       caaaee8
>     
>     
>     
>     
> 
>         
>           
>              .* (1 + X * (beta .* s_m)) 
>         
> 
>         
>           
>              + X * (beta .* s_a), 
>         
> 
>         
>           
>              sigma_obs 
>         
> 
>         
>           
>            ); 
>         
>     
>   
> 
> 
> three times, it'd probably be better to create a trend variable that will be the output of whichever the appropriate trend function is according to trend_indicator, and then there can be a single sampling statement for y that just uses trend. trend can be defined in a transformed parameters block.
> The particular code you posted for flat_trend probably won't work because m is a float and that function needs to return a vector, but rep_row_vector(m, T) should do the job.

@bletham  I seem to have 

got a working solution now - with the intercept around 325 k for that example you posted.

```
df = pd.read_csv('../examples/example_retail_sales.csv')
m = Prophet(growth='flat')
m.fit(df)
future = m.make_future_dataframe(periods=3652)
fcst = m.predict(future)
fig = m.plot(fcst)
```
flat_trend

I've added this to the documentation - but would prefer a more representative dataset without trend as a good use case for this option.

python/fbprophet/forecaster.py
```diff
@@ -1256,6 +1261,22 @@ def piecewise_logistic(t, cap, deltas, k, m, changepoint_ts):
...
+        t: np.array of times on which the function is evaluated.
+        m: Float initial offset.
+
+        Returns
+        -------
+        Vector y(t).
+        """
+        m_t = m * np.ones_like(t)
+        return m_t
+
```
ryankarlos, line 1279: @bletham  are you ok with this or would you prefer this just done via `piecewise_linear`  with a keyword which just returns `m` if `flat` ?

ryankarlos, line 1279: @bletham are you ok with this or would you prefer this just done via `piecewise_linear`  method with a keyword which just returns `m_t` if flat ?

bletham: Yeah, I like being explicit here.

ryankarlos:
Pushed requested changes now - ive left the notebook example as it is since I don't have another more representative example but feel free to amend if necessary.

bletham:
This looks great, nice work!
Yeah the example is a poor one; I do think it nice for the documentation to also provide best practices, so showing the forecast in a setting where it wouldn't really be appropriate isn't ideal. I'm going to merge this now to avoid another back and forth, but I think I'll just remove the figure from the documentation for now and just have it give the code to for producing the flat trend. And we can add a figure when an appropriate dataset comes along.

yezhengli-Mr9:
Hi @ryankarlos @bletham ,
I personally interested in this because it is loaded in  (https://github.com/facebook/prophet/blob/master/python/fbprophet/models.py#L287):
https://github.com/facebook/prophet/blob/8882c6a3e370736c382db14133f9c66f3b509c30/python/fbprophet/models.py#L280-L287 according to #1466.

Consequentially, I am interested in  **the programming language for** [python/stan/unix/prophet.stan](https://github.com/facebook/prophet/blob/master/python/stan/unix/prophet.stan)? Seemingly not via c++, R, Python, etc.
 
[Refers to #1842.]

fmartin92:
I think I may be missing something here, but if the model is of the form C/(1+exp(-k(t-m))), wouldn't setting k=0 make this constantly C/2, regardless of the value of m?

PR #1474, opened 2020-05-03
Removed redundant line in Prophet Python docstring

above-c-level:
This is fairly straightforward.


bletham:
Thanks for the fix!

PR #1479, opened 2020-05-08
Update requirements.txt

sleepingsuns:
1


bletham:
We had originally left plotly out as a requirement since it is a rather large dependency, and is only needed for optional plotting. As the plotly functionality has expanded (with component plots) it does seem more and more reasonable to have it install by default and then people could remove it from the requirements when installing if they didn't want it. However, it is important for now to keep the dependency no higher than 4.3.0. Unless there is functionality being used that requires 4.5.0, which I don't believe to be the case.

PR #1485, opened 2020-05-10
Remove Python 2 from setup.py

### LINKED ISSUE #1436: Mention that version 0.6 is only Python 3
Since you are using python 3 syntax on the setup.py I assume you just dropped support for python2. Great, but an easy to find notification would be great.

ryankarlos:
- [x] close #1436 
- [x]  passes CI
- [x] add note in docs/readme


bletham:
Perfect, thank you!

PR #1486, opened 2020-05-10
Add partial fit example and caveats to notebook

### LINKED ISSUE #46: periodically data load using partial_fit
In order to fit the model with data from other sources, like Prometheus, would be nice to have partial_fit implemented, so new data can be added periodically.

ryankarlos:
- [x]  closes #46
- [x] docs updated 

@bletham No idea how I never came across #46 before but this partial fitting would have made my life a lot easier :) . I've added this in the docs for now with some of your comments regarding caveats - would you like this to be implemented in the codebase as well ? I'm thinking this should be easy enough for users who follow the docs to carry out themselves - otherwise, I can do that in a follow up PR once this is merged - what do you think ?


ryankarlos:
@bletham  I've added this in the docs for now with some of your comments regarding caveats - would you like this to be implemented in the codebase as well ? I'm thinking this should be easy enough for users who follow the docs to carry out themselves - otherwise,I can do that in a follow up PR once this is merged - what do you think ?

bletham:
I think having a doc for this is good and what you have put seems really nice to me, though I don't think the quickstart is quite the right place for it. The quickstart is meant to be a really simple example that outlines the functionality that most people will be using before diving into the more advanced / detailed parts in later pages.

I'm not really sure which doc page would be right for this though. I've been thinking of adding a page for adding some of these discussions in the issues that seem to come up regularly. This is a great example of one of those. I haven't quite figured out what to call it though since it would likely end up as a grab-bag of not-necessarily-related topics. Something like "Additional topics"?.. Do you have any thoughts on that? and on starting a new nb page with this example?

ryankarlos:
Yes I agree - i was struggling to find the appropriate notebook to add this in as the others were focussing on more specialist specific topics - whilst this is more a performance improvement option .... ill create a new page with `Additional Topics` 

> I think having a doc for this is good and what you have put seems really nice to me, though I don't think the quickstart is quite the right place for it. The quickstart is meant to be a really simple example that outlines the functionality that most people will be using before diving into the more advanced / detailed parts in later pages.
> 
> I'm not really sure which doc page would be right for this though. I've been thinking of adding a page for adding some of these discussions in the issues that seem to come up regularly. This is a great example of one of those. I haven't quite figured out what to call it though since it would likely end up as a grab-bag of not-necessarily-related topics. Something like "Additional topics"?.. Do you have any thoughts on that? and on starting a new nb page with this example?

bletham:
Great, this will be a good place to put some answers for all the questions that come up regularly in the issues page. Thanks! I'll build the markdown and add it to the site with the next version push.

PR #1492, opened 2020-05-14
Extend model to other distributions

rostyboost:
This is a ticket, but opened as a PR because there's working code associated for illustrative purposes.
I couldn't find an existing ticket with a similar proposal, apologies if it has already been proposed in the past.

I'd like to gauge the interests of the maintainers to the idea of extending the original gaussian assumption of the model to cover other distributions.

As a motivating example, this PR shows how to replace the `y ~ normal(...)` with a:
```stan
  y ~ gamma(
      log(1 + exp(linear_trend(k_a, m_a, delta, t, A, t_change)
      .* (1 + X * (alpha .* s_m))
      + X * (alpha .* s_a))),
      log(1 + exp(linear_trend(k_b, m_b, delta, t, A, t_change)
      .* (1 + X * (beta .* s_m))
      + X * (beta .* s_a))));
```
which is particularly suited for time series that have the following 2 properties:
1. positive
2. asymmetric right-skewed risk: higher probability of encountering a rare "high" event than a rare "low" event

This leads to much nicer fits and uncertainty estimates for time series that follow these properties:
image
A _lot_ of time series fall into this class.

The codebase could for example be extended to allow arbitrary 1 or 2 parameters distributions from the laundry list of those available in [Stan](https://mc-stan.org/docs/2_23/functions-reference/continuous-distributions.html).


bletham:
These are some really cool results, and thanks for sharing the code. I opened an issue in #1500 to discuss!

PR #1505, opened 2020-05-20
Add check for R to Travis CI

ryankarlos:


.travis.yml
```diff
@@ -1,17 +1,30 @@
...
-  - rm -rf fbprophet/stan_model
-  - wget https://github.com/stan-dev/cmdstan/releases/download/v2.22.1/cmdstan-2.22.1.tar.gz -O /tmp/cmdstan.tar.gz > /dev/null
-  - tar -xvf /tmp/cmdstan.tar.gz -C /tmp > /dev/null
-  - make -C /tmp/cmdstan-2.22.1/ build > /dev/null
-  - CMDSTAN=/tmp/cmdstan-2.22.1 STAN_BACKEND=CMDSTANPY python setup.py develop test
+    - language: r
+      r:
+        - devel
+      install:
+        - R -e 'install.packages("remotes")'
```
ryankarlos, line 25: for some reason it wasn't picking up remotes or devtools automatically unless i installed it - i though the r devel version should already have that installed and available in namespace ?

bletham:
Hey, sorry for the slow reply on this. I've had a few conference submissions dominate my life the last couple weeks. Now that the deadline is passed...

Overall this looks great. I think just running the unit tests would be good though, not the whole `R CMD check` since it does take a really long time to run. The way I see it running the unit tests will make sure that the package isn't broken by a PR and that the master branch remains in an installable state. R CMD check will be necessary for pushing the package to CRAN, but those sorts of issues we can just clean up in batch when it comes time for a new package push.

Formatting with black is on my todo list for the summer (along with figuring out what it's equivalent is for R). In CI, I guess it raises an error and doesn't autocorrect? I don't think we need to raise lint errors in CI, that seems like it could get really annoying / be a barrier for people who haven't done much development before. The value of black is that it's really easy to fix lint issues, so it seems like merging PRs with lint issues and then just periodically running black on master is a low-friction way of keeping things in good shape. As for type hints, if we were starting from scratch I'd do it but it seems a bit low-pri to me now.

I agree that multiple py versions will make the testing slow, Py37 seems sufficient to me. Certainly I wouldn't expect a PR to break 3.8 if it doesn't break 3.7.

ryankarlos:
@bletham Thanks - so I’ve pushed the changes to run only tests with Travis. Unfortunately, it’s still taking over 20 minutes - I think installing devtools and other dependencies is taking the most time - not sure if there is any other workaround - would remotes be faster ?

PR #1509, opened 2020-05-21
Suppress warning from tqdm 

### LINKED ISSUE #1498: tqdm warning in notebook
#1338 added a progress bar to the cross validation using tqdm. Now, importing fbprophet in a jupyter notebook produces the warning
```
[...]/fbprophet/diagnostics.py:10: TqdmExperimentalWarning: Using `tqdm.autonotebook.tqdm` in notebook mode. Use `tqdm.tqdm` instead to force console mode (e.g. in jupyter console)
  from tqdm.autonotebook import tqdm
```
This will need to be cleaned up since it could be confusing to users who won't know what tqdm is and might think it's an error.

ryankarlos:
- [x] closes #1498 

@bletham  no idea what that warning means but i found this reported elsewhere  (https://github.com/jmcarpenter2/swifter/issues/66) and suppressed by importing `tqdm` from `tqdm.auto` instead - i checked and it works in `diagnostics.ipynb` for example (progress bars are ok as well).


bletham:
Thanks for working this out, and looking at the release notes in https://pypi.org/project/tqdm/ this is definitely the right fix.

PR #1512, opened 2020-05-25
Include mdape in performance metric docstring

ryankarlos:
addresses comment https://github.com/facebook/prophet/issues/911#issuecomment-633519884 

I haven't re-run the respective R cell because of an issue with my `rpy2` extension.


bletham:
Thanks!

Issue #1525, opened 2020-06-03
plot_components_plotly malfunctions when using m.add_country_holidays()


rteuwens:
**method:** `fbprophet.plot.plot_components_plotly()`
**issue:** not compatible with `m.add_country_holidays()`

`holidays` variable does not seem to be a Dataframe, causing an `AttributeError`.

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-50-fce739a52069> in <module>
----> 1 fig = fbprophet.plot.plot_components_plotly(m, forecast)
      2 fig.update_layout(layout)
      3 fig.show()

~\AppData\Local\Continuum\anaconda3\envs\forecast\lib\site-packages\fbprophet\plot.py in plot_components_plotly(m, fcst, uncertainty, plot_cap, figsize)
    713     if m.train_holiday_names is not None and 'holidays' in fcst:
    714         components['holidays'] = get_forecast_component_plotly_props(
--> 715             m, fcst, 'holidays', uncertainty)
    716 
    717     regressors = {'additive': False, 'multiplicative': False}

~\AppData\Local\Continuum\anaconda3\envs\forecast\lib\site-packages\fbprophet\plot.py in get_forecast_component_plotly_props(m, fcst, name, uncertainty, plot_cap)
    837         mode = 'markers'
    838         # Combine holidays into one hover text
--> 839         holiday_features, _, _ = m.make_holiday_features(fcst['ds'], m.holidays)
    840         holiday_features.columns = holiday_features.columns.str.replace('_delim_', '', regex=False)
    841         holiday_features.columns = holiday_features.columns.str.replace('+0', '', regex=False)

~\AppData\Local\Continuum\anaconda3\envs\forecast\lib\site-packages\fbprophet\forecaster.py in make_holiday_features(self, dates, holidays)
    532         row_index = pd.DatetimeIndex(dates.apply(lambda x: x.date()))
    533 
--> 534         for _ix, row in holidays.iterrows():
    535             dt = row.ds.date()
    536             try:

AttributeError: 'NoneType' object has no attribute 'iterrows'
```

bletham:
Thanks for raising this bug.

What's missing is that instead of using `m.holidays` it should first do
```
holidays = m.construct_holiday_dataframe(fcst['ds'])
holiday_features, _, _ = m.make_holiday_features(fcst['ds'], holidays)
```
(which is along the lines of 
[...]

PR #1526, opened 2020-06-04
Making sure 'holiday' is a Dataframe when using plot_components_ploty()

rteuwens:
Resolving issue #1525 **plot_components_plotly malfunctions when using m.add_country_holidays()**


bletham:
This looks great, thanks for the fix!

PR #1538, opened 2020-06-10
Faster builds for R CI

bletham:
See if MacOS build is faster for R unit tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,3 +30,4 @@ jobs:
         - R CMD INSTALL *tar.gz
       script:
         - R -e 'library(prophet); library(devtools); devtools::test()'
+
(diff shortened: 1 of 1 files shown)


ryankarlos:
Actually I was looking at this https://rstudio.github.io/renv/articles/ci.html 

Maybe we could use this to install the dependencies as a snapshot and then just load them in - during the Travis build .... the only downside is this would need to be updated every time new dependencies are added or modified.

bletham:
Actually it looks like when building the second time after cache:packages, the time was cut down to 7 minutes which seems much more reasonable and I think good enough for now. Since it's a standard travis feature I think it will be less likely to break in the future.

PR #1544, opened 2020-06-18
[work in progress] negative binomial likelihood

bletham:
I made a new branch to work on adding and evaluating a negative binomial likelihood. This is the first pass, and a notebook that tests it out.

[test_nbinom_likelihood.zip](https://github.com/facebook/prophet/files/4795620/test_nbinom_likelihood.zip)

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -13,6 +13,7 @@ from datetime import timedelta, datetime
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


bletham:
This comes with some numerical challenges in the hinge function. I think this PR provides a nice illustration of how to implement a new likelihood, but the particular use of NB likelihood for positive predictions will need some more consideration. I'm going to shelve this for now, see #1668 for discussion about strategies for getting positive forecasts.

Issue #1415, opened 2020-04-02
Serialize to/from json (Py)


bletham:
Right now the way to store a Prophet model in python is with pickle. This can occasionally produce issues (like #1361 and #1384), but even with workarounds for those, it's comes with other downsides like the fact that unpickling can execute arbitrary code.

The nicer approach would be to be able to serialize to and from json.

#1414 adds this, after which that would be the recommended approach for storing models, not pickle.

For convenience It is limited to fitted models; is there a scenario where one would want to store a model that hasn't yet been fit?

bkolb249:
When will this feature be released?

bletham:
In the next version release, which I'm aiming for the end of this month (so in 2-3 weeks).

bletham:
This is now shipped in v0.7, and documented in https://facebook.github.io/prophet/docs/additional_topics.html#saving-models

yfellous:
Hi Everyone ! I have this Error ? Help ?  

ModuleNotFoundError: No module named 'fbprophet.serialize'

bletham:
@yfellous check your version of fbprophet and make sure it is 0.7.1
```
import fbprophet
print(fbprophet.__version__)
```

Issue #1384, opened 2020-03-09
Large pickle sizes with v0.6


bletham:
Starting in v0.6 the StanBackend object is attached to the model as `m.stan_backend`. This causes the pickled size of the model to be huge: with the model in the quickstart example, it is 44M.

Clearing that attribute reduces the pickled size to 189K:
```
m.stan_backend = None

with open('model.pckl', 'wb') as fout:
    pickle.dump(m, fout)
```

This is related to but distinct from the pickle issues in #1361.

We really don't need `stan_backend` to be attached to the model after fitting (which is when it makes sense to pickle the model), so one option is to just wipe that attribute as the last step in `fit`. But this interacts with #482, since right now the stan fit object is being stored on the backend. So that will need to be refactored if we want to take this approach to solving the issue.

bletham:
It'd be great to have a unit test that verifies reasonable pickle size, to avoid regressions here in the future.

bletham:
Note that clearing this attribute as above is a workaround for this issue. `m.stan_backend` is only used during fitting, so you can still make predictions after loading the model from pickle if the attribute has been wiped.

CCadere:
Note that you can still make predictions if you wipe m.stan_backend or serialize the model (which also wipes the Stan objects), but will not be able to use the model for Cross Validation.

bletham:
Built-in json serialization is now pushed out in v0.7 and should be used instead of pickle as described here: https://facebook.github.io/prophet/docs/additional_topics.html#saving-models

PR #1546, opened 2020-06-19
Support clear redundant property in prediction.

eromoe:
Hi, 

I am using spark to schedule nearly half million timeseries trainning . I used to pickle model and save in hdfs , the file size is idea after remove `model.history` and `model.history_dates` .
But after update to 0.6,  the disk resource was quickly  eat up . 

I dig several hours to find this out, may not call problem , but it need concern .

- prophet 0.5  save 47898 model in 93M parquet file ( 1.98K per model ). 
- prophet 0.6  save 92 model in 77M parquet file   ( 857K per model ) . 

Since I have 574776 models.  The total size of all model  increased from 1118.4 M to  469G . That is really a disaster.

Then I tried to found which is the most big property by
```
for k in m.__dict__.keys():
    i = m.__dict__[k]
    print(k, len(dill.dumps(i)))
```

output : 
```
seasonality_params_list 714
growth 16
changepoints 1044
n_changepoints 5
specified_changepoints 4
changepoint_range 12
yearly_seasonality 14
weekly_seasonality 4
daily_seasonality 4
holidays 4
seasonality_mode 18
seasonality_prior_scale 12
changepoint_prior_scale 12
holidays_prior_scale 12
mcmc_samples 5
interval_width 12
uncertainty_samples 4
start 65
y_scale 198
logistic_floor 4
t_scale 62
changepoints_t 316
seasonalities 484
extra_regressors 832
country_holidays 4
stan_fit 4
params 993
history 8
history_dates 8
train_component_cols 8765
component_modes 550
train_holiday_names 4
fit_kwargs 21
stan_backend 1908072
```

It turns out to these which doesn't find in 0.5
```
fit_kwargs 21
stan_backend 1908072
```

As I saw, `stan_backend` is only used for fitting , would not use in prediction . 
I think we need a method to clean these redundant/useless property .

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1619,3 +1619,16 @@ class Prophet(object):
             weekly_start=weekly_start, yearly_start=yearly_start,
             figsize=figsize
         )
+
+
+    def clear_redundancies(self):
+        """Clear the redundant property in a fitted model which only used for prediction.
+
+        predict need `self.history` be not None
+        make_future_dataframe need `self.history_dates` be not None
+        `self.stan_backend` is only used in fit
+        """
+
+        self.history = -1
+        self.history_dates = -1
+        self.stan_backend = None


bletham:
Sorry for the extremely long delay on getting to this.

This did cause some problems for us too, including this exact size issue (#1384) and another issue in #1361. 

In the time since then, I actually added proper json serialization (#1415) which we will be maintaining. In our db, we switched from using pickled models to json serialized models. Once this ships in the next version (which I'm hoping will be in the next couple weeks), json serialization will be the recommended way to store models in python, and so this shouldn't be necessary anymore.

eromoe:
Oh, that is a good news !

PR #1556, opened 2020-06-26
edited package name

canute24:
Changed "unnittest to "unittest". I hope that's what it intended to be. Googled it too.
Thanks.


bletham:
Thanks for the fix! (and sorry for the slow merge!)

PR #1572, opened 2020-07-11
Add helper function to summarise coefficients of extra regressors

tcuongd:
I've seen a few Issues pop up about how to interpret the `beta` matrix / how to get the "coefficient" of an extra regressor, and I was thinking it would be worth having a helper function for this? Specifically, the function would transform the values in `beta` back to the original scale of the data values, and calculate lower / upper bounds if MCMC sampling was done.

Below is my working for the transformations:

* Additive regressor (ignoring seasonality)
```
y_scaled = trend +  beta * x_scaled
y / y_max = trend + beta * ((x - mu) / sigma)
y = trend * y_max + beta * y_max ((x - mu) / sigma)
```
So the incremental impact on `y` of a unit increase in `x` is
```
coef = beta * y_max / sigma
``` 
* For a multiplicative regressor the incremental impact would be `coef * trend(t)`.

Let me know if you guys think this is worth having.

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1565,8 +1565,7 @@ class Prophet(object):
         return pd.DataFrame({'ds': dates})
 
     def regressor_index(self, name):
-        """
-        Given the name of a regressor, return its index in the `beta` matrix.
+        """Given the name of a regressor, return its (column) index in the `beta` matrix.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


tcuongd:
CLA signed :)

bletham:
Thanks for the PR, and sorry for the long delay.

This does come up fairly frequently, and a utility for extracting this seems like it'd be a useful tool. Thanks for creating it!

A few things before merging:
- For the multiplicative case, we have
```
y/ymax = trend * (1 + ... + beta * (x - mu) / sigma)
y = ymax * trend * (1 + ... + (beta/sigma) * x + ...)
```
and basically the `ymax` term is absorbed into trend. So the coefficient we want for the multiplicative case would be just `(beta/sigma)`. This is because the `trend` in the equations above is the scaled trend - it's not the same trend that would be returned in the `forecast` dataframe; what's returned there is `trend * ymax`. That happens here:
https://github.com/facebook/prophet/blob/198ef890820d20ad280018c0d33d72091decba80/python/fbprophet/forecaster.py#L1325

So for the multiplicative mode it should just be (beta/sigma). Does that make sense?

- I've been trying to whenever possible take functionality out of the Prophet class, since it's gotten really large and has too many methods.  So I think it'd be nicer if this was a utility function that you passed a fitted model into, rather than being a Prophet method. You could make a new file "utilities.py" to put it into.

- Could you add a unit test for this? It could just fit a simple model, like here:
https://github.com/facebook/prophet/blob/198ef890820d20ad280018c0d33d72091decba80/python/fbprophet/tests/test_serialize.py#L31-L35
except it with regressors (I suppose one additive and one multiplicative would make sense). It doesn't need to verify values of beta (since there is some stochasticity in the model fitting), but just a check on the size of the output dataframe would be sufficient to make sure this isn't totally broken by some future change. If the function is put in a new file `utilities.py`, then the test would go in a new file `test_utilities.py`.

Thanks!

tcuongd:
Thanks for the feedback Ben, that's super helpful! 

- Yep I get what you mean about multiplicative regressors so I've made that change, thanks!
- I've moved the new functions to `utilities.py`
- I've also added `test_utilities.py` as well, do you think those tests are adequate?

tcuongd:
It looks like the CI is failing for `test_holidays.py` and `test_subdaily_holidays.py`, is this a known issue?

bletham:
This looks great, and yeah the CI failure is unrelated and needs to be fixed. Thanks for the contribution!

Issue #1416, opened 2020-04-02
How to handle covid-19 shock in the fbprophet


rajnish-garg:
Hi, I am using fbprophet on daily time series with 5 years of data. Due to this covid-19 pandemic most of our metrics are impacted. Because we are using last 20% of data for testing, so it is not able to capture the signals. Is there any recommendation on handling these issues.

rbagd:
I think the general answer is that you actually can't. Your forecasting model assumes a level of stationarity in your series and for heavily impacted series Covid-19 shock just breaks it. Maybe a model with a regime change might help but we haven't got any past pandemics to learn from. Time series that I am working on all have their seasonalities broken with European population in lockdown, there's not much you can do other than maybe switch to a short-term ARIMA.

Although I believe it might be a lost cause to try to model anything in the short-term, I am wondering what strategies people might use to mitigate the shock after the pandemic. Just set pandemic period to missing data and hope for the best? Can we do better?

bletham:
+1 for @rbagd's comment, Prophet assumes stationary seasonalities which is probably untrue for most human-related time series right now. In some areas there may be a long enough period of lockdown to check how stationary the seasonalities are within the lockdown period, and potentially fit a separate model just to that period. But there are so many frequent external shocks to the series (policy changes, etc.) that if things have stabilized, they probably won't stay stable for long.

As for after - I think we'll have to see what will end up working. Just throwing out the pandemic period as unuseful data could work but will really depend on what things look like after the pandemic clears up, and how similar they are to how things were before.

benmwhite:
>Just throwing out the pandemic period as unuseful data could work but will really depend on what things look like after the pandemic clears up

@bletham Or input the quarantine/lockdown periods as "holidays" in the model for full dark irony

mik1893:
I guess since the covid-19 pandemic has numerous impact which also vary on the nature of what you are trying to forecast and the business related to it it's pretty much impossible to define a standard set of rules to insert in a statistical modelling library...

lazaronixon:
great results here https://medium.com/@andrejusb/covid-19-growth-modeling-and-forecasting-with-prophet-2ff5ebd00c01 a used this model on my app profetadocorona.herokuapp.com

bletham:
[...]

PR #1592, opened 2020-07-23
covid recovery forecasting

SGokul1:
Covid Recovery Forecasting using Prophet


bletham:
Thanks for sharing this! The `notebooks` directly is actually specifically for generating documentation (the documentation at https://facebook.github.io/prophet/docs/quick_start.html is all generated from notebooks) so it's not the right place for this, but there is discussion of forecasting and COVID-19 in #1416, so that'd be a great place.

Issue #281, opened 2017-08-24
RMSE in Bayesian Context


reubenmc:
When accessing model fit, it seems like a lot of people have been using existing packages in R or Python.

However, with a package, we are simply getting the traditional RMSE as:
![image](https://user-images.githubusercontent.com/13903912/29680291-f991cec6-88b8-11e7-9589-e020be6f01d2.png)

In the Bayesian context where we have posterior samples from each prediction of interest, we can express this equation as:
![image](https://user-images.githubusercontent.com/13903912/29680353-265d0cd6-88b9-11e7-97c2-6822a983def1.png)

I know there's now a `m.predictive_samples(future)` function built into v0.2 so this wouldn't be hard to add. What are people thoughts? It seems like we are losing a lot of the uncertainty we gain from fitting a Bayesian model by simply using the posterior mean for model fit evaluation.

bletham:
This is a good suggestion - the new v0.2 has diagnostics (https://facebookincubator.github.io/prophet/docs/diagnostics.html) but they use just the MAP fit and the uncertainty intervals. We could also add diagnostics that use full posterior predictive samples.

PR #1612, opened 2020-08-06
Added Bayesian Root mean square error to diagnostics.py

Fortune-Adekogbe:


Fortune-Adekogbe:
I approached solving this using the equation for Bayesian Root mean square below.
<image src = "https://user-images.githubusercontent.com/13903912/29680353-265d0cd6-88b9-11e7-97c2-6822a983def1.png" width=300>
I also tried to make it conform to the format that pre-existing error measures had (majorly the rolling window parameter which necessitates the use of the rolling_mean_by_h function). Any help will be welcomed.

bletham:
Thanks for the contribution!

This code isn't quite computing what we'd want though. What we'd want for #281 is the RMSE averaged over the posterior of y (rather than computed just with the point estimate y_hat). The code here averages over `df['yhat'].values`, which is actually `yhat(t)` for each `t`, and not the posterior of `y` for a fixed `t`.

With the current API, there'd be no way to compute Bayesian RMSE inside `performance_metrics`. This is because the `cross_validation` function that produces the `df` input to `performance_metrics` does not compute the full posterior; it returns only the point estimate `yhat`, and quantiles of the posterior `yhat_lower`, `yhat_upper`. Right here is where `cross_validate` is making the prediction that returns only point estimate and quantiles, not the full posterior:
https://github.com/facebook/prophet/blob/40b170b95a7cdfdfe7f70190991bfb31a099007b/python/fbprophet/diagnostics.py#L241

So computing Bayesian RMSE properly would require a separate function for doing the cross validation in a way that maintains all posterior predictive samples. That said, an _approximate_ Bayesian RMSE could be computed within `performance_metrics`, which I think is the approach I would favor. Basically we could use `yhat`, `yhat_lower`, and  `yhat_upper` to approximate the posterior by moment-matching a Normal distribution, and then integrate over that approximate posterior. I think that'd be the most straightforward way of adding Bayesian RMSE, and I'd expect it to perform well in most cases.

Issue #1557, opened 2020-06-26
Plot continuous weekly seasonality with sub-daily data


anqus:
I have been recreating the plot_component daily and weekly graphs using the forecast dataframe.

```
forecast = m.predict(future)
fig = m.plot_components(forecast)
```

Recreating the daily component is straight forward, as it is repeated for every day in the dataset. So can be plotted as:

```
forecast['daily][:n].plot() 
# Where n = number of readings in 24hr period
```

However, the weekly component is less intuitive. It appears to match that created by plot_components when you choose the component value at 00:00 for each day. So can be plotted as (provided your data starts on Sunday):

```
forecast['weekly'][:n*7:n].plot()
# Where n = number of readings in 24hr period
```

This feels as though the plot is missing out on (n-1)*7 points of data for the weekly component. Would the weekly component be better displayed as:

```
# Plotting the whole component series for 1 week
# Where n = number of readings in 24hr period
forecast['weekly'][:n*7].plot()
```

or

```
# Taking an average for each day of the week
# Where n = number of readings in 24hr period
a = forecast['weekly'][:n*7]
groups = [a[x:x+n] for x in range(0, len(a), n)]
means = pd.Series([sum(group)/len(group) for group in groups])
means.plot()
```

bletham:
You raise a good point. As you note, the weekly seasonality plot shows the seasonal effect at just one point for each day. That happens here:
https://github.com/facebook/prophet/blob/0d260699e4776c449703a957f08eca8013461a47/python/fbprophet/plot.py#L287-L291

The reason this is done is really historical reasons: When Prophet was first released, it assumed daily data, so it makes sense to consider the seasonality as it will be applied to each day. As you note for sub-daily data, the plot is potentially missing important aspects of the seasonality. What we'd really want in that case is along the lines of the first thing that you suggested, `forecast['weekly'][:n*7].plot()`.

It should be straightforward to change the behavior here to do better. What we'd ideally want is for `plot_weekly` to:
- Identify frequency of the time series dates, as is done here:
https://github.com/facebook/prophet/blob/0d260699e4776c449703a957f08eca8013461a47/python/fbprophet/forecaster.py#L942-L943
- If frequency is daily, we can generate the plot exactly as is (evaluate weekly seasonality with daily frequency). If not, then use a finer grid, as is done for instance here:
https://github.com/facebook/prophet/blob/0d260699e4776c449703a957f08eca8013461a47/python/fbprophet/plot.py#L381-L384

[...]

PR #1615, opened 2020-08-07
Plot continuous weekly seasonality with sub daily data

anqus:
#1557 

This pull request allows Prophet to plot a continuous profile for the weekly component graph of sub daily data.

When calling plot_components(), the frequency of the data is now checked. If data is found to be sub daily, the weekly component will be plotted using plot_seasonality(), else the component will be plotted using plot_weekly(). 

Before (yosemite_temps data):
image

After Commit 1 & 2 (yosemite_temps data):
image

This pull request also alters the x labels and ticks, when plot_seasonality() is used for weekly components. It now displays the full weekday name, instead of the datetime. It also shows 8 xticks, to show every day of the week, and to illustrate the continuous nature of the component 

After Commit 3 (7 ticks - yosemite_temps data):
image

After Commit 4 (yosemite_temps data):
image

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -150,6 +150,9 @@ def plot_components(
 
     multiplicative_axes = []
 
+    dt = m.history['ds'].diff()
+    min_dt = dt.iloc[dt.values.nonzero()[0]].min() 
+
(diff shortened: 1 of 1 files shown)


bletham:
This is perfect, thanks for the great improvement!

PR #1621, opened 2020-08-12
make diagnostic images render correctly

collinsuzebu:


bletham:
Thanks for the PR! These markdown files are used to generate the documentation on the website, with the diagnostics one being here: https://facebook.github.io/prophet/docs/diagnostics.html

So the link is the correct one for the website, which is what we'd want here.

collinsuzebu:
Oh, I wasn't  paying attention to the documentation on the website. I'll close this PR now. Thanks.

Issue #1598, opened 2020-07-27
Another error with pip install 


mkleinbort:
Hello, 

When trying to run `pip install fbprophet` (in a python3.8 docker container) it tells me the `convertdate` module is not installed.

When I install the `convertdate` module it tells me the `lunarcalendar` module is not installed, and if I install `convertdate` and `lunarcalendar` it then tells me `holidays` is not installed.

Thus I need to do

```
pip install convertdate 
pip install lunarcalendar
pip install holidays
```

and then `pip install fbprophet`

(still not working, figuring out pystan)

bletham:
All of the dependencies for fbprophet are build dependencies, meaning they have to be installed before fbprophet can be installed. Usually if you are installing with pip and don't have the dependencies installed, it will first try to build a wheel, that will fail because the build dependencies are not in place, but then it will install the dependencies and install from source which succeeds. See #401 for lengthy discussion of this. It sounds like for you it is just exiting after the failure to build the wheel? What version of pip are you using?

You can get all the dependencies with
```
pip install -r requirements.txt
```
for this requirements.txt: https://github.com/facebook/prophet/blob/master/python/requirements.txt

but like I said I expect pip to install the dependencies itself, so am surprised to hear it isn't.

sarusso:
@bletham it seems like  there is a problem when using requirements to build the wheel itself (see the following  Dockerfile for how to reproduce it).

@mkleinbort this Dockerfile works:


```
FROM ubuntu:18.04

# Set non-interactive
ENV DEBIAN_FRONTEND noninteractive

# Update apt cache
RUN apt-get update

# Curl
RUN apt-get install curl -y

# Install get-pip script
RUN curl -O https://bootstrap.pypa.io/get-pip.py

# Install Python3 and Pip3, force pip to 20.2 to improve reproducibility
RUN apt-get install python3 python3-distutils -y 
RUN python3 get-pip.py && pip install -I pip==20.2

# Dev & build packages
RUN apt-get install python3-dev build-essential -y

# The following does not work for installing/building prophet (ModuleNotFoundError: No module named 'pandas')
#RUN pip3 install convertdate==2.1.2 lunarcalendar==0.0.9 holidays==0.10.3 cython==0.29.21 pystan==2.19.1.1 pandas==0.23.4 fbprophet==0.6

# Deps for successfully building prophet at the next step
[...]

PR #1629, opened 2020-08-18
Clearer pip install snippet

earthgecko:
This simple change may prevent further additions to and discussions in #401 and #1598 and the like.


bletham:
Looks good, thanks!

PR #1656, opened 2020-09-09
Correct serialize.py to handle numpy float

KDra:
Correct serialize.py to handle numpy floating values generated by the json serialisation.
We recursively correct the dictionary to convert alt numpy.floating instances to python 'float'.

### DIFF
--- a/python/fbprophet/serialize.py
+++ b/python/fbprophet/serialize.py
@@ -88,6 +88,21 @@ def model_to_json(model):
     model_json['params'] = {k: v.tolist() for k, v in model.params.items()}
     # Attributes that are skipped: stan_fit, stan_backend
     model_json['__fbprophet_version'] = __version__
+    
+    def std_json(level):
+        for k, v in level.items():
+            if isinstance(v, dict):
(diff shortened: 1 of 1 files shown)


bletham:
Hi, thanks for the PR. I guess there was a particular case that necessitated this adjustment? Could you create a new issue that provides code to reproduce the error this is fixing? Thanks!

KDra:
Hi. I don't have the code anymore but it is a known incompatibility between
the numpy float32 type and the python json library

bletham:
fbprophet shouldn't be directly serializing any numpy objects. Numpy arrays are converted to lists with `tolist()` before dumping to json:
https://github.com/facebook/prophet/blob/77da5b8c06d05d478af510198c685d454342db1f/python/fbprophet/serialize.py#L80-L81
which converts np floats to python floats.

There are a number of pandas objects that are serialized using panda's built-in serialization. I'm wondering if the numpy float is showing up in one of those? In either case, I'd like to track down which object has the numpy floats and fix the issue there rather than the generic functionality here, since it shouldn't be happening.

Issue #1678, opened 2020-09-17
option to disable Newton fallback?


quant9:
Per https://github.com/facebook/prophet/blob/master/python/fbprophet/models.py#L246, currently Prophet automatically falls back to Newton on unsuccessful LBFGS. However, for large timeseries Newton takes a very long time to converge (up to 30 mins vs 1 minute for LBFGS for my model on a 10-year, daily timeseries).

I wonder if there is interest in an option to error on the spot rather than falling back. This would allow the user to respecify the model instead of waiting a long time for Newton to finish. It would be very helpful for speeding up batch forecasting where say 10% of the models currently fail to Newton.

Any thoughts here would be helpful, thanks.

bletham:
I can definitely see value to being able to disable the fallback. The current behavior of falling back even if the algorithm is specified is pretty intentional (#877 and related) so I don't think we'd want to change that as the default, but as an option I can see it being useful.

Right now there really isn't a reasonable way to avoid it other than to locally edit the code in `PyStanBackend.fit`.

As for what the best interface to add this option to would be. I don't think it should go as a new arg to `Prophet()`. That constructor is already has too many args, and this particular one would not be used by most people. It could go as an argument to `Prophet.fit`, but even there to me it seems a little too infrequently used to go in such a high-level API. `Prophet.fit` already has the behavior that any extra kwargs are passed along directly to Stan, so to include it in those kwargs we'd have to add some kwarg filtering inside `PyStanBackend` and I'm not crazy about that either. I think perhaps the best solution would be to add a method to `PyStanBackend` for setting options. Right now this would be the only option, but it would allow for others in the future if they came up. So then the usage could look like:
```Python
m = Prophet()
m.stan_backend.set_options(newton_fallback=False)
m.fit(df)
```
How does that seem to you? Any other thoughts on the best way to get this in?

wkzhu:
Yes, as optimizer is specific to `fit` it makes sense to me to add a `set_options` functionality to `m.stan_backend`.
I can start a PR if that works for you.

bletham:
That'd be great, thanks!

PR #1702, opened 2020-10-03
add option to disable Newton fallback via set_options method

quant9:
Per https://github.com/facebook/prophet/issues/1678


bletham:
This looks great, thanks!

Issue #1551, opened 2020-06-24
Throwing "ModuleNotFoundError: No module named 'numpy' " during fbprophet installation


shyam-sreenivasan:
Trying to install fbprophet 0.6 and it keeps throwing similar errors for numpy, pystan and convertdate randomly.
Looks like prophet is expecting these modules to be installed obviously but somehow the prophet install begins before numpy is installed.

```
 Running setup.py install for fbprophet ... error
    ERROR: Command errored out with exit status 1:
     command: /home/cfx/PYTHONTEST374/bin/python -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-vqt2t7we/fbprophet/setup.py'"'"'; __file__='"'"'/tmp/pip-install-vqt2t7we/fbprophet/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-3eqd04na/install-record.txt --single-version-externally-managed --home /tmp/pip-target-4jtx3jpl --compile --install-headers /home/cfx/PYTHONTEST374/include/site/python3.7/fbprophet
         cwd: /tmp/pip-install-vqt2t7we/fbprophet/
    Complete output (42 lines):
    running install
    running build
    running build_py
    creating build
    creating build/lib
    creating build/lib/fbprophet
    creating build/lib/fbprophet/stan_model
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-vqt2t7we/fbprophet/setup.py", line 148, in <module>
        """
      File "/home/cfx/PYTHONTEST374/lib/python3.7/site-packages/setuptools/__init__.py", line 145, in setup
        return distutils.core.setup(**attrs)
      File "/usr/local/lib/python3.7/distutils/core.py", line 148, in setup
        dist.run_commands()
      File "/usr/local/lib/python3.7/distutils/dist.py", line 966, in run_commands
        self.run_command(cmd)
      File "/usr/local/lib/python3.7/distutils/dist.py", line 985, in run_command
        cmd_obj.run()
      File "/home/cfx/PYTHONTEST374/lib/python3.7/site-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
[...]

PR #1705, opened 2020-10-09
[python] Add setup requirements removing the need for a two-step install

john-bodley:
There is at least one [post](https://discuss.streamlit.io/t/error-when-deploying-with-docker/2873/4), several GitHub issues (https://github.com/facebook/prophet/issues/1551, https://github.com/facebook/prophet/issues/401), and the [Python documentation](https://facebook.github.io/prophet/docs/installation.html#python) which recommends that  you install `pystan` prior to installing `fbprophet` as the build has an _implicit_ dependency on `pystan`, `numpy`, etc.

The issue with this two-step approach is it's problematic when using tools like [`pip-tools`](https://github.com/jazzband/pip-tools), [`pip-compile-multi`](https://github.com/peterdemin/pip-compile-multi), etc. which purposely install/build all the requirements via a single `pip` command, and though `pip` installs "dependencies before dependant", without _explicit_ mention there is no guarantee that certain packages are installed prior to the building of the `fbprophet` package resulting in errors of the form (truncated):

```
Building wheel for fbprophet (setup.py): finished with status 'error'
ModuleNotFoundError: No module named 'numpy'
```

A potentially viable solution (tested) is to _explicitly_ define the requirements (specified in `requirements.txt`) as `setup_requires` which invokes [setuptools.dist.Distribution.fetch_build_eggs](http://code.nabla.net/doc/setuptools/api/setuptools/dist/setuptools.dist.Distribution.html#setuptools.dist.Distribution.fetch_build_eggs) thus ensuring the requirements are installed prior to building the package and removing the need for the two-step approach.

### DIFF
--- a/docs/_docs/installation.md
+++ b/docs/_docs/installation.md
@@ -37,13 +37,10 @@ Prophet is on PyPI, so you can use pip to install it:
 
 ```
 # bash
-# Install pystan with pip before using pip to install fbprophet
-$ pip install pystan
-$
 $ pip install fbprophet
 ```
 
-The major dependency that Prophet has is `pystan`. PyStan has its own [installation instructions](http://pystan.readthedocs.io/en/latest/installation_beginner.html). Install pystan with pip before using pip to install fbprophet.
+The major dependency that Prophet has is `pystan`. PyStan has its own [installation instructions](http://pystan.readthedocs.io/en/latest/installation_beginner.html).
 
 After installation, you can [get started!](quick_start.html#python-api)
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -117,7 +117,7 @@ with open('README.md', 'r', encoding='utf-8') as f:
     long_description = f.read()
 
 with open('requirements.txt', 'r') as f:
-    install_requires = f.read().splitlines()
+    requirements = f.read().splitlines()
 
 setup(
     name='fbprophet',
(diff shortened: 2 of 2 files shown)


quoimec:
Bump. This issue plagues my Dockerfiles requiring multiple requirement.txt files to install.

philippeboyd:
@quoimec In the mean time, on our end, I've temporary resolved it and we still keep one `requirements.txt`

```bash
pip install $(grep -ivE '^fbprophet' requirements.txt)
pip install fbprophet
```

So until this PR is merged, I just ignore `fbprophet` on the first `pip install` and install it after.

bletham:
Thanks for the PR. I know this is a big source of frustration but I'm not sure yet what the right fix is. I've looked into this particular approach in the past and it didn't look like it would work. There is a lot of discussion around this in #401. We tried to handle this with a pyproject.toml file (the way recommended by pip) but that didn't work because it builds the wheel in a separate build environment that may have different versions of pystan installed. We ran into an issue where pystan would build correctly in the system environment where the C++ toolchain was working properly, but failed in the build environment (see a description of it here: https://github.com/facebook/prophet/issues/401#issuecomment-452484397). This made the install error out.

As best as I understand it, the approach in this PR does a similar thing as using pyproject.toml: it sets up a build environment, uses `easy_install` to install the packages in the build environment, builds a wheel, and then installs that wheel outside the build environment. If you have a different version of pystan in the build environment vs. the system, or C++ toochain issues like we ran into with the pyproject.toml approach, then the wheel may fail to install. Is that correct?

The way I see it, the current behavior of having to pip install all of the dependencies before installing everything is annoying, but at least it consistently works. A situation where you can't install at all because the wheel won't build in the build environment is much worse, and that's what happened to us with pyproject.toml. So I'm pretty hesitant to do anything that might lead to that, and thus break the package for people.

bletham:
Oh, and I'm really sorry about the terribly slow response on this.

Issue #1677, opened 2020-09-17
v0.7.1 changelog and how to disable tqdm


fbonaita:
Hi,

I'm considering upgrading prophet from v0.6 to v0.7.1 but before doing so it'd be helpful to know exactly what the changes are about.
I couldn't find the changelog for either v0.7 or v0.7.1 on https://github.com/facebook/prophet/releases unfortunately. 
Is it possibly reported somewhere else?

On a different note, I noticed the new code uses tqdm when running `cross_validation` with `parallel` set to **None** but I was wondering if there's an easy way to disable it? I'm using `cross_validation` within another for loop with tqdm there too and they seem to conflict (duplicating progress bars while the process is running) and I'd like to keep only the higher-level progress bar.
Would removing the line of code below locally and reloading the function from that fix it? 
https://github.com/facebook/prophet/blob/77da5b8c06d05d478af510198c685d454342db1f/python/fbprophet/diagnostics.py#L189
More generally, I think adding a boolean flag to the code could be an idea worth considering.

bletham:
We're still trying to get the R 0.7 update pushed out (endless build issues with windows) and that's why the release hasn't been tagged yet or any type of changelog. In short, there shouldn't be any changes that would impact any existing forecasting workflows. There were a number of bugfixes, including a fix for an issue where pandas 1.1.0 was not backwards compatible and broke the holidays functionality. There is multiprocessing for cross_validation, as well as the ability to directly specify cutoffs. And there is now built-in json serialization for storing and loading models.

For tqdm: I don't think there is an easy way to disable it. It looks like there is some discussion of this on the tqdm side and a recommendation there (https://github.com/tqdm/tqdm/issues/619) but it seems pretty ad-hoc. I agree that it would be good to be able to have a bool flag to `cross_validate` that disables the use of tqdm. Yes, removing the use of tqdm from that line should be all that is required.

fbonaita:
Thanks for clarifying those points. 
I also previously missed the model storing/loading capability so that's good to know for when I decide to refactor my code.

Ok, I'm going to be away for a couple of weeks now but I'll try and have a look when I'm back. 
I was also considering whether it could be useful to add the ability to specify directly the number of iterations/cutoffs. I personally found it often easier to think in terms of iterations whether than period or cutoff dates and I implemented a workaround locally but I was wondering whether for `diagnostics.py` it'd be better to use a new arg or avoid a proliferation of flags and rethink the code a bit more.

[...]

PR #1706, opened 2020-10-09
Tqdm flag

fbonaita:
As discussed in https://github.com/facebook/prophet/issues/1677 this pull requests adds the flag `disable_tqdm` to the cross-validation function. It allows to disable the progress bar that appears when `parallel` is set to _None_.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -186,8 +186,7 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
 
     else:
         predicts = [
-            single_cutoff_forecast(df, model, cutoff, horizon, predict_columns)
-for cutoff in (tqdm(cutoffs) if not disable_tqdm else cutoffs)
+            single_cutoff_forecast(df, model, cutoff, horizon, predict_columns) for cutoff in (tqdm(cutoffs) if not disable_tqdm else cutoffs)
         ]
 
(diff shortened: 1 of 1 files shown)


python/fbprophet/diagnostics.py
```diff
@@ -186,7 +186,7 @@ def map(self, func, *iterables):
     else:
         predicts = [
             single_cutoff_forecast(df, model, cutoff, horizon, predict_columns)
-            for cutoff in tqdm(cutoffs)
+            for cutoff in (tqdm(cutoffs) if disable_tqdm==False else cutoffs)
```
mjt91, line 190: I might hit a very sensitive subject here: You compare booleans, which is never a really good idea. It maybe a good idea to not use `==False`, as recommended by PEP8 standards (see link below). My suggestion is:

```suggestion
for cutoff in (tqdm(cutoffs) if not disable_tqdm else cutoffs)
```

See [PEP8 - Programming Recommendations](https://www.python.org/dev/peps/pep-0008/#programming-recommendations) and this [StackOverflow Thread from 2011](https://stackoverflow.com/questions/4591125/is-it-safe-to-replace-with-is-to-compare-boolean-values).

Best
MJT

fbonaita: Fair point, I'll commit your suggestion.
Thank you.

bletham:
This is great, thanks a lot! (and sorry for the delay in merging!)

PR #1710, opened 2020-10-14
Update performance_metrics() in diagnostics.py

Pawel-Kranzberg:
Include smape as a valid metric.
Add support for monthly horizons.


bletham:
This all looks great to me, thanks for the PR and sorry for the long delay in merging!

PR #1711, opened 2020-10-14
Add SMAPE to diagnostics.R

Pawel-Kranzberg:
Add symmetric mean absolute percentage error.
Include mdape and smape as default performance metrics.

### DIFF
--- a/R/R/diagnostics.R
+++ b/R/R/diagnostics.R
@@ -222,6 +222,7 @@ prophet_copy <- function(m, cutoff = NULL) {
 #' 'mae': mean absolute error,
 #' 'mape': mean percent error,
 #' 'mdape': median percent error,
+#' 'smape': symmetric mean absolute percentage error,
 #' 'coverage': coverage of the upper and lower intervals
 #'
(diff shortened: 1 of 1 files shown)


R/R/diagnostics.R
```diff
@@ -500,6 +501,24 @@ mdape <- function(df, w) {
...
+#'
+#' @return Array of symmetric mean absolute percent errors.
+#'
+#' @keywords internal
+smape <- function(df, w) {
+  sape <- abs(df$y - df$yhat) / ((abs(df$y) + abs(df$yhat)) / 2)
+  if (w < 0) {
+    return(data.frame(horizon = df$horizon, smape = sape))
+  }
+  return(rolling_median_by_h(x = sape, h = df$horizon, w = w, name = 'smape'))
```
bletham, line 518: The SMAPE should be rolling_mean_by_h and not rolling_median_by_h, right?

Pawel-Kranzberg: Definitely, sorry for that.

bletham:
Great, thanks for the addition!

Issue #1721, opened 2020-10-21
UserWarning in plot when using logistic growth


iamaziz:
Hi, I am getting a warning alert when using `model.plot_components(forecast)` with the results from logistic growth. 

```
UserWarning: FixedFormatter should only be used together with FixedLocator
  ax.set_yticklabels(yticklabels)
```
It references this line 
https://github.com/facebook/prophet/blob/e41ed25646f44f713c110c30c07c678e4a07728e/python/fbprophet/plot.py#L421-L422

Environment:
- MacOS
- Python 3.8
- fbprophet 0.6
- matplotlib 3.3.2

The alert is gone after I tried the workaround solution described here https://stackoverflow.com/a/63755285/2839786 i.e. added `ax.set_yticks(ax.get_yticks().tolist())` beween the two lines above.

bletham:
Thanks for raising this, and sorry for the slow reply. Looking at the SO post, it looks like this is from a change introduced in matplotlib 3.3.1. I upgraded my matplotlib to 3.3.2 and have been trying to reproduce the error but haven't been able to yet - is there any chance you can post the code that led you to the error? It looks like it would be during a components plot with multiplicative seasonality but I haven't run into it yet.

iamaziz:
Thanks for getting back. Indeed, it is with multiplicative seasonality and logistic growth. Here is the sample code I used when I ran into the warning:

```python

model = Prophet(seasonality_mode='multiplicative', growth='logistic')
y['cap'] = 2e7
results = model.fit(y)

fake_future = model.make_future_dataframe(periods=180)
fake_future['cap'] = 2e7
forecast = model.predict(fake_future)

fig1 = model.plot(forecast, figsize=(20, 7))

trends = model.plot_components(forecast)
```

I'm using matplotlib 3.3.2 and fbprophet 0.6
![image](https://user-images.githubusercontent.com/3298308/98421911-7231a880-2058-11eb-9f23-f52574b422b4.png)

And by adding the line below [here](https://github.com/facebook/prophet/blob/e41ed25646f44f713c110c30c07c678e4a07728e/python/fbprophet/plot.py#L421-L422):
```python 
yticklabels = ['{0:.4g}%'.format(y) for y in yticks] 
ax.set_yticks(ax.get_yticks().tolist())
ax.set_yticklabels(yticklabels) 
```

the warning is gone: 
[...]

PR #1732, opened 2020-11-09
Fix "UserWarning" in model.plot_components

iamaziz:
See the related issue https://github.com/facebook/prophet/issues/1721


bletham:

Issue #1737, opened 2020-11-14
[Feature Idea] ablility to specify which columns to use in .fit function


kevkevinpal:
I noticed right now we need to specify `ds` and `y` for the dataframe that we are using in the Prophet.fit() function.

I was wondering if people thought adding an attribute called `date_col` and another called `y_col` to be able to specify the name of the column we want to use and otherwise just default to `date_col='ds'` and `y_col='y'` would be a good idea.

example use 
```
prophet = Prophet(yearly_seasonality=True, daily_seasonality=False)
prophet.fit(df, date_col='date_col_name', y_col='y_col_name' )

```

I can start working on it if people think its a good addition

Thanks,
Kevin

bletham:
@seanjtaylor and I discussed this idea pretty early on. We ultimately decided not to pursue it because the convenience inside the package and in the documentation for having a fixed data representation is pretty high, whereas the difficulty of converting column names to `ds` and `y` outside the package is pretty low (a single line of pandas or dplyr). Is there a particular use case that is driving this, or a situation where it would be a lot more convenient to not have to rename columns?

kevkevinpal:
ahh I see that makes sense would it make sense to be able to pass in those attributes and then mutate the df passed in directly to .fit or would that be confusing as everything else would be referred to as `ds` and `y`?

I didn't have a specific situation that would be a lot more convenient just saw it and thought it seemed a bit strange but your explanation helps!

PR #1738, opened 2020-11-14
[python] Add attribute to fit method to allow custom column names

kevkevinpal:
I opened an issue for this as well to track https://github.com/facebook/prophet/issues/1737

I noticed that when calling the .fit function we must have the columns named as `ds` and `y`. In my use that meaned I had to mutate my Dataframe in order to use this library. So I allowed an optional attribute with the name of `date_col` and `y_col` to allow the user to input the name of the columns they want to use. 

This makes it so users don't have to mutate the column names of their Dataframe. If there are any changes to be made feel free to leave a comment. Also if this feature is not needed/wanted feel free to close this PR as well

Still need to add tests to this PR and here is a snippet showing how it works

Screen Shot 2020-11-14 at 2 49 17 PM

Notice how we can pick out the columns we want to use vs having to rename to `ds` and `y`

Thanks,
Kevin

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1097,9 +1097,18 @@ class Prophet(object):
         """
        
         if date_col:
-             df.rename(columns={date_col:'ds'}, inplace=True)
+            if (date_col not in df):
+                raise ValueError(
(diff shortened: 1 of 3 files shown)


estepm4 (approved):
Looking good

bletham:
I think we're going to hold off on this for now as discussed in #1737

PR #1758, opened 2020-12-18
Added optional colour to plot_cross_validation_metric

LukeScales1:
Hi there,
This addition allows the comparison of different model's performance on the one plot:

multiple_models

Handy for comparing the performance of different model configurations or, as in the example shown here, models trained on different datasets

### DIFF
--- a/python/fbprophet/plot.py
+++ b/python/fbprophet/plot.py
@@ -469,7 +469,7 @@ def add_changepoints_to_plot(
 
 
 def plot_cross_validation_metric(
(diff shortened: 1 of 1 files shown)


LukeScales1:
As an example here is what it looks like using the existing method to compare different model configurations:
same-same

bletham:
This looks really useful, thanks for the PR! I did change it from "colour" to "color" though, which I hope you don't mind, to be consistent with the other uses of "color" in the package.

Issue #1696, opened 2020-09-29
plot_forecast_component() fails to plot any `name` containing spaces


cmcrowley:
Most names of `generated_holidays` contain spaces, and so fail to plot when passed to `plot_forecast_component()`. The [tidyeval approach](https://cran.r-project.org/web/packages/ggplot2/news/news.html) is to forego using `aes_()` or `aes_string()` and instead use quasiquotation in `aes`.

This is the same issue as documented [here](https://stackoverflow.com/questions/51658629/ggplot-aes-string-doesnt-work-with-spaces). [Here](https://github.com/tidyverse/ggplot2/issues/2447#issuecomment-364955240), Tidyverse creator Hadley Wickham responds to this issue by suggesting changing use of aes_string.

Reprex:
```
library(prophet)
df <- data.frame('ds' = seq.Date(from=as.Date('2015-07-01'),
                                 to=as.Date('2020-07-01'),
                                 by=1),
                 'y' = rnorm(1828))
m <- prophet()
m <- add_country_holidays(m, country_name = 'US')
m <- fit.prophet(m, df)
future <- make_future_dataframe(m, periods = 50)
forecast <- predict(m, future)
plot_forecast_component(m=m, fcst=forecast, name=m$train.holiday.names[1])
```

```
Error in parse(text = elt) : <text>:1:5: unexpected symbol
1: New Year
        ^
```

R Session Info: 
```
> sessionInfo()
R version 3.6.0 (2019-04-26)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows >= 8 x64 (build 9200)

Matrix products: default

Random number generation:
 RNG:     Mersenne-Twister 
 Normal:  Inversion 
 Sample:  Rounding 
 
locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252    LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] prophet_0.6.1 rlang_0.4.6   Rcpp_1.0.5   

loaded via a namespace (and not attached):
 [1] pillar_1.4.4         compiler_3.6.0       prettyunits_1.1.1    tools_3.6.0          pkgbuild_1.0.8       gtable_0.3.0         jsonlite_1.7.0      
[...]

PR #1766, opened 2020-12-30
used regex to format the `name` variable for plotting holiday components

arid-mangoss:
fixes issue/bug   #1696 

Removes all single quotes (') and double quotes (") in the `name` variable
AND replaces all whitespace with an underscore (_). Now, `ggplot2::aes_string()` can handle a column
name like the holiday `New Year's Day`.

### DIFF
--- a/R/R/plot.R
+++ b/R/R/plot.R
@@ -177,11 +177,17 @@ prophet_plot_components <- function(
 plot_forecast_component <- function(
   m, fcst, name, uncertainty = TRUE, plot_cap = FALSE
 ) {
-  formatted.name <- gsub("'|\"", "", name)
-  formatted.name <- gsub("\\s+", "_", formatted.name)
+
+  wrapped.name <- paste0("`", name, "`")
+  
+  lower.name <- paste0(name, '_lower')
+  lower.name <- paste0("`", lower.name, "`")
+
+  upper.name <- paste0(name, '_upper')
+  upper.name <- paste0("`", upper.name, "`")
 
   gg.comp <- ggplot2::ggplot(
-      fcst, ggplot2::aes_string(x = 'ds', y = formatted.name, group = 1)) +
+      fcst, ggplot2::aes_string(x = 'ds', y = wrapped.name, group = 1)) +
     ggplot2::geom_line(color = "#0072B2", na.rm = TRUE)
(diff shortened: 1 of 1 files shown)


bletham:
Thanks for tackling this! Can you show an example of this working? I don't think it will right now because I believe ggplot will be looking for `formatted.name` in the `fcst` dataframe, which it will not find because in the dataframe it is still the unformatted name.

arid-mangoss:
I see what you mean. I had a really hard time building it locally on my machine to test it so I didn't get a chance to do that previously. Sorry about that. 

I went ahead and corrected it using backticks. Here is an example of it working (note that I trimmed out some of the `if` statements in the code to make it run easily):
```R
x = seq(0, 10, 0.02)
y = sin(x)

name <- 'something with symbols # @  :)'
lowercolname <- paste0(name, '_lower')
uppercolname <- paste0(name, '_upper')

fcst <- data.frame(ds = x, 
                   'something with symbols # @  :)' = y, 
                   'something with symbols # @  :)_lower' = y - 0.2, 
                   'something with symbols # @  :)_upper' = y + 0.2, 
                   check.names = FALSE)

wrapped.name <- paste0("`", name, "`")
lower.name <- paste0(name, '_lower')
lower.name <- paste0("`", lower.name, "`")

upper.name <- paste0(name, '_upper')
upper.name <- paste0("`", upper.name, "`")

gg.comp <- ggplot2::ggplot(
  fcst, ggplot2::aes_string(x = 'ds', y = wrapped.name, group = 1)) +
  ggplot2::geom_line(color = "#0072B2", na.rm = TRUE)

gg.comp <- gg.comp +
  ggplot2::geom_ribbon(
    ggplot2::aes_string(
      ymin = lower.name, ymax = upper.name
    ),
    alpha = 0.2,
    fill = "#0072B2",
    na.rm = TRUE)

gg.comp
```
image

---
Note, I think there is one exception to what can be inputted. If the string `name` contains a backtick then it will break. I didn't know if it was worth addressing this but I think it would be possible -- though this covers 99.9% of cases, I think.

Cheers

bletham:
This looks great to me, and I tried it on the test case in #1696 and it works for that case too. Thanks for the fix!

PR #1769, opened 2021-01-05
Prophet wu

UlricWu:
fix: test_diagnostics.TestDiagnostics.test_performance_metrics

>  it's a unit test bug to assert equal to the df_none.columns

feat: add iter & tol_param in fit

> exploring the parameters  iter & tol_param in Prophet().fit() makes it easier for users to train models

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -1070,7 +1070,7 @@ class Prophet(object):
         m = df['y_scaled'].mean()
         return k, m
 
-    def fit(self, df, iter=1e4, tol_param=1e-8, **kwargs):
+    def fit(self, df, iter=1e4, **kwargs):
         """Fit the Prophet model.
 
         This sets self.params to contain the fitted model parameters. It is a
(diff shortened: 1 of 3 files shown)


bletham:
Thanks for the PR!. There's pretty clear value to being able to override the default number of iterations. I think there are a few tweaks that would be beneficial here though.

I think it'd be best to not include `iter` as a kwarg to `Prophet.fit`, since it won't always be meaningful. For instance it won't be meaningful if you are doing MCMC sampling, and could be confusing in that case. I think it'd be better to pass it in through the existing `**kwargs` input which is used to pass along configuration to Stan.

This should already work for the Pystan backend. You can see right here that it sets 1e4 as a default, but then overrides that with anything that was specified in the kwargs to `Prophet.fit` (the overriding happens in `args.update`):
https://github.com/facebook/prophet/blob/ff68962b2a3951e917f0a17c7958fc0f9e1bafb0/python/fbprophet/models.py#L248-L254

In the CmdStanPy backend it is hard-coded a 1e4:
https://github.com/facebook/prophet/blob/ff68962b2a3951e917f0a17c7958fc0f9e1bafb0/python/fbprophet/models.py#L85-L92
So that'd be good to fix, but it would be good to have it still default to 1e4. You would do this like
```Python
if 'iter' not in kwargs:
    kwargs['iter'] = 1e4
```
so we set it to 1e4 only if it hasn't already been specified in the kwargs to `Prophet.fit`. And then this line could be removed:
https://github.com/facebook/prophet/blob/ff68962b2a3951e917f0a17c7958fc0f9e1bafb0/python/fbprophet/models.py#L91

Does that sound good?

UlricWu:
Cool.
I can't agree with you anymore.

bletham:
The proposed PR in #1814 should work for this purpose as well.

PR #1778, opened 2021-01-10
Added growth='flat' functionality in R

arid-mangoss:
Hello, 

Firstly, I would like to thank the FB core DS team for making this open source. It is an extremely useful tool! Especially to the author @bletham for his commitment to this project.
___

I didn't want to keep having to use the python version of `prophet` whenever I wanted to run a model such as (in `Python`)
```python
m = Prophet(growth='flat')
```
So, I went ahead and implemented that same functionality in the `R` package which creates the feature requested in #614. 
___
**Motivation**
Sometimes, the data you are trying to forecast has a seasonality whose period is longer than the interval you have collected data. If you don't have a full cycle's worth of data then I've noticed that the `prophet` forecast will trend up or down where in a perfect world it should stay flat (since there isn't enough data to pick up on the seasonality). So, setting `growth='flat'` prevents it from picking up on the tail end of the data.
___
**Example**
Here is an example in `R` that "proves" that it works [(data from here)](https://facebook.github.io/prophet/docs/non-daily_data.html#monthly-data) -- I intentionally picked an increasing data set to show that the model does indeed stay 'flat', picking the most reasonable intercept (the mean of the data):
```R
> df <- read.csv('../examples/example_retail_sales.csv')
> m <- prophet(df, growth = 'flat', seasonality.mode = 'multiplicative')
Disabling weekly seasonality. Run prophet with weekly.seasonality=TRUE to override this.
Disabling daily seasonality. Run prophet with daily.seasonality=TRUE to override this.
> future <- make_future_dataframe(m, periods = 3652)
> fcst <- predict(m, future)
> plot(m, fcst)
```
image

Of course, the non flat model fits the data much better:
```R
> df <- read.csv('../examples/example_retail_sales.csv')
> m <- prophet(df, growth = 'flat', seasonality.mode = 'multiplicative')
Disabling weekly seasonality. Run prophet with weekly.seasonality=TRUE to override this.
Disabling daily seasonality. Run prophet with daily.seasonality=TRUE to override this.
> future <- make_future_dataframe(m, periods = 120, freq = 'month')
> fcst <- predict(m, future)
> plot(m, fcst)
```
image

---
**Other**
I also added unit tests for the functions and updated the documentation as well as the Jupyter notebooks for the demos.

Note -- I accidentally applied a VScode auto-formatter to one of my commits, which changed many small pieces of one of the files. I went back and reversed everything I could so that my changes weren't obfuscated by many smaller changes.


R/R/prophet.R
```diff
@@ -1177,7 +1198,7 @@ fit.prophet <- function(m, df, ...) {
     X = as.matrix(seasonal.features),
     sigmas = array(prior.scales),
     tau = m$changepoint.prior.scale,
-    trend_indicator = as.numeric(m$growth == 'logistic'),
+    trend_indicator = switch(m$growth, 'linear'=0, 'logistic'=1, 'flat'=2),
```
bletham, line 1201: This is cool, I didn't know about this function.

bletham:
This all looks fantastic, thanks for the PR and the really significant contribution!

juan-hincapie27:
Hello,
I have been trying to model the growth = 'flat' on RStudio but I'm still getingback the error that it should be either 'linear' or 'logistic'. Can anyone show me how to update the package so it supports the 'flat' trend?
Thanks in advance!

arid-mangoss:
@juan-hincapie27 

Hi Juan, 

This issue you are having is because the change I made has not yet been published to the CRAN repository, which is likely where you are getting your packages from when you do a `install.packages('prophet')`.

Thats no problem though, we can work around that. All you will have to do is just install it from this GitHub repository. Do that using:

```R
devtools::install_github('facebook/prophet/R')
```

This should point to the master branch and tries to install the package from there. Note that it will take a while because it downloads the source code from GitHub and then proceeds to build it. The build process can take quite a while (10 minutes on my computer sometimes). Also, note that it often works better on Linux or Mac, I have personally had issues with building on a windows machine because of C++ compiler issues. 

Do report back with your findings

Cheers

bletham:
An update on this: We were just this week successful in getting the new release on CRAN so this should work with a regular CRAN install now.

juan-hincapie27:
@bletham and @arid-mangoss thank you very much for your responses. Its working well now!

PR #1779, opened 2021-01-10
Makes it possible to build on windows in R

arid-mangoss:
I've been trying to build on a Windows machine and was unable to. In my recent PRs I've been building on a Linux VM in order to actually test the code. For production, I will need Windows, so I went ahead and fixed this.

I didn't see an issue for this, however, there is an issue over in `rstantools` which mentions this issue (found here: https://github.com/stan-dev/rstantools/issues/81)

The errors that I was getting when building on Windows was:

``` 
==> Rcmd.exe INSTALL --preclean --no-multiarch --with-keep.source R

* installing to library 'C:/Users/sssna/Desktop/prophet/R/renv/library/R-4.0/x86_64-w64-mingw32'
* installing *source* package 'prophet' ...
** using staged installation
** libs
  .
  .
  .
[snip]
  .
  .
  .
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for 'prophet' in inDL(x, as.logical(local), as.logical(now), ...):
 unable to load shared object 'C:/Users/sssna/Desktop/prophet/R/renv/library/R-4.0/x86_64-w64-mingw32/00LOCK-R/00new/prophet/libs/x64/prophet.dll':
  LoadLibrary failure:  The specified module could not be found.

Error: loading failed
Execution halted
ERROR: loading failed
* removing 'C:/Users/sssna/Desktop/prophet/R/renv/library/R-4.0/x86_64-w64-mingw32/prophet'
* restoring previous 'C:/Users/sssna/Desktop/prophet/R/renv/library/R-4.0/x86_64-w64-mingw32/prophet'

Exited with status 1.
```

For Windows, it seems as if the package `RcppParallel` was missing. I added this and that fixed the above errors. 

Then a warning came about when trying to build and when loading the package:
```
replacing previous import ‘Rcpp::LdFlags’ by ‘RcppParallel::LdFlags’ when loading ‘prophet’
```

This seemed to be caused by `Rcpp` and `RcppParallel` exporting the same function `LdFlags()`. So in the roxygen2 docs in `prophet.R` I did an `exclude=LdFlags` when importing `RcppParallel`

See here 
https://github.com/facebook/prophet/blob/db1e12217314cf70966e1f95c4803ab1af22a216/R/R/prophet.R#L79-L85

I ran the tests before and after this fix on Windows and Linux. All the test cases stayed the same. 

Cheers


bletham:
This is fantastic, as you saw in the rstan issue we had issues with this that were preventing getting the package on CRAN and were very frustrating for us to try to debug so we'll have to try again now, thanks for the fix!

Issue #1789, opened 2021-01-21
Changepoints with flat growth


Thuener:
Even if you put values for changepoints flat growth will ignore it. It is possible to get a more stepwise trend?
```python
import pandas as pd
import os
from fbprophet import Prophet
from fbprophet.plot import add_changepoints_to_plot

DATA_all = pd.read_csv('data.csv', parse_dates=['ds'])
m = Prophet(growth='flat', changepoints= ['2012-11-01', '2013-07-01'])
m.fit(DATA_all)

future = m.make_future_dataframe(10)
forecast = m.predict(future)

fig = m.plot(forecast)
add_changepoints_to_plot(fig.gca(), m, forecast);
```

![image](https://user-images.githubusercontent.com/770605/105368790-cb41bf00-5bf9-11eb-8c8a-a7df431d96e3.png)

Thuener:
Manage to construct a class to do what I want. If it is interesting I can create a PR to add stepwise growth.

```python
class ProphetStepWise(Prophet):
    
    def fit(self, df, **kwargs):
        m = super().fit(df, **kwargs)
        if self.growth == 'flat' and self.changepoints is not None:
            kinit = self.stepwise_growth_init(self.history, self.changepoints)
            self.params['m_'] = kinit[1]
        return self
        

    @staticmethod
    def stepwise_growth_init(df, changepoints):
        k = 0
        if len(changepoints) == 0:
            k = 0
            m = df['y_scaled'].mean()
        else:
            m = []
            for cp in changepoints:
                df_  = df[df.ds <= cp]['y_scaled']
                m.append(df_.mean())
            cp = changepoints.iloc[-1]
            df_  = df[df.ds > cp]['y_scaled']
            m.append(df_.mean())
        return k, m
    
    @staticmethod
    def stepwise_trend(t, m, changepoint_ts):
        if changepoint_ts[0] == 0:
            m_t = m * np.ones_like(t)
        else:
            m_t = np.ones_like(t)
            for s, t_s in enumerate(changepoint_ts):
                indx = (t <= t_s) & (m_t == 1.)
                m_t[indx] = m[s]

            t_s = changepoint_ts[-1]
            indx = (t > t_s) 
            m_t[indx] = m[-1]
        return m_t
    
    def sample_predictive_trend(self, df, iteration):
        """Simulate the trend using the extrapolated generative model.
        Parameters
        ----------
        df: Prediction dataframe.
        iteration: Int sampling iteration to use parameters from.
[...]

PR #1794, opened 2021-01-25
add implementation for stepwise trend in Python

Thuener:
Trying to fix issue #1789

I still need to do: 

- [x] test unix prophet.stan file 
- [x] find a volunteer to do the R code part
- [ ] unit tests

### DIFF
--- a/python/fbprophet/forecaster.py
+++ b/python/fbprophet/forecaster.py
@@ -155,9 +155,9 @@ class Prophet(object):
 
     def validate_inputs(self):
         """Validates the inputs to Prophet."""
-        if self.growth not in ('linear', 'logistic', 'flat'):
+        if self.growth not in ('linear', 'logistic', 'flat', 'stepwise'):
             raise ValueError(
-                'Parameter "growth" should be "linear", "logistic" or "flat".')
+                'Parameter "growth" should be "linear", "logistic", "flat" or "stepwise".')
         if ((self.changepoint_range < 0) or (self.changepoint_range > 1)):
             raise ValueError('Parameter "changepoint_range" must be in [0, 1]')
         if self.holidays is not None:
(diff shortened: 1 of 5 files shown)


arid-mangoss:
Hi @Thuener, 
I would be glad to take a stab at the R version of this. Before I start, I will wait to see what @bletham thinks about this feature. If he gives his blessing then I will jump on it :)

bletham:
Thanks for the PR, and sorry for the slow follow-up on this.

The commit is really clean and looks great. Like I said on the issue I'm really hesitant to expand the built-in trends into too large of a collection because it starts to drift the package away from the type of problems it is built around and can be confusing to the end user to understand what particular trend they might want. And there is just a really large set of trends that could be of interest. For this stepwise trend, what I'd like to do is not merge it right now but do put this PR in the documentation as the prototypical example for what needs to be done to customize the trend. If we start to see a lot of people in the future posting time series that look like they would benefit from a stepwise trend then we can revisit whether or not this should be built-in, but my sense now is that it isn't an especially common trend shape and that this PR will be much more valuable as an example for others who want to implement something from the long tail of functions.

So thanks again for the contribution, I'm going to close this and will change the documentation to point to this PR as the example for adding a new trend.

blunghino:
found my way here from the docs. stepwise trend would have been a useful feature for me!

rbagd:
I also just came around this PR as we were thinking of submitting a similar PR with our own implementation of a step-wise trend.

@bletham, would you reconsider adding (piecewise constant | flat discontinuous | step-wise) trend into the main package? I think it's a type which is common enough to reappear across datasets and industries and now I find that we wasted hours into reinventing the wheel. 😉

As counter-arguments to your reasoning, I would say

- Current implementation of `growth="flat"` is the only non-piecewise type of trend there is and I'd even argue the piecewise version should be the default choice for `growth="flat"` for more coherence across all trend types and their link with changepoints. You can always simulate a global flat trend by disabling changepoints, so it's a little bit of a weird situation.
- Maintaining custom Prophet forks is really challenging over time.

salva:
Hi, I am using Prophet for predicting sales of garments in a retail company and the trend for that is mostly stepwise as the mayor factor is the popularity of the collection and that factor usually remains constant along every season.

So, add this feature, please!!!

python/fbprophet/forecaster.py
```diff
@@ -1295,6 +1344,44 @@ def flat_trend(t, m):
         """
         m_t = m * np.ones_like(t)
         return m_t
+    
+    @staticmethod
+    def stepwise(t, deltas, k, m, changepoint_ts):
```
mycaule, line 1349: `deltas` and `k` actually unused here

python/fbprophet/forecaster.py
```diff
@@ -1142,9 +1179,13 @@ def fit(self, df, **kwargs):
         elif self.growth == 'flat':
             dat['cap'] = np.zeros(self.history.shape[0])
             kinit = self.flat_growth_init(history)
-        else:
+        elif self.growth == 'logistic':
```
mycaule, line 1182: should be `'logistic'` by default.

andreastn87:
I am currently working with prophet and a stepwise built-in trend would be really helpfull for me too!

PR #1798, opened 2021-01-27
Clean docs

arid-mangoss:
I updated the docs with some formatting issues, and I build it on my computer which generated new `Makevars` files


bletham:
Thanks for fixing the documentation issues, and sorry for the slow merge on this. I was a bit concerned by the change to the Makevars since they were autogenerated by rstantools, so I wanted to test the build in a few environments. I patched in the new Makevars and haven't yet run into any issues so it looks good to go!

Issue #1552, opened 2020-06-24
CmdStanR backend? 


jgabry:
It's great that prophet allows CmdStanPy as a backend, so I wanted give you an update and let you know that on the R side CmdStanR is now finally also available (https://github.com/stan-dev/cmdstanr). We haven't done a CRAN release yet but the interface is stable enough now that people are using it and other packages (e.g. brms) are already starting to offer CmdStanR as an optional backend. 

If you're interested in offering a CmdStanR backend definitely let us (me or @rok-cesnovar) know if you have any questions or run into any trouble.

bletham:
This is great to hear! We have had problems with rstan in the past where passing large matrices back from Stan to R has noticeable overhead (it can be several seconds), so I'll be interested to see if CmdStanR is an improvement on that. I think we could closely parallel the refactoring done to add the CmdStanPy backend to add support. For reference, #1045 was the issue for the CmdStanPy interface.

tcuongd:
Hey I'm keen to give this a go :) I'm not too sure what the best way to go about this is though. It seems like the approach we currently take is to build the C++ code for the stan model on install, then save it as a C++ module in the `stanmodels` object that can then be called in source code (ref: https://github.com/facebook/prophet/blob/master/R/R/stanmodels.R).

I'm not sure if `cmdstanr` can work directly from C++ code (just looking at this: https://mc-stan.org/cmdstanr/reference/cmdstan_model.html), so would we need another way to compile and save the stan model on install? Usually `cmdstan_model()` would recognize if the executable is up to date, so we could compile on install then call `cmdstan_model("inst/prophet.stan")` to instantiate the model in the code, but I'm not sure if that's good practice. Unless we can save the model as an RDS object instead?

Would appreciate any guidance on this!

tcuongd:
I've made a start on this here: https://github.com/facebook/prophet/pull/1806
I've just added functionality in the code to switch between the two backends for `fit.prophet` so far, but still not sure of the best way to handle installation and initial compilation.

How it's currently done in that draft PR assumes that `rstan` will be the default backend and will be installed regardless, but the user has the option to choose `cmdstanr` to fit if they have it installed. Once `cmdstanr` is available on CRAN perhaps we'll be able to remove the `rstan` and `Rcpp` dependencies and use `cmdstanr` as the default?

rok-cesnovar:
Nice! 

[...]

PR #1806, opened 2021-02-01
Add support for cmdstanr backend

tcuongd:
This is a start on: https://github.com/facebook/prophet/issues/1552

#### TODO

- [x] Changed the code in `prophet.R` and added `stan_backends.R` to allow switching between "rstan" and "cmdstanr" for `fit.prophet`. Checked locally and working as expected - see predictions on the test dataset below.
- Testing - similar to our CmdStanPy support, we'll run the test suite twice, once on the cmdstanr backend. Will be in a separate PR once we get R CI working.
- Generate docs, etc. - will be a separate PR.

image

#### Outstanding questions

- [x] Do we need to allow the user to specify the backend on install like in the Prophet implementation? I'm not too sure of the best way to do this. 
**Will leave as future improvement.**
- [x] Agree on the best way to initialize the cmdstan model object. The current `.load_model_cmdstanr()` function will just read from the `prophet.stan` file, and if it hasn't been compiled yet (i.e. first time use), it will compile and save an executable to the same folder. In later runs it should find the executable and compilation won't be needed. 
**Ideally this is done on install, but will leave as a future improvement.**


rok-cesnovar:
Thanks. Let me know if there is anything we can do to help or simplify stuff on our side. 

With cmdstanr you could also provide a model with within-chain parallelization if that is something you think your users would require. Do you have any example codes where fitting in Stan is the bottleneck? I can prepare an example.

bletham (commented):
Thanks for the PR, and sorry for the slow review. Overall this looks great to me and I would be happy to merge it in once it's finished. I have a few minor comments below, but it looks like these are the major outstanding questions:
* Building the model: Currently the rstan model is compiled at install time and the cmdstanr model is compiled on first use. That's fine for now. Ultimately I think it would be nicer to have it compile at install time (possibly when specified via an environment variable like it is for Py:
https://github.com/facebook/prophet/blob/fae1ac1f9a878879960722e8b3d72d45a0340b94/python/setup.py#L30-L32
) but it's fine for now.

* Specifying the backend to use: Right now this is done by passing the backend explicitly to `fit.prophet`. This is a little different from the Py implementation - in Py this is passed in on Prophet() init, and then if unspecified it tries them both to see which one works:
https://github.com/facebook/prophet/blob/fae1ac1f9a878879960722e8b3d72d45a0340b94/python/fbprophet/forecaster.py#L143-L154
It looks like in this PR, if unspecified it will default to rstan, and if that didn't work then it would just error, right?

I think I'd prefer to specify it in `prophet()` instead of in fit just to keep the interface as close to the Py one as possible (and I don't think it will really change much in the R because it can just be added as an attribute the same as it is in Py). We don't have to go as far as having it loop through different backends to find the functional one, but it would be nice to be able to set cmdstanr as the backend in some way that doesn't require explicitly doing it in every call to prophet(). The reason for that is because I think it would make testing a lot easier.

One way that I think this could be done would be via environment variables. If backend isn't specified, it could look for an environment variable R_STAN_BACKEND that could be RSTAN or CMDSTANR, and will use that if specified. If no environment variable, then it could just default to rstan. This would be along the lines of what is done in the Py version for specifying which backend to build at install time.

* Testing:  Basically all of the current unit tests should pass under both backends. In Py this is tested in the CI testing here, where basically the package is installed and tested twice, once under each baseline:
https://github.com/facebook/prophet/blob/fae1ac1f9a878879960722e8b3d72d45a0340b94/.github/workflows/build-and-test.yml#L26-L33
We don't have CI set up for R (it didn't look like github actions had support for R CI and we had to move off of Travis) so we don't have to worry about setting it up in CI. I think the ideal thing would be if here when we run the tests, it can check which backends are available, and run the tests under both if they're both available:
https://github.com/facebook/prophet/blob/fae1ac1f9a878879960722e8b3d72d45a0340b94/R/tests/testthat.R#L10
If we have the ability to specify the backend via an environment variable, then we could use `check_cmdstanr` to check if cmdstanr is set up, and if so, set the flag to cmdstanr and re-run all the tests now using that.

Let me know your thoughts on these! It all seems pretty doable to me.

tcuongd:
Thanks heaps @bletham! Really great feedback.

I've made the changes as you suggested (I think), it'd be great if you could take another look :)

I did have a few issues running the full test suite though:

- There are about 10 tests, mainly for cross-validation, that fail with the `cmdstanr` backend. In all cases the optimization routine fails to converge. Not really sure what could be the cause here - cmdstan should return the same results as rstan in all cases. @rok-cesnovar Do you have any ideas? The MCMC tests run fine.
- I'm not confident that all the logic in `testthat.R` runs. It seems that the checks terminate after running the first `test_check()` statement, and the intended re-run with the cmdstanr backend doesn't start. I'm not sure if this is what you had in mind. Perhaps you meant we would use the CI logic to re-run the tests instead?

bletham:
OK other than the test issues, from a code perspective this all looks great to me now. Sorry for a second slow review, as you may have seen we just pushed out a new version to CRAN and that's a whole process that usually takes a bit of back and forth to get things to pass on their end (this time it was a week from submission to acceptance) and I didn't want to merge any R changes while we were still working on getting the CRAN submission through.

For the `testthat.R`: I think we can remove that logic, I was envisioning just in the CI would would do something like
```
Rscript testthat.R
export R_STAN_BACKEND=CMDSTANR
Rscript testthat.R
```
to test them both (and could also do this locally). I don't think it necessary for a R check to necessarily run both.

Let me try to reproduce the failed test in cmdstanr. Some of the datasets used in unit tests have really bad fit in prophet, which doesn't really matter for the purpose of unit testing but I can see how the optimization could be numerically unstable. As for why it would fail in cmdstanr but not in rstan, I'm not totally certain but I wonder if the packages are shipping with different versions of cmdstan that might be linked to the optimization libraries in different ways or have differences in how they were compiled? In either case, I'm sure we can adjust the tests to make them more robust to numerical issues across different platforms.

I actually did notice in our CRAN push of v1.0 this last week that the tests pass for all of the platforms except for r-oldrel-macos-x86_64, where there are some failures, and I wonder if this could be related to what you're seeing (test flakiness depending on platform, https://www.r-project.org/nosvn/R.check/r-oldrel-macos-x86_64/prophet-00check.html ). Let me try to reproduce, but in the event I'm not (if it is platform dependent) if you could list the failing tests I could look at what we could do to make them more robust. (Assuming that is the source of the failure).

tcuongd:
Sorry for the super late reply @bletham, just got around to looking at this again. Here are the specific errors I ran into when testing:

`test_diagnostics.R`

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L25-L26

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L37-L38

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L51-L52

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L63-L64

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L83-L84

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L98-L101

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L111-L116

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L125-L126

https://github.com/facebook/prophet/blob/0616bfb5daa6888e9665bba1f95d9d67e91fed66/R/tests/testthat/test_diagnostics.R#L137-L138

All of the test failures come from running cross_validation(). Note that the initial model fit in those tests converged normally.

tcuongd:
Okay so I managed to fix the cross_validation errors:

* Most of the cross_validation tests require fitting the model on <100 points of data, so we were using the Newton algorithm for the MAP estimation. cmdstan didn't like this for some reason. If I just use the `lbfgs` algorithm for all fits (i.e. remove the check for the number of datapoints), the models fit fine and the cv checks pass.

However, doing this introduced errors in other tests 😅 :

```
── Error (test_prophet.R:33:3): fit_predict_no_changepoints ────────────────────
Error: Fitting failed. Unable to retrieve the draws.

── Error (test_prophet.R:52:3): fit_predict_changepoint_not_in_history ─────────
Error: Fitting failed. Unable to retrieve the draws.

── Error (test_prophet.R:475:3): auto_weekly_seasonality ───────────────────────
Error: Fitting failed. Unable to retrieve the draws.

── Error (test_prophet.R:546:3): auto_daily_seasonality ────────────────────────
Error: Fitting failed. Unable to retrieve the draws.
```

There were also a couple of warnings:
```
── Warning (test_prophet.R:475:3): auto_weekly_seasonality ─────────────────────
Fitting finished unexpectedly! Use the $output() method for more information.

── Warning (test_prophet.R:546:3): auto_daily_seasonality ──────────────────────
Fitting finished unexpectedly! Use the $output() method for more information.
```

Again it's kinda weird that cmdstan would return different results to rstan for MAP estimation, might be as you said that rstan is based on a slightly older version of cmdstan. Have you guys experienced this issue with MAP estimation before @rok-cesnovar  / any ideas on how we might resolve this? Which solver usually works as a good default?

tcuongd:
>  package doesn't have much modularity around preparing the arguments to be passed into the stan model 

This is exactly why I didn't try to dig into it :sweat_smile: thanks heaps for doing that! Yeah that does look pretty weird, although not sure how to raise a more minimal example...

Either way happy to merge as-is, everything else seems to be working. I'll raise a documentation PR next and make it clear that cmdstanr is experimental.

bletham:
Yeah constructing a minimal example of bad optimizer convergence might be a challenge. Alright let's merge this and at least in most normal situations it seems to be working great, and it's a perfect foundation to keep building on.

bletham:
Nice work, this was a major effort!

PR #1841, opened 2021-03-11
Optimize rolling_mean_by_h to make it faster with large window

lzs-raja:
The origin rolling mean algo takes O(nw) time where n is total length and w is window length. When doing with large n and large w, it will be very slow. I tried to optimize it to O(n+w) time by doing window first, then adding from left side and deleting from right side instead of creating a new window. It can decrease computation time from 30 mins to 0.3 secs when doing with millions of data.

### DIFF
--- a/python/fbprophet/diagnostics.py
+++ b/python/fbprophet/diagnostics.py
@@ -121,12 +121,12 @@ def cross_validation(model, horizon, period=None, initial=None, parallel=None, c
     predict_columns = ['ds', 'yhat']
     if model.uncertainty_samples:
         predict_columns.extend(['yhat_lower', 'yhat_upper'])
-        
+
     # Identify largest seasonality period
     period_max = 0.
     for s in model.seasonalities.values():
         period_max = max(period_max, s['period'])
-    seasonality_dt = pd.Timedelta(str(period_max) + ' days')    
+    seasonality_dt = pd.Timedelta(str(period_max) + ' days')
 
     if cutoffs is None:
         # Set period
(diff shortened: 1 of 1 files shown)


python/fbprophet/diagnostics.py
```diff
@@ -414,36 +415,30 @@ def rolling_mean_by_h(x, h, w, name):
...
-        xbar = float(xm[i])
-        j = i - 1
-        while ((n < w) and j >= 0):
-            # Include points from the previous horizon. All of them if still
-            # less than w, otherwise just enough to get to w.
-            n2 = min(w - n, ns[j])
-            xbar = xbar * (n / (n + n2)) + xm[j] * (n2 / (n + n2))
-            n += n2
+    while i >= 0 and j >= 0:
+        while (n_sum < w) and (j >= 0):
```
bletham, line 432: There is a bug here where if we reach `j<0` before we reach `n_sum >= w`, we do not want to include the point in the res_x below. This should be easy to fix by putting a check before the `append` that n_sum >= w.

bletham (commented):
This is fantastic, thanks for the contribution. Not only is it much more efficient, but the code is easier to understand!

There is one minor bug in a situation that wasn't covered by the unit tests but that you can see with this example:
```Python
x = [1., 2., 3., 4., 5., 6., 7.]
h = [1., 2., 2., 2., 2., 3., 4.]
df = rolling_mean_by_h(x, h, w=3, name='test')
```
The output of this is
```
   horizon      test
0      1.0  1.000000
1      2.0  3.500000
2      3.0  4.333333
3      4.0  5.500000
```
which is all correct except for the horizon=1.0 row: there is only 1 point with horizon<=1, and so with w=3 that point should be left out. The bug should be easy to fix as I note in the comment below. Other than that, this is good to merge! We should in the future make the same improvement to the R version of this function: https://github.com/facebook/prophet/blob/8882c6a3e370736c382db14133f9c66f3b509c30/R/R/diagnostics.R#L345

tcuongd:
Hi @lzs-raja ! Thanks again for this - just wondering how you're going with the comments above? This would be hugely beneficial so let us know if you need any additional help getting it through :)

PR #1844, opened 2021-03-15
Pakage rename

bletham:
Rename fbprophet to prophet.

Make fbprophet a shim package that just imports prophet. Any code currently importing from fbprophet will continue to work, but will get a warning message that they should update to use the prophet name.

Update all of the notebooks to use the new package name.

I'll wait until after it is pushed to pypi to update the markdown documentation and references in the website.

# Testing

## Check that all references are changed:
grep -R fbprophet *

## Check new prophet package works:
Uninstall prophet and fbprophet:
```bash
pip uninstall fbprophet
pip uninstall prophet
```
Make sure they're gone:
```python
import fbprophet  # ModuleNotFoundError
import prophet  # ModuleNotFoundError
```
Install prophet and check that it works:
```bash
cd python
python setup.py install
cd prophet/tests
python -m unittest
```

## Check the shim works:
```bash
cd ../../../python_shim
python setup.py install
cd fbprophet/tests
python -m unittest
```

## Final check by running all notebooks with updated package name
All successful


seanjtaylor (approved):
tested. Both shim and the new package work fine and pass all tests.

Issue #1834, opened 2021-03-04
Upgrade CmdStanPy interface


bletham:
The CmdStanPy version is currently pinned at 0.9.5, which came out Mar 11 2020.

Since then, there has been a breaking change in the interface for extracting parameter values as described here: https://github.com/stan-dev/cmdstanpy/pull/287

Based on the dates, I expect that versions 0.9.67 and later are not compatible with fbprophet.

It should be straightforward to adapt our code to the new interface in that PR. Everything that needs to be changed on our end should be in these two functions, where we use CmdStanPy to do the optimization/sampling and then extract the parameter values:
https://github.com/facebook/prophet/blob/bbc44bf67c58ebb4f0a5fabda7f6cfdd793330f7/python/fbprophet/models.py#L87
https://github.com/facebook/prophet/blob/bbc44bf67c58ebb4f0a5fabda7f6cfdd793330f7/python/fbprophet/models.py#L117

The extraction of parameter values is what has changed in CmdStanPy, and the logic for doing that in our code is in this function:
https://github.com/facebook/prophet/blob/bbc44bf67c58ebb4f0a5fabda7f6cfdd793330f7/python/fbprophet/models.py#L173

Adapting that to the new interface will allow us to keep up with the latest CmdStanPy releases.

PR #1849, opened 2021-03-20
Upgrade CmdStanPy interface

tcuongd:
Fixes https://github.com/facebook/prophet/issues/1834

I ran the test suite locally with cmdstanpy version 0.9.68:

```
fbprophet/tests/test_diagnostics.py ...........                                                                               [ 20%]
fbprophet/tests/test_prophet.py ............s...s.....................                                                        [ 92%]
fbprophet/tests/test_serialize.py ...                                                                                         [ 98%]
fbprophet/tests/test_utilities.py .                                                                                           [100%]
```

Should we also update the version of `cmdstanpy` in `requirements.txt`?


python/fbprophet/forecaster.py
```diff
@@ -1170,6 +1170,7 @@ def fit(self, df, **kwargs):
         else:
             self.params = self.stan_backend.fit(stan_init, dat, **kwargs)
 
+        self.stan_fit = self.stan_backend.stan_fit
```
tcuongd, line 1173: I added this because `forecaster.stan_fit` is currently `None`. I think this is a clearer way to expose the model object (rather than having to dig into `self.stan_backend`).

python/fbprophet/models.py
```diff
@@ -128,7 +128,7 @@ def sampling(self, stan_init, stan_data, samples, **kwargs) -> dict:
                                           inits=stan_init,
                                           iter_sampling=iter_half,
                                           **kwargs)
-        res = self.stan_fit.sample
+        res = self.stan_fit.draws()
```
tcuongd, line 131: The `.sample` property now has a deprecation warning, and the guidance is to use `.draws()` instead.

bletham:
This is great, thanks for taking care of this. I went ahead and upgraded the version in requirements and also in the CI just to verify that everything is working in CI too, and it's all green!

PR #1850, opened 2021-03-20
Handle np.int and np.float deprecations

tcuongd:
As per https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations

The deprecation message:

```
DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    .astype(np.float)
```

These changes are assuming we want to keep `int` and `float` flexible in these cases. Or perhaps we should just use `np.int64` and `np.float64` anyways to be explicit?


bletham:
Thanks for the fix! I don't know that there's any particular reason for why t is using single vs. double precision, but doesn't seem like it would matter much.

Issue #1838, opened 2021-03-08
Feature Request: Legends for all plots


timkofu:
Maybe I'm missing a setting, but there currently seems to be no inbuilt in way to enable legends for the plot functions. Would it be possible to add this feature? Many thanks.

bletham:
That is correct. The components plot only have 1 line so there isn't need for a legend there, but I can see why one might want to add a legend to the `m.plot()` plot. A flag "include_legend" to `m.plot` would be a nice addition.

In the meantime, you can do this directly in matplotlib like so:
```Python
fig = m.plot(fcst)  # This returns a matplotlib Figure
ax = fig.gca()  # Now you have the matplotlib Axis
ax.legend([....])  # Put whatever legend you want here.
```
Once you have `ax`, you can make arbitrary changes to the plot using matplotlib.

vshargin:
I can help adding this to the Python package.

PR #1858, opened 2021-03-27
Add include_legend flag to m.plot()

vshargin:
Added the `include_legend` to show legend in `m.plot()` as proposed in issue #1838


bletham:

seangrant82:
@vshargin I was using 
`from prophet.plot import plot_plotly, plot_components_plotly`

`plot_plotly(m, forecast)`

is there an example that uses this flag?

PR #1871, opened 2021-04-16
Add model_to_dict and model_from_dict functions

sd2k:
This commit adds the model_to_dict and model_from_dict functions, using
all of the logic that previously lived in model_to_json and
model_from_json, and converting those functions to simply reuse the new
ones.

This is useful because sometimes the user may want to serialize the dict
in some other way (e.g. another JSON serialization library such as ujson
or orjson, or something entirely different).


bletham:
Looks great to me, thanks for the addition!

matbmeijer:
Hi, when will this be released in pypi?

PR #1891, opened 2021-04-30
Add CI for R

tcuongd:
Reference: https://github.com/r-lib/actions/tree/master/examples#standard-ci-workflow


tcuongd:
@bletham R CI is working now for Ubuntu. Some open questions:

- Do we need to run this for other platforms too?
- The structure of the R job is slightly different to the Python job - I've mainly copied the file from the r-lib/actions examples. Let me know if you want to change anything / make anything clearer.

bletham:
Amazing, this will be so useful! Thanks!

PR #1913, opened 2021-05-13
add setup_requires on pystan so build requirements are in the metadata

CJ-Wright-Citadel:
It seems that pystan is a build time requirement (since it is used in the `from prophet.models import StanBackendEnum` import in `build_models`). It seems however, that it isn't listed in the `setup_requires` so that the installing system doesn't know that it is a build requirement.

This PR adds the needed pystan build requirement.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -129,7 +129,7 @@ setup(
     author_email='sjtz@pm.me',
     license='MIT',
     packages=find_packages(),
-    setup_requires=[
+    setup_requires=['pystan~=2.19.1.1'
     ],
     install_requires=install_requires,
     python_requires='>=3',


hildobby (approved):
I think this is a very needed PR given the current incompability with pystan 3!

Djailla:
Hello, I strongly approve this is a needed PR !

tcuongd:
Thanks for this! I'm just wondering why this wasn't done in the first place, not sure if there was a particular reason for it.

I'll dig a little further, but in the meantime @bletham do you remember why off the top of your head?

tcuongd:
Heya, the latest changes to setup.py as a part of this PR: https://github.com/facebook/prophet/pull/2010/files
covers the core problem here.

Instead of using setuptool's `setup_requires`, we've gone for the recommended approach of specifying the build system requirements in a `pyproject.toml`. Now the "proper" way to build prophet from source will be to do `python -m build .`.

Hopefully most end users won't need to do that though, since the PR also allows us to build wheels for different OS and python versions, so Prophet should mostly just work out of the box (pending changes needed for Windows and python 3.9).

I'll close this now, but if you have any other questions let me know!

PR #1961, opened 2021-07-08
Fixing link to pystan docs

fhightower:


fhightower:
Ok, I'll let you make the changes you see fit.

Issue #1994, opened 2021-08-12
Update README.md, fix broken link Linux installation


purwowd:
Fixes a broken link to Pystan documentation about Linux installation instructions

PR #1996, opened 2021-08-12
 Update README.md, fix broken link Linux installation #1994

purwowd:


README.md
```diff
@@ -67,7 +67,7 @@ pip install pystan==2.19.1.1
 pip install prophet
 ```
 
-The default dependency that Prophet has is `pystan`. PyStan has its own [installation instructions](http://pystan.readthedocs.io/en/latest/installation_beginner.html). Install pystan with pip before using pip to install prophet.
+The default dependency that Prophet has is `pystan`. PyStan has its own [installation instructions](https://pystan.readthedocs.io/en/latest/installation.html). Install pystan with pip before using pip to install prophet.
```
tcuongd, line 70: Thanks for raising this! But it seems like we need to change this line altogether, since `pystan >=3` is not compatible with Prophet.

It should read something more similar to 
"Install `pystan==2.19.1.1` with `pip` before installing Prophet with `pip`. Note that the version **must** be pinned, as the latest `pystan` version is incompatible with Prophet."

purwowd: Hi @tcuongd thanks for your explanation. I will close this PR & issue.

Issue #49, opened 2017-02-28
multivariate modeling 


jmarca:
I see from issue #18 that there are no concrete plans yet for prophet to handle multivariate data. I am interested in tackling this feature; any pointers on where to start?

My application area is vehicle traffic.  For example, numbers of cars and trucks in each lane moving past a point (yes I also read that sub-daily data is also not yet well supported).  While I can predict traffic lane by lane by car by truck, it would make more sense to model all lanes and vehicle types at once, since they are related (lots of trucks in the right lanes means fewer cars in those lanes, more cars in left lanes, etc etc.).

seanjtaylor:
Thanks for the interest in taking this on! Here are my current thoughts.  

## Simultaneous models

```
y_1(t) = f_1(t) + e_1(t)
y_2(t) = f_2(t) + e_2(t)
```

The two ways y_1(t) and y_2(t) can be related are:

1. If f_1 and f_2 are related somehow.  The way to do that would be to have all of their parameters have a prior distribution so they are shrunk toward the same values.  E.g. they share related seasonality, trend, and changepoints.
2. If e_1(t) and e_2(t) are correlated.  This is a bit harder to model and interpret.

We could do 1. pretty easily in Stan, but the benefits for sharing the parameters would be small unless you have many time series (I'd say at least 5-10).

## Hierarchy of time series

```
y_1(t) = f_1(t) + e_1(t)
y_2(t) = f_2(t) + g(y_1(t)) + e_2(t)
```

In this case there's a hierarchy of time series.  So it's a bit simpler (assuming you know the order):

1. you fit the `y_1` model by itself
2. you regress `y_1(t)` on `y_2(t)` and learn their relationship.
3. instead fitting `y_2` model, you forecast residual: `[y_2(t) - g(y_1(t))]`
4. To forecast `y_2`, you forecast `y_1`, use the model from (2) to and the forecast from (3) to construct the `y_2` forecast.

This strategy requires more bookkeeping and model fits, whereas the first strategy everything would be done in one fitting step.

Curious if @bletham has thoughts.

bletham:
[...]

PR #1998, opened 2021-08-14
Segment-level forecasts using a top-level regressor - example docs

tcuongd:
This PR adds a case study / walkthrough on the top-level-regressor approach to segment-level forecasts outlined by Ben Letham here: https://github.com/facebook/prophet/issues/49#issuecomment-430383681

We use cross-validation to compare this strategy against forecasting each segment individually.

[nbviewer link](https://nbviewer.jupyter.org/github/tcuongd/prophet/blob/tcuongd-multivariate-tlr-example/notebooks/multivariate_top_level_regressor.ipynb).


tcuongd:
R CI failure is unrelated, should be fixed once https://github.com/facebook/prophet/pull/1999 is merged and reflected in this branch.

PR #2010, opened 2021-08-28
[Draft] Python Wheels for PyPi

tcuongd:
Having a go at creating wheels for the Python package with the aim of publishing to PyPi. We're trying to create a wheel that has both `pystan` and `cmdstanpy` models compiled, so the end user just has to install the wheel and `prophet` should work out of the box.

### Tasks

- [x] Install `cmdstan` as a part of `setup.py`, and include the core `cmdstan` files in the prophet distribution as well. This allows end users to fit models on the `CMDSTANPY` backend without an existing installation of cmdstan, and ensures that we the model is fit using the same version of cmdstan that compiled it.
- [x] Use `pyproject.toml` to specify build dependencies, with the aim of using `pypa/build` as the build engine. This seems to be the recommended approach to [building wheels](https://packaging.python.org/tutorials/packaging-projects/).
- [x] Remove imports of `prophet.models` from the `setup.py` script -- the build system was complaining that packages could not be found (because the build happens in an isolated environment). I think this is actually a bit cleaner.
- [x] Create a `wheel.yml` Github Actions workflow. The aim is to run this when we release the package, and publish to PyPi automatically. Currently using [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/options/) to try and simplify this process.
- [x] Get all relevant wheels built successfully via `wheel.yml` (across different platforms and for Python versions 3.6, 3.7, 3.8).
    - [x] macOS. Test wheel here for Python 3.8: https://test.pypi.org/project/prophet/#description
    - [x] Linux -- having issues with running `auditwheel` to create the `manylinux` wheels.
- [x] Finalise the testing + automated publishing via `wheel.yml`
- [x] Spot check that building from source on the end user's machines is not broken (would be required for versions / platforms where we haven't published a wheel).

### Open questions

**Windows**

Getting the necessary build tools on Windows will be tricky, but I think `cmdstanpy` might make it easier with the `install_cxx_toolchain` utility function. This should be its own PR though.

**cmdstan**

* The PR as it currently stands removes the option for the user to use their own `cmdstan` installation. Should we still make this available for end-users who want to build their own wheel?


.github/workflows/wheel.yml
```diff
@@ -0,0 +1,61 @@
+name: "Create Python Wheels"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
tcuongd, line 7: Temporary, for testing.

python/setup.py
```diff
@@ -16,26 +17,77 @@
...
 if platform.platform().startswith('Win'):
     PLATFORM = 'win'
 
 MODEL_DIR = os.path.join('stan', PLATFORM)
 MODEL_TARGET_DIR = os.path.join('prophet', 'stan_model')
-
+CMDSTAN_VERSION = "2.26.1"
 
 def get_backends_from_env() -> List[str]:
-    from prophet.models import StanBackendEnum
```
tcuongd, line 31: This was causing the build system to fail, since the build happens in an isolated environment with only the dependencies needed for `setup.py`. I think removing this makes things a bit cleaner, but keen to hear thoughts.

python/setup.py
```diff
@@ -16,26 +17,77 @@
...
     for backend in get_backends_from_env():
-        StanBackendEnum.get_backend_class(backend).build_model(target_dir, MODEL_DIR)
+        print(f"Compiling {backend} model")
+        if backend == "CMDSTANPY":
+            build_cmdstan_model(target_dir)
+        elif backend == "PYSTAN":
+            build_pystan_model(target_dir)
+
+
+class BDistWheelNonPure(_bdist_wheel):
```
tcuongd, line 79: This ensures that the wheel gets tagged with the appropriate operating system / platform, not just "any".

python/setup.py
```diff
@@ -16,26 +17,77 @@
...
+    def finalize_options(self):
+        _bdist_wheel.finalize_options(self)
+        self.root_is_pure = False
+
+
+class InstallPlatlib(install):
+    """Ensure wheel has correct platlib specification when being repaired by auditwheel."""
```
tcuongd, line 86: Taken from here: https://github.com/google/or-tools/issues/616#issuecomment-371480314 to try and resolve the platlib error raised by `auditwheel`, but I'm not sure if has an effect.

python/setup.py
```diff
@@ -129,15 +181,15 @@ def with_project_on_sys_path(self, func):
     author_email='sjtz@pm.me',
     license='MIT',
     packages=find_packages(),
-    setup_requires=[
```
tcuongd, line 132: Ensures that we read from `pyproject.toml` for build dependencies instead.

freddyaboulton:
@tcuongd Thanks for starting this! I think this is really cool.

I tried some stuff out on my fork: https://github.com/freddyaboulton/prophet/pull/1 to no avail as well.

I think the problem with auditwheel is that it's finding compiled code in the wheel when the package does not specify any extensions in the setup.py so it thinks it's "pure".

We can obviously skip the audit step but then the problem is that cmdstan dynamically links against some libraries that would not be included in the wheel so it's actually not exportable.

One thing that came to mind is that if cmdstan statically linked against its dependencies, then we could skip the audit step and I think it should "work". So maybe that's something worth trying? 

I'm pretty new to this myself so I don't have any concrete answer (and it's possible what I'm saying here doesn't make sense at all!) but just want to let you know that this is cool and I'm willing to help if possible.

Another point to discuss is that I added testing to the cibuildwheel script and the macosx build passes but then the tests start failing in a really weird way?

https://github.com/freddyaboulton/prophet/runs/3498406868

tcuongd:
Thanks heaps for looking into this as well @freddyaboulton !! Yeah definitely a new area for me to so still trying to piece everything together 😅 

**Linux wheel**

Yeah the Linux one is weird. I tried to bundle the necessary cmdstan files into the wheel, and didn't run the repair command, but the wheel was rejected by PyPi when I tried to upload it. I've received some good feedback about this here: https://github.com/pypa/packaging-problems/issues/542 (seems like we'll need to change our `setup.py` file).

**MacOS tests**

Oh I didn't realise `cibuildwheel` had wheel tests baked in as well, nice find! I think what's happening on your branch is that it's trying to run the tests against the source code (`{project}/tests`), but that's not where we built the binaries. 🤞 if we change that to `{package}/tests` it'll work.

freddyaboulton:
@tcuongd great call adding the custom build_ext command to setup.py! I was able to fix the unit test issue and add a couple of cmdstanpy tests to verify the built wheels would pass the unit tests for both back ends as expected! I opened up those changes as a PR to this branch.

freddyaboulton:
@tcuongd this looks ready to merge for me. What do you think?

tcuongd:
@freddyaboulton Thanks for the reminder! Totally dropped the ball on this one 😅  Yeah it's probably a good idea to merge some of this into master, so it's easier to extend and test before we do a release

tcuongd:
Alrighty I will merge this into `master` now since it's passing the regular build-and-test CI, and we can iterate on it. I've also set the wheel workflow runs to be triggered on `release create` and `workflow_dispatch` (i.e. manual runs). Since the manual run only becomes trigger-able after this is merged into master, I'll do another check that nothing's broken after the merge.

freddyaboulton:
@tcuongd When is this being released?

octavd:
@tcuongd looking forward for this also.

lmmentel:
Awesome! Looking forward to it. It'll have a lot of impact reducing the install times 🙌

akosfurton:
I believe [this PR ](https://github.com/facebook/prophet/pull/2088) + a follow up PR to remove the PyStan backend should unblock https://github.com/facebook/prophet/issues/2041

Then, prophet 1.1 can be released to PyPi https://github.com/facebook/prophet/issues/2064

freddyaboulton:
@tcuongd Any update on when prophet 1.1 will be released?

akosfurton:
> @tcuongd @akosfurton Any update on when prophet 1.1 will be released? Thank you!

I don't have write access to merge the PR, only @tcuongd and @bletham do. Will try to contact them.

PR #2013, opened 2021-08-31
Import extraDistr's rlaplace explicitly

laresbernardo:
Hi fellow FBers. I propose explicitly importing `extraDistr`'s `rlaplace` to avoid "No extraDistr package imported" error. Specifically, when running in CentOS8. The alternative is to load `extraDistr` (which IS a dependency so it's actually installed) first and then `prophet` to avoid this issue as it is.


tcuongd (commented):
Thanks for this! Could you possibly give a bit more context about where you're seeing the error? Or is there a more informative error log?

I'm asking because we would only need to have `importFrom` statements if we intended to use functions without calling their corresponding package. For example, we have `importFrom(dplyr, "%>%")` so that we can use `%>%` throughout the code without having to write `dplyr::"%>%"` every time. 

For the `laplace` function, a quick search shows that we do write the full name `extraDistr::laplace()` in all places, so it shouldn't be causing any issues. I'm not sure if these changes would fix the problem you originally saw 🤔

PR #2026, opened 2021-09-14
Update hdays.py

conradtan:
Included Singapore (SG) Holidays

### DIFF
--- a/python/prophet/hdays.py
+++ b/python/prophet/hdays.py
@@ -1538,3 +1538,147 @@ class Georgia(HolidayBase):
 
 class GE(Georgia):
     pass
+
+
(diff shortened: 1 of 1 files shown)


tcuongd:
Hey thanks for this! Just looking at https://github.com/dr-prodigy/python-holidays#available-countries, Singapore seems to be available in the `holidays` package. Could you try to upgrade to the latest `holidays` version (`0.11.3.1`) and see if you can add SG holidays?

If there are any you feel are missing, it's best to make a PR directly to the `holidays` package. We will soon be working on removing `hdays.py` altogether, it's kind of a hacky patch to fill some gaps in `holidays` (which has been developed a lot over the years).

PR #2027, opened 2021-09-15
Update setup.py

anukaal:
updated classifiers


tcuongd (approved):
Thanks for this! I've just removed the Python 3.9 classifier for now since `pystan 2.19.1.1` is not available for Python 3.9. We will eventually remove `pystan` altogether and use `cmdstanpy` instead (which will enable us to build wheels for python 3.9), that will probably come with the next release.

tcuongd:
Checks are failing (my fault), we'll just need to merge with master once this PR is in: https://github.com/facebook/prophet/pull/2040 then it should be all good.

anukaal:
> Checks are failing (my fault), we'll just need to merge with master once this PR is in: #2040 then it should be all good.

Yeah it would be fine ig !

PR #2044, opened 2021-10-07
hdays: remove countries defined in python-holidays

ourownstory:
To simplify maintenance, this PR does:
- remove countries defined in python-holidays in hdays.py
- remove now unused dependencies in hdays.py

I also created an issue at python-holidays to add the last few missing countries:
https://github.com/dr-prodigy/python-holidays/issues/538

I did run all tests locally. passed. 

Did not change documentation, as nothing changed functionality-wise.

### DIFF
--- a/python/prophet/hdays.py
+++ b/python/prophet/hdays.py
@@ -160,6 +160,253 @@ class ID(Indonesia):
     pass
 
 
+# ------------ Holidays in India---------------------
+class India(HolidayBase):
+    """
+    Implement public holidays in India
+    Reference:
+    https://en.wikipedia.org/wiki/Public_holidays_in_India
+    Please note:
+    India is a culturally diverse and fervent society, celebrate various
+    holidays and festivals. We only implement the holidays that **all states
+    and territories** celebrate.
+    """
+
+    def __init__(self, **kwargs):
+        self.country = "IN"
+        HolidayBase.__init__(self, **kwargs)
+
+    def _populate(self, year):
+        # --------------------------------
+        # Three national days
+        #     Republic Day
+        #     Independence Day
+        #     Gandhi Jayanti
+        # --------------------------------
+        # Republic Day
+        name = "Republic Day"
+        self[date(year, 1, 26)] = name
+
+        # Independence Day
+        name = "Independence Day"
+        self[date(year, 8, 15)] = name
+
+        # Gandhi Jayanti
+        name = "Gandhi Jayanti"
+        self[date(year, 10, 2)] = name
+        # --------------------------------
+        # Hindu holidays
+        #     Diwali
+        #     Holi
+        # --------------------------------
+
+        # Diwali, Holi
+        # http://www.theholidayspot.com/diwali/calendar.htm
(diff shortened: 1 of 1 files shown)


ourownstory:
Any updates?

tcuongd:
Sorry @ourownstory I totally missed this, thanks so much! I will have a look in the coming week

tcuongd:
CI checks are passing which is great. I did a quick check to see if the number of unique holiday names would be roughly equal (or higher) using the `holidays` package. Here's the comparison (note it includes the few countries still yet to be added to `holidays`):

```
	original	new
AE	7	21
BD	7	7
BR	8	13
BY	9	9
CN	7	7
EG	10	18
GE	17	17
ID	14	14
IN	22	8
MY	13	41
PH	12	12
PK	13	13
RU	10	9
TH	15	15
TU	11	11
VN	6	12
```

The only thing that's worrying to me is that India now only has 8 distinct holiday names, compared to the 22 we had. `make_holidays_df()` for 1995-2045 also returns 342 rows now instead of 1,096. The holidays that seem to be missing are:

```
'All Saints Day',
'All Souls Day',
'Boxing Day',
'Christmas Day',
'Day of Ashura',
'Easter Sunday',
'Eid al-Fitr',
'Feast of Holy Family',
'Feast of Pentecost',
'Feast of the Blessed Virgin',
'Feast of the Sacrifice',
'Fest of St. Theresa of Calcutta',
'Good Friday',
'Maundy Thursday',
'Mawlid',
"New Year's Day",
'Palm Sunday'
```

Unfortunately I won't be able to dig into this now, but would it be possible to:
* Raise an issue / question about this in the python-holidays repo. We may end up being able to just move the logic in `hdays.py` to that package.
* Keep India in `hdays.py` for now, until it's rectified in python-holidays

Keen to merge this PR in after that!

tcuongd (changes requested):
In short I think we should keep India in `hdays.py` for now, happy to merge this PR after that adjustment :)

ourownstory:
@tcuongd Thank you for running the tests and checks!
Great suggestions. I added a note about India to my issue on [dr-prodigy/python-holidays](https://github.com/dr-prodigy/python-holidays/issues/538)  

Will re-add India to my current PR, so we can merge -  as the issue on the other repo may take a while to be implemented.

ourownstory:
@tcuongd holidays for India have been re-added.

tcuongd (approved):
Thanks so much and sorry for the late review! Keen to work with the holidays package to eventually align on all the other holidays as well.

tcuongd:
Note: next we'll want to re-generate the R holidays data file.

tcuongd:
Ah, unfortunately since the R package data can't have UTF-8 chars so we need to convert them to ascii, but some characters can't be converted. The countries that had holidays which failed were:

```
Failed to parse utf-8 holidays for:
GE
RU
```

It's only two countries so I'm inclined to just leave them out of the R package data, but I'm also conscious that people may have existing forecasts that would change. I'll probably raise a PR later to add the manual GE and RU holidays back.

Issue #2041, opened 2021-10-04
Remove `pystan` backend, move to `cmdstanpy`


tcuongd:
### Context

Pystan 3+ makes some breaking changes to Prophet python package:
* MAP estimation is removed.
* Windows support is removed.

We've been continuing to use Pystan 2.19.1.1, but this is only supported up to Python 3.8 and no longer receives development. Many of our issues also relate to pystan issues, which will become harder to debug over time and may need hacky patches.

The current installation process with pystan is as follows:

**Mac OS / Linux**

* End users usually use prophet from PyPi.
* End user must install `pystan=2.19.1.1` first, as it is required for building the wheel
* End user installs `prophet`, which includes compiling the pystan model on their machine.

**Windows**

* End users usually use prophet from `conda-forge`.

### Proposal

* Use `cmdstan` + `cmdstanpy` to build binary distributions for Pypi (using Github Actions) and conda-forge, across all platforms. 
  * Packaging pre-compiled stan models per platform + python version means the end-user does not need to perform the build process on their machine. They won't even need their own `cmdstan` installation since we will package the necessary executables into Prophet itself. 
  * Only the lightweight wrapper library `cmdstanpy` is required as a dependency.
* Remove the dependency on `pystan` altogether.

This should **not** change how end-users typically interact with the Prophet package. The general flow of

```python
m = Prophet()
m.fit(df)
future = m.make_future_dataframe()
m.predict(future)
```

will remain unchanged, and all other diagnostic functions will work as normal. 

It's just that behind the scenes, the model object will now be a `cmdstanpy` object rather than a `pystan` one. The argument `stan_backend='PYSTAN'` will also be deprecated. This will affect end users who retrieve the actual `pystan` object and work with it directly, but we expect this to be the minority.

We have already refactored the [setup.py](https://github.com/facebook/prophet/blob/master/python/setup.py) script to accomodate this new process (it currently supports both `pystan 2.19.1.1` and `cmdstanpy` for backwards compatibility). 
We have also [built and tested PyPi wheels](https://github.com/facebook/prophet/actions/runs/1299679630) for macOS and Linux for Python versions 3.6-3.8. If you follow the instructions [here](https://github.com/facebook/prophet/issues/664#issuecomment-913161698), you can test these binary distributions on your machine.

### Work required

For the next minor release (prophet 1.1.0), we should consider the following:

- [ ] Update existing conda build scripts to use the `cmdstan` + `cmdstanpy` backend.
[...]

Issue #2064, opened 2021-11-10
Release a new version to pypi


akosfurton:
Hello @tcuongd or @bletham ,

I noticed that there has not been a release of prophet to PyPi in ~6 months.

In that time, https://github.com/facebook/prophet/pull/2010 has been merged to create Python Wheels for PyPi, which makes the installation experience significantly easier from a `pip install prophet` command.

Could you please upload these wheels to PyPi?

tcuongd:
Yeah sorry we've dropped the ball on this one. Ben's been busy and I haven't had time to finish https://github.com/facebook/prophet/issues/2041

It's a bit hard to give an ETA since holiday season is coming up but I'll at least get the components of issue 2401 ready and let Ben know.

ocardia:
Any news on this? 👋🏻

tboddyspargo:
I was really excited to see #2041 and #2088 get completed, but I was hoping that there might be a forthcoming wheel released to PyPI soon that includes those changes. It would help my team considerably to be able to drop the `pystan` dependency.

@bletham, @tcuongd - Do you think a new wheel release on PyPI will be possible soon? Thanks for all your work on this!

nlhkh:
I am also waiting for the new release. There are lots of good commits that have not been made available in the last year.

akosfurton:
@winedarksea, can you reach out to @bletham on Meta internal chat about how to publish releases to Pypi?

cc: @orenmatar

akosfurton:
https://github.com/facebook/prophet/pull/2202

tcuongd:
Thanks everyone for your patience and help! With this release: https://github.com/facebook/prophet/releases/tag/v1.1
we now have wheels on PyPI (https://pypi.org/project/prophet/).

PR #2088, opened 2021-12-20
Simplify setup.py, make CmdStan the default backend

WardBrian:
Working wheels (cmdstanpy only): https://github.com/WardBrian/prophet/actions/runs/1855227353

This WIP PR would be the next step for #2041. 

Currently it:
1. Makes the repackaging of CmdStan reproducible (doesn't assume an existing installation), and optional (can be disabled with environment variable `PROPHET_REPACKAGE_CMDSTAN=false`)
2. Checks to see if cmdstan was repackaged before setting the cmdstan path
2. Simplifies `setup.py` accordingly
3. Makes `CMDSTANPY` the default backend
4. Removes the distinction between the unix and windows model, as cmdstanpy can compile the "unix" model on all platforms

I'd like to test the current setup and wheel builds, hence this PR and the changes to the github actions files. I am still waiting to hear back from my company about the CLA, so this might sit as a draft until after the holidays.

Still yet to be done:
5. Refactor to remove PyStan entirely


tcuongd:
Thanks for starting this Brian! Happy to help where I can. I've just merged in some changes to `setup.py`: https://github.com/facebook/prophet/pull/2089 that allow building on Windows using `cmdstanpy==0.9.68`. I'm conscious that we can simplify + refactor setup.py quite a bit in this PR, but at least it gives us a working example to compare against.

WardBrian:
@tcuongd I've merged to try to maintain the changes in that PR, but alongside these repackaging changes.

Have you tested that it is not necessary to have the toolchain installed on end-user's machines? I'm not sure how much is statically linked

WardBrian:
@tcuongd are you able to allow workflow runs for this PR so I can continue testing?

tcuongd:
> Have you tested that it is not necessary to have the toolchain installed on end-user's machines? I'm not sure how much is statically linked

Hmm are you specifically talking about the Rtools toolchain on Windows, or more generally for Linux and MacOS as well?

In the `wheel.yml` workflow, we:

* Build the package in an isolated environment
* Then install the built wheel into the global python environment, and run the tests in the `tests/` folder.

The second step would just use all the pre-packaged code, so I _think_ if there's any issues these tests would fail. Would definitely appreciate a second look at this to confirm though.

The MacOS and Linux packages seem to work fine testing on my machines, but I'm not sure whether the Windows package would work for someone who doesn't have Rtools installed. We'd want to call out that prerequisite pretty clearly, although I don't think it needs to be automated as a part of the installation.

WardBrian:
Yeah, specifically I wonder if someone on windows needs to have a Windows-compatible gcc/glibc. I can try to check once I get these wheels building

WardBrian:
@tcuongd - I’m not very familiar with the test suite, could you help figure out why this test is failing? https://github.com/facebook/prophet/runs/4768342262?check_suite_focus=true

Otherwise we passed on Ubuntu. It looks like at least macOS failed because of a GitHub rate limit downloading cmdstan

tcuongd:
> could you help figure out why this test is failing?
Yeah sure I can take a look. Really weird that it's failing though if we haven't changed any other code.

> GitHub rate limit downloading cmdstan
Ah yeah, I remember now, this is why I opted to download cmdstan via `wget` - it's much less fickle and faster as well.

WardBrian:
For me, the same test fails on `main`

WardBrian:
@tcuongd I'm happy to continue working on this if you get a chance to debug that test

.github/workflows/wheel.yml
```diff
@@ -146,31 +141,15 @@ jobs:
           python -m pip install --upgrade pip
           python -m pip install cibuildwheel build
 
-      - name: "Restore cmdstan cache"
```
akosfurton, line 149: I believe these lines are causing the rate limiting. They should not be removed and cmdstan should be downloaded via wget

WardBrian:
I have not looked into them besides confirming they occur on `main` if I use the cmdstanpy backend.

.github/workflows/wheel.yml
WardBrian: I suspect you're correct here. Downloading via wget makes `setup.py` much more complicated if we care about doing the right thing for users who didn't pre-download cmdstan, though. I'll look into it, though I think the test failures are not related to this

WardBrian:
This is the kind of error we have seen when multiple cmdstan processes are trying to write to the same file at the same time, usually in threaded environments where the chain ID is not being explicitly set. Does that sound like a plausible explanation for this error @malmashhadani-88 ?

akosfurton:
@WardBrian , exactly as hypothesized.

The test is running multiple models in parallel (cross-validating the time series), so all of the output files were being written to the same location.

By adding `output_dir = mkdtemp(),` to line 96 of models.py, all tests appear to pass

WardBrian:
@malmashhadani-88 glad it was a simple fix. Could you be more specific about what I should change in the test?

WardBrian:
@akosfurton I believe I have made the change as you describe

.github/workflows/wheel.yml
akosfurton: ^ These don't appear to be an issue anymore. All CI builds fail on the same test

README.md
```diff
@@ -61,44 +61,22 @@ If you have custom Stan compiler settings, install from source rather than the C
...
-
-It is also possible to install Prophet with two backends:
-
-```bash
-# bash
-$ CMDSTAN=/tmp/cmdstan-2.22.1 STAN_BACKEND=PYSTAN,CMDSTANPY pip install prophet
-```
+The default dependency that Prophet has is `cmdstanpy`.
+When installed using Pip, Prophet will install its own version of CmdStan.
+If this is undesired, you can set the environment variable `NO_REPACKAGE_CMDSTAN` when building from source.
```
tcuongd, line 69: `NO_REPACKAGE_CMDSTAN=True` ?

README.md
```diff
@@ -61,44 +61,22 @@ If you have custom Stan compiler settings, install from source rather than the C
...
-
 ### Anaconda
 
-Use `conda install gcc` to set up gcc. The easiest way to install Prophet is through conda-forge: `conda install -c conda-forge prophet`.
+The easiest way to install Prophet is through conda-forge: `conda install -c conda-forge prophet`.
 
 ### Windows
 
-On Windows, PyStan requires a compiler so you'll need to [follow the instructions](https://pystan2.readthedocs.io/en/latest/windows.html). The easiest way to install Prophet in Windows is in Anaconda.
+The easiest way to install Prophet in Windows is in Anaconda.
```
tcuongd, line 79: We should add something here about being able to install from `pip` as well, but that will firstly require installing the `RTools` toolchain (e.g. via `cmdstanpy`).

python/prophet/models.py
```diff
@@ -63,28 +65,16 @@ class CmdStanPyBackend(IStanBackend):
...
-        )
+        local_cmdstan = pkg_resources.resource_filename("prophet", f"stan_model/cmdstan-{self.CMDSTAN_VERSION}")
+        if os.path.exists(local_cmdstan):
+            cmdstanpy.set_cmdstan_path(local_cmdstan)
 
     @staticmethod
     def get_type():
         return StanBackendEnum.CMDSTANPY.name
 
-    def _add_tbb_to_path(self):
```
tcuongd, line 74: Just wondering if you've tested the pip Windows install on your machine (you can download the wheel)? I added this because I was getting an error about TBB when trying to load the model.

python/requirements.txt
```diff
@@ -1,6 +1,5 @@
 Cython>=0.22
-cmdstanpy==0.9.77
-pystan~=2.19.1.1
```
tcuongd, line 3: If we're removing this should we catch import errors in parts of the code which use pystan? With a check for the version as well. Otherwise happy to leave this in for now and do that in another PR.

python/setup.py
```diff
@@ -20,91 +17,17 @@
 from setuptools.command.develop import develop
 from setuptools.command.test import test as test_command
 
-PLATFORM = "unix"
-if platform.platform().startswith("Win"):
-    PLATFORM = "win"
 
-MODEL_DIR = os.path.join("stan", PLATFORM)
```
tcuongd, line 27: Yeah I'm not sure what the history is behind having different model code for Windows, but if the compilation and tests pass on Windows then happy to make this simplifying change.

python/setup.py
```diff
@@ -186,24 +95,30 @@ def build_cmdstan_model(target_dir):
...
+
     # Clean up
     for f in Path(MODEL_DIR).iterdir():
         if f.is_file() and f.name != model_name:
             os.remove(f)
-    prune_cmdstan(cmdstan_dir)
+
+    if not os.environ.get("NO_REPACKAGE_CMDSTAN", False):
+        prune_cmdstan(cmdstan_dir)
+        pass
```
tcuongd, line 113: Is this redundant?

README.md
tcuongd: Hmm since this is specific to Prophet I wonder if we can call it `PROPHET_REPACKAGE_CMDSTAN` and have the default as `True` instead (makes the code easier to read I think).

python/setup.py
```diff
@@ -129,50 +53,35 @@ def prune_cmdstan(cmdstan_dir: str) -> None:
...
+        if platform.platform().startswith("Win") == "win":
+            try:
+                cmdstanpy.utils.cxx_toolchain_path()
+            except Exception:
+                install_cmdstan_toolchain()
+        print("Installing cmdstan to", cmdstan_dir)
+        if os.path.isdir(cmdstan_dir):
+            rmtree(cmdstan_dir)
+
+        if not cmdstanpy.install_cmdstan(
```
tcuongd, line 77: Neat!

python/setup.py
```diff
@@ -129,50 +53,35 @@ def prune_cmdstan(cmdstan_dir: str) -> None:
...
-            install_cmdstan_toolchain()
-
-    with cmdstanpy.utils.pushd(cmdstan_dir):
-        clean_all_cmdstan()
-        build_cmdstan()
-    cmdstanpy.set_cmdstan_path(str(cmdstan_dir))
+    from multiprocessing import cpu_count
+
+    if not os.environ.get("NO_REPACKAGE_CMDSTAN", False):
+        if platform.platform().startswith("Win") == "win":
```
tcuongd, line 68:
```suggestion
        if platform.platform().startswith("Win"):
```
?

tcuongd (commented):
This is looking great so far, thanks heaps. I've just done a first pass and haven't looked at the build / test results just yet, but will do that soon. Just left a few questions / suggestions, mostly small.

tcuongd:
> For the Python 3.7 tests, the specific test test_diagnostics.test_cross_validation fails
> 
> The failure is due to this line:
> 
> params = self.stan_to_dict_numpy( self.stan_fit.column_names, self.stan_fit.optimized_params_np )
> 
> which fails because the values in these variables are incorrectly parsed from the cmdstan CSV output.
> 
> The cmdstan CSV output looks like it did not output completely correctly. It is missing a line-break after the parameter names and their values.
> 
> The CSV has both the parameter names and values on the same line, which is not the format the cmdstanpy expects.
> 
> Sample CSV attached below: [prophet_model-20220211103525.csv](https://github.com/facebook/prophet/files/8050040/prophet_model-20220211103525.csv)
> 
> Environment:
> 
> * Python 3.7
> * Test: test_diagnostics.test_cross_validation
> * Test only fails on method=Newton and fails when the cross validation is parallelized using ThreadPoolExecutor or ProcessPoolExecutor
>   
>   * Possibly a race condition with the csv file names only being unique to the second?
> 

Amazing find, thank you @malmashhadani-88!

python/requirements.txt
akosfurton: There are a few references to pystan in `setup.py` and `models.py`

They appear straightforward to remove. This would mean that not only is `cmdstanpy` the default backend, but also that `pystan` will not be supported at all

python/prophet/models.py
WardBrian: I was waiting for the other test issues to be resolved before trying on my windows machine. Doing that now.

python/requirements.txt
WardBrian: I can remove PyStan all together, but the main reason the requirements changed here is so I could do local testing and it snuck into a commit. 

I think it would be worth a separate PR to remove pystan, as it can also do a lot of cleanup/simplifying of the code

python/prophet/models.py
WardBrian: I believe I've located the windows issue

akosfurton:
Looks great to me. 

Agree on the 2 next steps, and then should be ready to merge.

I believe this PR + a follow up PR to remove the PyStan backend should unblock https://github.com/facebook/prophet/issues/2041 

Then, prophet 1.1 can be released to PyPi https://github.com/facebook/prophet/issues/2064

WardBrian:
I'm also happy to put together the PR which removes the pystan dependency. Would the project like to maintain the IStanBackend abstraction or just collapse everything down to the CmdStanBackend class?

akosfurton:
What else is left before we can merge this branch?

WardBrian:
Someone needs to approve the workflow runs to make sure tests do pass (I've confirmed locally on Ubuntu and Windows, but still good to check)

I will then do one final commit which re-adds pystan to requirements and turns off the wheel building GHA for PRs, at which point it can be merged.

WardBrian:
Hi @tcuongd - sorry for the silly error. Would you be able to run the workflows again now?

WardBrian:
Looks like the most recent run failed on a duplicate copy of the windows build, but suceeded on the main one!
https://github.com/facebook/prophet/actions/runs/1853320460

I've removed the duplicate and we should be good to go. It's currently only building wheels for py36,37, and 38 - I assume that 39 and 3.10 will also be desired?

WardBrian:
I can confirm the Linux wheel works on Ubuntu 20.04, Python 3.8.12, _without CmdStan installed separately_

WardBrian:
Running on my fork here: https://github.com/WardBrian/prophet/actions/runs/1855227353

WardBrian:
I've just released adding back pystan will prevent building on newer versions. I'll run it [once](https://github.com/WardBrian/prophet/runs/5223171372?check_suite_focus=true) to test and them go back to only building up to 3.8 in the final commit.

akosfurton:
> I've just released adding back pystan will prevent building on newer versions. I'll run it [once](https://github.com/WardBrian/prophet/runs/5223171372?check_suite_focus=true) to test and them go back to only building up to 3.8 in the final commit.

Good point. You can add 3.9 and 3.10 after removing pystan

WardBrian:
https://github.com/WardBrian/prophet/actions/runs/1855227353#artifacts Wheels for 3.6-3.10 on all 3 platforms. Pushing final commit for this PR now, @tcuongd if you could review I believe this is ready to merge

WardBrian:
@tcuongd - is there anything else you'd like to see done in this PR before merging?

akosfurton:
@bletham , could you review this PR as well? It is ready to merge

CJSmith-0141:
+1 on this, this PR solves an issue I'm having with python 3.10 and prophet

ShantanuKumar:
Hi, I tried to use these changes to train model using prophet but I get this error about missing `libtbb.dylib`

WardBrian:
Can you share the error?

ShantanuKumar:
This is generated on macos. I was able to fix it inside a docker container with linux by manually installing `tbb`. I tried to do the same on macos using `brew install tbb` but that doesn't work.
```
INFO:prophet:Disabling yearly seasonality. Run prophet with yearly_seasonality=True to override this.
INFO:prophet:Disabling daily seasonality. Run prophet with daily_seasonality=True to override this.
INFO:cmdstanpy:start chain 1
INFO:cmdstanpy:finish chain 1
WARNING:prophet.models:Optimization terminated abnormally. Falling back to Newton.
INFO:cmdstanpy:start chain 1
INFO:cmdstanpy:finish chain 1
  File "/Users/developer/.pyenv/versions/abb-crc-env/lib/python3.9/site-packages/prophet/models.py", line 105, in fit
    self.stan_fit = self.model.optimize(**args)
  File "/Users/developer/.pyenv/versions/abb-crc-env/lib/python3.9/site-packages/cmdstanpy/model.py", line 435, in optimize
    raise RuntimeError(msg)
RuntimeError: Error during optimization.
RunSet: chains=1
 cmd:
        ['/Users/developer/.pyenv/versions/3.9.0/envs/abb-crc-env/lib/python3.9/site-packages/prophet/stan_model/prophet_model.bin', 'random', 'seed=36598', 'data', 'file=/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/tmpffcs4eo7/j7fjjyax.json', 'init=/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/tmpffcs4eo7/rs05hk_o.json', 'output', 'file=/private/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/tmpxll2ozny/prophet_model-202203180948-1.csv', 'method=optimize', 'algorithm=lbfgs', 'iter=10000']
 retcodes=[-6]
 error_msgs:
        /private/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/tmpxll2ozny/prophet_model-202203180948-1-stderr.txt
chain_id 1:
dyld[30142]: Library not loaded: @rpath/libtbb.dylib
  Referenced from: /Users/developer/.pyenv/versions/3.9.0/envs/abb-crc-env/lib/python3.9/site-packages/prophet/stan_model/prophet_model.bin
  Reason: tried: '/private/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/pip-install-oqo49qox/prophet_deddf0531f6c41cf8980f3b9563b7023/python/build/lib.macosx-11.5-x86_64-3.9/prophet/stan_model/cmdstan-2.26.1/stan/lib/stan_math/lib/tbb/libtbb.dylib' (no such file), '/private/var/folders/9w/9_4n35r57d707zkrk86rvh9w0000gn/T/pip-install-oqo49qox/prophet_deddf0531f6c41cf8980f3b9563b7023/python/build/lib.macosx-11.5-x86_64-3.9/prophet/stan_model/cmdstan-2.26.1/stan/lib/stan_math/lib/tbb/libtbb.dylib' (no such file), '/usr/local/lib/libtbb.dylib' (no such file), '/usr/lib/libtbb.dylib' (no such file)
```

WardBrian:
I will investigate this. In the mean time, you may be able to use `brew install tbb@2020` to get it working

WardBrian:
@ShantanuKumar were you using the downloads from the link I posted above (https://github.com/WardBrian/prophet/actions/runs/1855227353)?

ShantanuKumar:
@WardBrian I build the library locally from the commit without `pystan` on your forked branch.
```
pip install git+https://github.com/WardBrian/prophet.git@18f19f7f989bc38489a1601fa111193f2772dce1#subdirectory=python
```

WardBrian:
Are you able to download the correct wheel from that GithubAction run and try that?

ShantanuKumar:
I only see wheels for 3.8, but I am using python 3.9

WardBrian:
If you download the zip file you will find wheels for many versions

ShantanuKumar:
That worked!

ShantanuKumar:
Can we please merge this?

tcuongd:
Hmm not sure why CI hasn't been running for the more recent pushes, it ran here: https://github.com/facebook/prophet/actions/runs/1853320459
I'd like to kick off another CI check before merging; @WardBrian would you be able to give me push access to your fork? I can usually get CI working again by making a push from my account

tcuongd (approved):
Amazing stuff!!! 

* Great to see that `cmdstanpy.install_cmdstan()` is fast now and we don't need to use `wget`; makes the wheel build workflow so much cleaner.
* The refactor to `setup.py` looks all good to me, again much cleaner when we make use of the `cmdstanpy` internals.
* Thanks so much to everyone for figuring out tricky bugs like the interaction between multiprocessing and writing to csvs, and the Windows TBB stuff.

Happy to merge this PR but please keep in mind:
* I'm unable to release a new package version to PyPI, so for now Prophet on PyPI will remain at 1.0.1, with all of the pystan installation code, and will only support Python 3.6-3.8.
* The changes in this PR will be accessible by cloning the Prophet repo and building the package manually from source (this is effectively the 'development version', which I've set to 1.1). I've added a section for how to do this in the README. (ps: @WardBrian could you have a read of these changes and let me know if it makes sense) 
* If you don't want to build from source, but want to use Prophet without worrying about `pystan`, please download the latest wheels here: https://github.com/WardBrian/prophet/actions/runs/1855227353 -- the downloaded wheel files can be installed using `pip`.
 
I understand this is confusing but 🤞 we can get a PyPI release soon (then everyone will _actually_ be able to "just pip install prophet", heh).

thomasaarholt:
(been following this PR with quite some interest)
What is the blocker for a PyPi release?

tcuongd:
> (been following this PR with quite some interest) What is the blocker for a PyPi release?

I think Ben (one of the package authors) has the release process set up on his machine, but I haven't been able to get in touch with him recently.

WardBrian:
@tcuongd the README changes look good to me. I'm actually not 100% sure if the final section ("Using `cmdstanpy` with Windows requires a Unix-compatible C compiler such as mingw-gcc.") is true with the pre-compiled installations from wheels, but it certainly is if building from source.

Shame about the PyPi releases. This is part of why on CmdStanPy we do releases through and [action](https://github.com/stan-dev/cmdstanpy/blob/cc6daf27c0d5032eb3dbb0204df5991e32f64210/.github/workflows/release.yml#L93)

Issue #2063, opened 2021-11-10
Error:   Holidays in France are not currently supported!


aLeXgr2020:
I'm using R and want to fetch built-in country holidays for France. I tried the below:

```
fit_prophet <- add_country_holidays(fit_prophet, country_name = "FR")
fit_prophet <- add_country_holidays(fit_prophet, country_name = "FRA")
fit_prophet <- add_country_holidays(fit_prophet, country_name = "France")
```
but I'm getting the below error:

`Holidays in FR are not currently supported!
`

aLeXgr2020:
The problem occurs in Python's version of Prophet, but it doesn't produce an error there. 
It simply doesn't add any country holidays when `country_name` is `FR` , `FRA` or `France`.

onlykimu:
also Bangladesh is not included

PR #2094, opened 2022-01-04
Regenerate generated_holidays.rda

andrew-pollock:
PR #1902 updated R/data-raw/generated_holidays.csv to include additional holidays (including those for France) but didn't run generated_holidays.R, so generated_holidays.rda was never updated. `get_holiday_names()` in R selects from generated_holidays.rda so updating the .csv alone doesn't fix the missing holidays in R.

This PR just runs generated_holidays.R to regenerate /data/generated_holidays.rda based on the newer csv file.


andrew-pollock:
This fixes issue #2063 if you install the dev version of the package.

tcuongd:
Ohh my bad! Great catch and thank you for fixing this :)

tcuongd (approved):

onlykimu:
also  Bangladesh, 8th most populated country is no included

PR #2105, opened 2022-01-22
Typo inthe Contributing Markdown File. 

jgruber99:
There was a typo in the documentation. When one wants to run tests. Instead of "test", the documentation said "tests",  which is not the right command, and caused confusion for me.


tcuongd (approved):
Might need to update these instructions in the future to use `tox` or `unittest` directly since `setup.py test` has been deprecated: https://setuptools.pypa.io/en/latest/userguide/commands.html#test-build-package-and-run-a-unittest-suite

PR #2123, opened 2022-02-24
Fix for Windows if using PyStan

peeveen:
Prophet won't build on Windows+PyStan without some extra_compile_args passed to StanModel.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -7,6 +7,7 @@ import os
 import pickle
 import platform
 import subprocess
+import struct
 import sys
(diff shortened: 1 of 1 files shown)


akosfurton:
Thanks @peeveen for your contribution

This PR https://github.com/facebook/prophet/pull/2088 compiles the model during build time and ships it pre-packaged with the Prophet library.

This would remove any Windows / Unix specific differences going forward.

peeveen:
Fair enough, i'll close this.

Issue #2132, opened 2022-03-07
Calling `cross_validation(model)` without `model.fit` throws AttributeError


ivanmkc:
Calling `cross_validation(model)` without `model.fit` throws AttributeError.

```
df = model.history.copy().reset_index(drop=True]()

AttributeError: 'NoneType' object has no attribute 'copy']()
```

tcuongd:
For 1), I agree that there should be a clearer error message.

For 2), it does seem unintuitive at first (typically the ML workflow is train on subset -> validate -> fit on entire history), but it seems the cross_validation function was designed to:
* Only take a `model` argument and not a `dataframe` argument.
* Pick the largest seasonal period in the data. When the end user instantiates a `Prophet()` object, they may not set seasonalities yet, and these may be inferred later during the fitting process.
We can probably scope out a clearer API but just need to keep those two things in mind.

PR #2140, opened 2022-03-23
adding no history error handling to cross_validation

jul-carras:
This is in response to #2132. I mimicked the history checking style from the `.predict()` method:

https://github.com/facebook/prophet/blob/f16c782661c33440331205d5f6455995e3a037f8/python/prophet/forecaster.py#L1197-L1198

I'm very much open to changing the language. This first stab was to raise a useful error and to try and explain very simply why `model` needs to be fit first. [ivanmkc](https://github.com/ivanmkc) makes a good point about the concern for temporal leakage and I think it warrants a discussion.


tcuongd (approved):
Oh awesome, thank you for this! Wording makes sense to me :)

PR #2142, opened 2022-04-01
Improved execution time of rolling_mean_by_h

RaymondMcT:
Saw https://github.com/facebook/prophet/pull/1841 had been left hanging for almost a year but it was the right idea. This improves the rolling_mean_by_h function bringing it from O(nw) to O(n+w). Algorithm improvements to both the python and R functions.


tcuongd (approved):
Did some spot checks and this logic gives the same results as the original, so LGTM! Thanks so much for writing this in both R and Python!

PR #2114, opened 2022-02-02
Change deprecated `append` to `concat`

nlhkh:
Pandas `frame.append` method is deprecated. Updating to the newer `pd.concat` method to prepare Prophet for future Pandas, and avoid generating warning log in runtime.

PR #2147, opened 2022-04-05
Pandas Deprecation Warnings Fix

winedarksea:
This is meant as a simple test contribution while I build up to a more serious commit.

Solves the following two warnings in latest pandas while retaining the existing behavior:
```
prophet\python\prophet\forecaster.py:896: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.
  components = components.append(new_comp)

\prophet\python\prophet\serialize.py:156: FutureWarning: The behavior of Timestamp.utcfromtimestamp is deprecated, in a future version will return a timezone-aware Timestamp with UTC timezone. To keep the old behavior, use Timestamp.utcfromtimestamp(ts).tz_localize(None). To get the future behavior, use Timestamp.fromtimestamp(ts, 'UTC')
  setattr(model, attribute, pd.Timestamp.utcfromtimestamp(model_dict[attribute]).tz_localize(None))

```

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -893,7 +893,7 @@ class Prophet(object):
         group_cols = new_comp['col'].unique()
         if len(group_cols) > 0:
             new_comp = pd.DataFrame({'col': group_cols, 'component': name})
-            components = components.append(new_comp)
+            components = pd.concat([components, new_comp], axis=0)
         return components
 
     def parse_seasonality_args(self, name, arg, auto_disable, default_order):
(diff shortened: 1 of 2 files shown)


winedarksea:
@jul-carras could you also add the other changed file to your commit then I can close this one?
to [python/prophet/serialize.py](https://github.com/facebook/prophet/pull/2147/files#diff-04d63838a83d8d46d55563d55b04d5b6681f1f4ff2a824c93b1c3f932c7927a7)

jul-carras:
sorry @winedarksea I didn't see the other fix! 

#2114 was done by @nlhkh, but it might be worth keeping the `Timestamp` update in its own commit, unless the maintainers want to group these future-proofing, pandas-related changes into one.

tcuongd:
Feel free to make the timestamp commit in a different PR, @winedarksea

PR #2148, opened 2022-04-05
Remove pystan, leave backend abstraction for now

### LINKED ISSUE #2041: Remove `pystan` backend, move to `cmdstanpy`

WardBrian:
Closes #2041, follows on from #2088 

This does not refactor the `IStanBackend` class or anything like that yet. It just removes pystan from the requirements.txt, and deletes the imports of it


WardBrian:
After this is merged you the build_wheel can run and build for python 3.6-3.10. Until PyPi is updated that link will be the best bet

jlopezpena:
This is great, I can confirm checking out this PR solves all the build issues on Mac with python 3.9+. It would be great to have this merged, and perhaps a new pypi packaged released, as the fact that the "official" pypi release doesn't work with any recent python is not great. IMO, the code refactoring can be postponed as it will not alter any functionality.

WardBrian:
I think leaving the backend code for now could even be useful if things like #1973 get another look

akosfurton:
This looks great!

Looking forward to this being merged and Prophet released to PyPi again

WardBrian:
@tcuongd are you able to review this? I know there are some permissions issues blocking a PyPi release as well, but this is one of the final roadblocks as I understand

nlhkh:
Thank you! I think Prophet really need this because pystan is really problematic to install.

nicolaerosia:
@tcuongd kind ping :)

tcuongd:
Yep this LGTM, thanks heaps! I'll trigger a `wheel.yml` workflow after the merge to test the 3.9-3.10 wheels as well.

tcuongd:
Successfully built wheels without pystan: https://github.com/facebook/prophet/actions/runs/2275140778
Huge thanks to @WardBrian!

ShantanuKumar:
Does this mean we will soon have a pypi release?

PR #2165, opened 2022-05-05
wheel.yml: Do not build Python 3.6 wheels

tcuongd:
`cmdstanpy >=1.0.1` (requirement introduced https://github.com/facebook/prophet/pull/2148) requires Python >= 3.7: https://github.com/stan-dev/cmdstanpy/releases/tag/v1.0.1

Wheel build test: https://github.com/facebook/prophet/actions/runs/2275140778

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -69,7 +69,7 @@ jobs:
           CIBW_ENVIRONMENT: >
             STAN_BACKEND="${{ env.STAN_BACKEND }}"
             PIP_CACHE_DIR="${{ env.PIP_DEFAULT_CACHE }}"
-          CIBW_BUILD: cp36-* cp37-* cp38-* cp39-* cp310-*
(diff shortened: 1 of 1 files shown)


tcuongd:
@WardBrian I think this is fine, I see a lot of packages requiring Python >= 3.7

tcuongd:
Hmm running into rate limiting issues again while downloading cmdstanpy to build the wheel...

```
 WARNING:cmdstanpy:Cannot connect to CmdStan github repo.
  Cannot connect to github.
  Cannot connect to CmdStan github repo.
  HTTP Error 403: rate limit exceeded
```

WardBrian:
Does this action have a GitHub token set in the environment? I believe that raises the rate limit considerably

tcuongd:
It doesn't, and I think I need elevated permissions to add one. Good to know though -- I'll raise that.

I just hit re-run on the specific job and everything ran fine (we have cmdstanpy-only wheels now, woohoo!). We might just have to keep hitting re-try if the wheel build fails for now.

PR #2178, opened 2022-05-23
Only encode to JSON once when serializing a Python model

### LINKED ISSUE #2133: Documentation suggests double-encoding Python models to JSON
The [additional_topics page](https://facebook.github.io/prophet/docs/additional_topics.html#saving-models) suggests saving/reading a Prophet model in Python like so:

```python
import json
from prophet.serialize import model_to_json, model_from_json

with open('serialized_model.json', 'w') as fout:
    json.dump(model_to_json(m), fout)  # Save model

with open('serialized_model.json', 'r') as fin:
    m = model_from_json(json.load(fin))  # Load model
```

However, this encodes the model to JSON twice over:
*  `model_to_json` encodes the model to a string of JSON
* then `json.dump` encodes this string further when writing the file

This means that the `serialized_model.json` file contains one string only as the root node - everything within it is escaped with backslashes. So the beginning of my file is

```
"{\"growth\": \"linear\", \"n_changepoints\": 25, \"specified_changepoints\": false, \"changepoint_range\": 0.8, \"yearly_seasonality\": false, \"weekly_seasonality\": false, \"daily_seasonality\": false,
...
"
```

Obviously this is ok if we then read in the file using the example code, as it then double-decodes it. But for general use it would be better if the json file were more readable.

Surely instead it would be better to do:

```python
from prophet.serialize import model_to_json, model_from_json

with open('serialized_model.json', 'w') as fout:
    fout.write(model_to_json(m))  # Save model

with open('serialized_model.json', 'r') as fin:
    m = model_from_json(fin.read())  # Load model
```

Or is there a reason why you'd want to double-encode?

tomsutch:
Changed documentation to remove suggestion to double-encode JSON when serializing Python models (and the corresponding changes for loading them again).

Closes #2133

### DIFF
--- a/docs/_docs/additional_topics.md
+++ b/docs/_docs/additional_topics.md
@@ -36,14 +36,13 @@ In Python, models should not be saved with pickle; the Stan backend attached to 
 
 ```python
 # Python
-import json
 from prophet.serialize import model_to_json, model_from_json
 
 with open('serialized_model.json', 'w') as fout:
-    json.dump(model_to_json(m), fout)  # Save model
+    fout.write(model_to_json(m))  # Save model
 
 with open('serialized_model.json', 'r') as fin:
-    m = model_from_json(json.load(fin))  # Load model
+    m = model_from_json(fin.read())  # Load model
 ```
 The json file will be portable across systems, and deserialization is backwards compatible with older versions of prophet.
 
--- a/notebooks/additional_topics.ipynb
+++ b/notebooks/additional_topics.ipynb
@@ -122,14 +122,13 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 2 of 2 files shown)


tcuongd:
Thanks! Usually we'd also need to use the nbconvert script to update the corresponding `.md` file (this is what actually gets rendered on the docs site), but given this doesn't affect any output cells, you can just edit the `.md` file directly: https://github.com/facebook/prophet/blob/main/docs/_docs/additional_topics.md#saving-models

After that this should be good to merge!

tomsutch:
Thanks @tcuongd - sorry I missed that! Hopefully all done now :)

tcuongd (approved):

PR #2185, opened 2022-05-25
Make holidays data internal and exported at the same time

### LINKED ISSUE #2166: [R] `generated_holidays` should be internal
`generated_holidays` dataset is exported as `data`, but at the same time it is used internally. This causes issues while trying to use `prophet` without loading. For instance the following code would throw an error `object 'generated_holidays' not found` (WITHOUT loading `prophet` first!).
```r
prophet::add_country_holidays(prophet::prophet() , "US")
```

This could be problematic even if `prophet` is loaded, if someone overwrites `generated_holidays`.
```r
library(prophet)
generated_holidays <- 1
add_country_holidays(prophet() , "US")
```

I believe the solution would be to make `generated_holidays` internal (parameter `internal = TRUE` in `use_data`).

If dataset is still expected to be avaialble to package users, it could be tagged with `@export` manually (following [this blogpost](https://coolbutuseless.github.io/2018/12/10/r-packages-internal-and-external-data/)).

If this is acceptable approach I could make a PR with required changes.

bartekch:
This PR Fixes #2166 

`generated_holidays` dataset is internal now, so there is no possibility for a user to overwrite data used by `prophet` internal functions. At the same time `export` tag is added, so a user could still use `generated_holidays` directly after loading `prophet`.


koconder (commented):
Looks good

tcuongd (approved):

Issue #664, opened 2018-08-27
Upload wheel to PyPI


xliiv:
Could you upload wheel format to PyPI?

Helpful resource:
* https://pythonwheels.com/
* https://github.com/pypa/twine#using-twine

xliiv:
What if `pystan` also provided wheel?

bletham:
pystan does provide wheels, but you still have to have build tools because Stan models are translated to C++ and compiled on-the-fly.

xliiv:
So there must be a benefit from wheels ;) since `pystan` requires whole build kit and still provides wheels format.. . ;)
Thanks for explanation again :)

bletham:
Yeah, we'll have to consider this more.

DaveGreasley:
Hi, 

I am currently trying to install prophet in a Google Cloud Python Function (https://cloud.google.com/functions/). The creation of the function fails with the following error:

> Deployment failure:
>
>Build failed: {"message": "`pip_download_wheels` had stderr output:\n  Failed building wheel for fbprophet\nERROR: Failed to build one or more wheels\n\nerror: `pip_download_wheels` returned code: 1", "code": 1, "type": "USER_ERROR"}

This is similar to an error I get locally when installing prophet through PIP but it seems to handle it locally and continue with the installation. This could be more of an issue with how GCP are handling errors on install but maybe uploading a wheel would combat this?

chaitan94:
Is this being considered anytime soon?

fbprophet takes quite some time to build, in fact, in one of our application, it almost triples our build times. I am guessing by having a wheel in PyPI, will help improve this by a huge factor. If so, maybe I can try to figure out a way on how to this and make a contribution if that is okay with the maintainers. Any pointers to the same will be appreciated.

CaselIT:
It would be really helpful for windows users, since installing all the required python build tool-chain is tedious and prone to errors. 

In the meantime here's a link to the conda compiled version https://anaconda.org/conda-forge/fbprophet

jgb:
+1 Could this be considered?
fbprophet/pystan is really a pain, it slow everything down and needs extra dependencies on the OS level.
Please consider building a wheel and publishing it to pypi to simplify everyone's life.

bra-fsn:
> +1 Could this be considered?
> fbprophet/pystan is really a pain, it slow everything down and needs extra dependencies on the OS level.
> Please consider building a wheel and publishing it to pypi to simplify everyone's life.

[...]

PR #2190, opened 2022-06-03
Build wheels with native `cibuildwheel`action

abitrolly:
This removes manual `pip caching`, which doesn't seem to speed up builds, because `pip` is not a C++ build system, so it can not trace C++ artifacts. TBH, the best way to benefit from caching is to integrate it into the action.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -95,6 +95,7 @@ jobs:
           CIBW_BUILD: cp37-* cp38-* cp39-* cp310-*
           CIBW_SKIP: "*musllinux*"
           CIBW_ARCHS: native
+          CIBW_BEFORE_ALL: cd python
           CIBW_BUILD_FRONTEND: build
           CIBW_TEST_REQUIRES: pytest
           CIBW_TEST_COMMAND: pytest --pyargs prophet
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -11,12 +11,11 @@ env:
 
 jobs:
   make-wheel-windows:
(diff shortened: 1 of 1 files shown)


abitrolly:
See the result of the run - https://github.com/abitrolly/prophet/actions/runs/2438504455

abitrolly:
Here is also the code to upload wheels to PyPI https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#github-actions (#664)

WardBrian:
This is a great simplification of the workflow

abitrolly:
@WardBrian do you know what should happen now? The repo had been quite silent last week

ShantanuKumar:
By the way @abitrolly I just downloaded the built macos wheels and it doesn't have the Apple silicon (arm64) wheels. Do you think it would be possible to build those too?

WardBrian:
Someone like @tcuongd needs to review this. 

For Mac M1 wheels I believe we would need Mac M1 GitHub action runners, https://github.com/github/roadmap/issues/528

abitrolly:
It looks possible to cross-compile for M1 https://cibuildwheel.readthedocs.io/en/stable/faq/#apple-silicon but it is better to use M1 runners if they are already available. Once this PR is merged, it would be possible to experiment with M1 wheels in either way.

akosfurton:
@tcuongd , can you please review? This looks ready to merge and simplifies builds by a lot

tcuongd (approved):
Yeah this is awesome, thank you so much! Lots of redundant code and I don't think the pip caching is important anyways given we'll only run the wheels workflow once in a while.

Also definitely keen to try the M1 cross compilation, but seems like we'd need someone with an M1 machine to test the wheels.

PR #2192, opened 2022-06-05
chore: light refactor and logic simplification

koconder:
Some light python refactoring of if/else logic and string formatting.


tcuongd (approved):
Thanks :)

PR #2203, opened 2022-06-24
Add source distribution and upload jobs

tcuongd:
Following the guidance provided by cibuildwheel and pypa: 

* https://cibuildwheel.readthedocs.io/en/stable/deliver-to-pypi/#github-actions 
* https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

The upload part is conditional on a release being published. I've tested the workflow end-to-end on my fork running manually (normally we would do this to produce wheels for testing first): https://github.com/tcuongd/prophet/actions/runs/2557799360 and by publishing a release: https://github.com/tcuongd/prophet/actions/runs/2557851926

It has successfully uploaded to TestPyPI :)

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -1,8 +1,8 @@
-name: "Create Python Wheels"
+name: "Create Python distributions and upload to PyPI"
 
 on:
   release:
-    types: [ created ]
+    types: [ published ]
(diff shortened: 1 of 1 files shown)


WardBrian:
This looks like a good solution to me. Very exciting!

PR #2205, opened 2022-06-24
Update Python installation instructions for 1.1

tcuongd:
If https://github.com/facebook/prophet/actions/runs/2558206256 succeeds and 1.1 lands on PyPI, we can update and simplify the documentation for installing Prophet on Python 🎉

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,45 +58,25 @@ If you have custom Stan compiler settings, install from source rather than the C
 
 ## Installation in Python - PyPI release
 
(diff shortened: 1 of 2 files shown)


tcuongd:
@WardBrian I can never select you as a reviewer for some reason, but it'd be great if you could read over this to check I haven't missed anything!

WardBrian:
I believe reviewers are only suggested if they have write access to the repository

WardBrian (commented):
👍 always good to have a reason to delete this much old info

Issue #2231, opened 2022-07-13
Pandas Deprecation: df.astype('timedelta64[ns]')


ghost:
I continue to get the following deprecation warning when I run the plot.py area in my Prophet model: 

```
C:\Users\BLOOB013\Anaconda3\envs\textiles-forecast\lib\site-packages\prophet\plot.py:539: FutureWarning: casting timedelta64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.
  x_plt = df_none['horizon'].astype('timedelta64[ns]').astype(np.int64) / float(dt_conversions[i])
C:\Users\BLOOB013\Anaconda3\envs\textiles-forecast\lib\site-packages\prophet\plot.py:540: FutureWarning: casting timedelta64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.
  x_plt_h = df_h['horizon'].astype('timedelta64[ns]').astype(np.int64) / float(dt_conversions[i])
```

I think there are a few areas where this occurs. It is similar to #2147, with the df.append vs df.concat methods that pandas put out. 

Looking at the pandas documentation, using [.view](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.view.html?highlight=view#pandas.Series.view) should solve our problem. 

I am happy to take this on and make the pull request to fix this, but wanted to bring it up and see if anyone else had some better ideas. Also, how often do we use this method? It could be a fairly extensive fix over a few files. But, this is the only area giving me this warning so far.

PR #2244, opened 2022-08-06
Update plot.py astype

KarateKat95:
Updating plot.py to replace pandas series .astype() method with .view for datetime64 functions, as this functionality is now deprecated. 

Please see more details in #2231 where this issue was raised.

### DIFF
--- a/python/prophet/plot.py
+++ b/python/prophet/plot.py
@@ -521,7 +521,7 @@ def plot_cross_validation_metric(
 
     # Some work because matplotlib does not handle timedelta
     # Target ~10 ticks.
-    tick_w = max(df_none['horizon'].view('timedelta64[ns]')) / 10.
+    tick_w = max(df_none['horizon'].astype('timedelta64[ns]')) / 10.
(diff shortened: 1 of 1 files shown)


python/prophet/plot.py
```diff
@@ -541,8 +541,8 @@ def plot_cross_validation_metric(
         if np.timedelta64(1, dt) < np.timedelta64(tick_w, 'ns'):
             break
 
-    x_plt = df_none['horizon'].astype('timedelta64[ns]').astype(np.int64) / float(dt_conversions[i])
-    x_plt_h = df_h['horizon'].astype('timedelta64[ns]').astype(np.int64) / float(dt_conversions[i])
+    x_plt = df_none['horizon'].view('timedelta64[ns]').view(np.int64) / float(dt_conversions[i])
```
tcuongd, line 544: I think the warning is referring specifically to the second cast, i.e. `.astype(np.int64)`. The first cast to a timedelta shouldn't change.

KarateKat95: Got it - should have just made that edit.

PR #2254, opened 2022-08-26
Enable qemu emulation and aarch64 build for prophet

thechopkins:
Prophet==1.1 does not generate an aarch64 wheel. This adds QEMU emulation for linux builds and adds the additional aarch64 platform for linux. 

Note: This follows the pattern set by [matlplotlib](https://github.com/matplotlib/matplotlib/blob/main/.github/workflows/cibuildwheel.yml#L37-L42) using the [CIBW_ARCHS](https://cibuildwheel.readthedocs.io/en/stable/options/#archs) option.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -94,8 +94,8 @@ jobs:
         find . -name '*.tar.gz' -exec mv '{}' dist/ \;
         find . -name '*.whl' -exec mv '{}' dist/ \;
 
-    # - name: Upload
-    #   uses: pypa/gh-action-pypi-publish@v1.4.2
-    #   with:
-    #     user: ${{ secrets.PYPI_USERNAME }}
-    #     password: ${{ secrets.PYPI_PASSWORD }}
+    - name: Upload
+      uses: pypa/gh-action-pypi-publish@v1.4.2
+      with:
+        user: ${{ secrets.PYPI_USERNAME }}
+        password: ${{ secrets.PYPI_PASSWORD }}
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -11,15 +11,18 @@ env:
 
 jobs:
   make-wheels:
-    name: Make ${{ matrix.os }} ${{ matrix.cibw_archs }} wheels
+    name: Make ${{ matrix.os }} ${{ matrix.cibw_arch }} ${{ matrix.py_version }} wheels
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
         os: ["macos-latest", "ubuntu-latest", "windows-latest"]
-        cibw_archs: ["native"]
-        include:
-          - os: ubuntu-latest
-            cibw_archs: "aarch64"
+        cibw_arch: ["native", "aarch64"]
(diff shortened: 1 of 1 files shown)


tcuongd:
Thanks! Are you able to kick off the workflow manually on your fork to see if it produces the wheels as expected?

thechopkins:
@tcuongd I'm was a bit unsure of the implications of running the workflows from my fork, but I've disabled just the push command from the workflow which should allow them to run isolated from the facebook/prophet workflows, and they are running [now](https://github.com/thechopkins/prophet/actions/workflows/wheel.yml).

thechopkins:
@tcuongd this has completed successfully. The emulated build is slower, but i can make a follow up PR to build those images in parallel

tcuongd:
> @tcuongd this has completed successfully. The emulated build is slower than the native ones (because emulation is slow), but i can make a follow up PR to build those images in parallel if its too painful image

Oh wow yeah 4 hours is ages...sorry I thought if it was like an hour all up it'd be fine to include in the next release but it's way longer than I expected. Is emulation the only way to build these wheels? For releases it doesn't really matter how long it takes, but we'd want to use the build workflow as a test for bigger changes too, and a 4h feedback loop seems too long.

thechopkins:
> Oh wow yeah 4 hours is ages...sorry I thought if it was like an hour all up it'd be fine to include in the next release but it's way longer than I expected. Is emulation the only way to build these wheels? For releases it doesn't really matter how long it takes, but we'd want to use the build workflow as a test for bigger changes too, and a 4h feedback loop seems too long.

@tcuongd A big reason things were taking a long time is because all python versions were being built serially. I've resolved that with the most recent change by incorporating them into the matrix. The long tail for a package build takes ~ 90 mins now which is a little bit longer (~ 15 mins) than the existing build pipeline (though not by much). I'd argue that the increase in time is worth it to get the `aarch64` build. 

There are really two options to get arm64 builds running on GHA. The first is through [emulation w/ QEMU](https://dan-jenkins.co.uk/enabling-native-arm64-builds-with-github-actions/) the second is through [self-hosted actions](https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/).   The first is easier (especially here since the GHAs aren't self-hosted), the second is faster because then `arm64` is actually just `native`.  

Let me know if the new runtime is acceptable and we can get this in. 

Note: Another option is that you can only run the wheels+tests for certain architectures on a standard build but run the full gambit on releases.

tcuongd:
Oh great stuff!! I think this is much better and we can merge this for 1.1.1 :)

tcuongd (approved):

thechopkins:
@tcuongd thanks for merging, it looks like a recent change (introduced at the beginning of the week is now causing this to break).  We can revert and i can take a look at what about the new logic is causing the failure

thechopkins:
@tcuongd actually, I just retried the build on my fork synced with `prophet:master` and it is succeeding, so it might have been an intermitent issue which seems possible since the error message is: `Remote end closed connection without response`

tcuongd:
Yeah sometimes we hit the request limit when downloading cmdstan and we just need to re-run the failed jobs.

PR #2260, opened 2022-09-05
Add example dataset affected by COVID and guide for docs

tcuongd:


docs/nbconvert_template.tpl
```diff
@@ -1,4 +1,4 @@
-{%- extends 'markdown.tpl' -%}
+{%- extends 'markdown/index.md.j2' -%}
```
tcuongd, line 1: For nbconvert >= 6.0

PR #2267, opened 2022-09-16
Update hdays.py

NBekmuratov:
add Kyrgyzstan


NBekmuratov:
Sent a request for CLA Check

NBekmuratov (commented):
(y)

tcuongd (approved):
Thanks heaps! I'll approve this since it's been up for a while already (my bad!) but in the future we should make additions directly to the `holidays` package (it gets released quite often).

PR #2299, opened 2022-11-09
10X faster predict by replacing DataFrame with dict

orenmatar:
Profiling the predict code yielded this result:
before_change

Evidently about 90% of the predict time is spent on either the init of a DataFrame or its __getitem__ method, both are notoriously slow.
A very simple change from dataframe to dict in sample_model and sample_model_vectorized has no risk and implications beyond sample_posterior_predictive since the df is converted to a dict there anyway.
Profiling after the change gives:

after_change

and in runtime it was reduced from ~300ml to ~38ml on a local test

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -1467,10 +1467,10 @@ class Prophet(object):
         sigma = self.params['sigma_obs'][iteration]
         noise = np.random.normal(0, sigma, df.shape[0]) * self.y_scale
 
-        return pd.DataFrame({
+        return {
             'yhat': trend * (1 + Xb_m) + Xb_a + noise,
             'trend': trend
-        })
+        }
 
     def sample_model_vectorized(
         self,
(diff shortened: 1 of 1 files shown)


nicolaerosia:
@tcuongd can we please merge this and release a new version with wheels for python 3.11 as well?

orenmatar:
@tcuongd any updates on this? It should be really easy to integrate and yields a big runtime improvement

tcuongd (approved):
Wow great find, thanks heaps! Yeah I think changing the return type for these internal functions is fine, we just need to make it clear which functions are affected in the release notes.

tcuongd:
For some reason CI didn't kick off, I'll take a look later

tcuongd:
@orenmatar Do you mind just signing the CLA? Thanks!

orenmatar:
@tcuongd Signed it just now

Issue #2229, opened 2022-07-13
'list' object has no attribute 'tolist' 


RiskySignal:
While fit my data by [updating fitted model](https://facebook.github.io/prophet/docs/additional_topics.html), the following error appears
```
Traceback (most recent call last):
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/joblib/externals/loky/process_executor.py", line 436, in _process_worker
    r = call_item()
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/joblib/externals/loky/process_executor.py", line 288, in __call__
    return self.fn(*self.args, **self.kwargs)
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/joblib/_parallel_backends.py", line 595, in __call__
    return self.func(*args, **kwargs)
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/joblib/parallel.py", line 263, in __call__
    for func, args, kwargs in self.items]
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/joblib/parallel.py", line 263, in <listcomp>
    for func, args, kwargs in self.items]
  File "/home/jiangszhang/private_sz/abnormal_detection/abnormal_detection/kafka_comsumer/detection_utils.py", line 84, in get_abnormal_score
    prophet_model.fit(staff_dataframe, init=warm_starting_setting)
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/prophet/forecaster.py", line 1171, in fit
    self.params = self.stan_backend.fit(stan_init, dat, **kwargs)
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/prophet/models.py", line 90, in fit
    kwargs['inits'] = self.prepare_data(kwargs['init'], stan_data)[0]
  File "/dockerdata/jiangszhang/anaconda3/envs/abnormal_detection/lib/python3.7/site-packages/prophet/models.py", line 164, in prepare_data
    'y': data['y'].tolist(),
AttributeError: 'list' object has no attribute 'tolist'
```

After some tries, I found `prophet==1.0.1` works fine but the latest version `prophet==1.1` not.

[...]

PR #2328, opened 2023-01-03
Fix updating fitted model

dlazerka:
As reported in https://github.com/facebook/prophet/issues/2229.


tcuongd:
Thanks :) this will be fixed in https://github.com/facebook/prophet/pull/2335

PR #2334, opened 2023-01-10
Speed up `fourier_series`

yoziru:
This PR speeds up the fourier series computation by:
- speeding up the conversion to days since epoch
- pre-allocating the numpy array and optimizing the for loop

```py
%timeit Prophet.fourier_series(DATA["ds"], 7, 3)
```
`main`
169 µs ± 1.38 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

This PR
30.9 µs ± 699 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

Also added in this PR:
- raise error if `series_order` is < 1
- added typing to the method

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -435,6 +435,9 @@ class Prophet(object):
         -------
         Matrix with seasonality features.
         """
+        if not (series_order >= 1):
+            raise ValueError("series_order must be >= 1")
+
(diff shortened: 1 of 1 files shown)


tcuongd (approved):
Amazing find, thank you! Will definitely include this in the next release. I found the PR super clear as well.

PR #2336, opened 2023-01-11
Simplify setup.py and tests

tcuongd:
Modernize the build system, moving as much information out of `setup.py` as possible and preventing the need to run `python setup.py [command]`, as that is no longer encouraged. Based on @bryevdv 's approach for Bokeh (`https://github.com/bokeh/bokeh/pull/12306`). We'll need to make updates to `setup.py` to enable `py3-none` wheels (mentioned by @WardBrian below) anyways, and this should make it easier to do that.

* Moved most of the project information into `pyproject.toml` apart from `version` which which still reads from `__version__.py`.
* Moved the dependencies list out of `requirements.txt`, and added extras lists for `dev` (building, testing, docs) and `parallel` (for parallel cross-validation).
  * Deletes `requirements.txt`
* Editable installation is now done with `pip install -e .` instead of `python setup.py develop`, and we use the new `editable_wheel` command to do this.
* Test suite is now run with `pytest` instead of `python setup.py test`, which removes the need for the custom test command.
* All docs updated to reflect the new development and testing commands.
* CI now runs checks on Python 3.8 for all platforms: macOS, Ubuntu, and Windows. Should prevent errors like this: `https://github.com/facebook/prophet/pull/2340` from slipping through.


tcuongd:
Something strange is happening in the wheel build workflow for macOS and Windows (errors here: https://github.com/facebook/prophet/actions/runs/3897780760)

* macOS is failing the warm-start tests introduced here: https://github.com/facebook/prophet/pull/2335 -- initial values are being rejected.
* Windows is producing wrong results for the fourier series, after the speed-up change introduced here: https://github.com/facebook/prophet/pull/2334

This happens across all Python versions, 3.7-3.11. I'm not sure how these tests would work fine on Ubuntu but not on mac and Windows...
For the macOS error, the stan program must be producing different sampling outputs compared to Ubuntu.
For the Windows error, numpy must be producing different values??

I'll need to dig into it further on the weekend but @WardBrian @akosfurton @yoziru if you've come across similar issues before let me know.

WardBrian:
Seems like @yoziru has the Windows issue covered. I don't have a Mac to test, unfortunately.

As an aside, it might be worth updating the build so that the wheels are marked with the abi `py3-none`. This means that any python3 version will work, regardless of minor version, since we don't link to the Python API. Reduces the number of wheels build. Here's a [good example](https://github.com/joerick/python-ctypes-package-sample). Cibuildwheel supports this [as of version 2.8](https://cibuildwheel.readthedocs.io/en/stable/changelog/#v280). It builds the wheel once but still tests for all versions you request.

tcuongd:
@WardBrian thanks for that link! It's definitely something I want to do for the next release. I noticed from the example it'd require some changes to `setup.py`, and I've been wanting to clean that up for a while now so wanted to do that first.

I came across how Bokeh structured their build scripts - they have a similar need to compile external libraries and package them in the wheel - and found it quite clean so I've tried to follow that pattern here. Essentially we move all the project metadata to `pyproject.toml` and only keep the necessary custom commands in `setup.py`. Would you have time to take a look? Thanks heaps!

WardBrian:
Moving as much as possible into the pyproject file is a good idea, I think. It looks good here

Issue #2331, opened 2023-01-04
Prophet 1.0


Leep2023:
Hi, I use Prophet 1.0. And I got the following message:

C:\Miniconda3\lib\site-packages\prophet\forecaster.py:896: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.

Do you know why ? Or is there a solution for that ? I wonder if I update to Prophet 1.1, will the problem be solved ?

Best Regards,
Lee

Leep2023:
It looks there is an ongoing task: https://github.com/facebook/prophet/issues/2131

Nil-Andreu:
I think is due to the pandas `append` method, that will be deprecated in the future.

So this method is basically to add in a given dataframe another dataframe. This can be done:

```python
# With append
df.append(df_other)

# With concat
pd.concat([df, df_other])
```

And basically the one that will be supported in the future is the second one.

tcuongd:
Please upgrade to the latest prophet version, this warning has been addressed.

Leep2023:
I followed the advice from tcuongd to upgrade to the latest version, it works fine now. Many thanks !

Issue #2131, opened 2022-03-05
pandas frame.appendis deprecated 


Conet1502:
fbprophet\forecaster.py:891: FutureWarning: The frame.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.

jul-carras:
Looks like there is a [pull request](https://github.com/facebook/prophet/pull/2114) from [nlhkh](https://github.com/nlhkh) for this already. Should be merged in soon.

nlhkh:
The maintainers disappeared. It has been 2 months and no response.

jul-carras:
Could just be a busy time. I've looked at closed pull requests and sometimes it takes a few months to bring them in.

Nil-Andreu:
I think is due to the pandas `append` method, that will be deprecated in the future.

So this method is basically to add in a given dataframe another dataframe. This can be done:

```python
# With append
df.append(df_other)

# With concat
pd.concat([df, df_other])
```

And basically the one that will be supported in the future is the second one.

PR #2337, opened 2023-01-12
[NEW] Upgrade Pandas Method Concat

Nil-Andreu:
The related issues complaining about the `append` method alert that was going to be deprecated in future pandas methods:
- [Issue](https://github.com/facebook/prophet/issues/2331)
- [Issue](https://github.com/facebook/prophet/issues/2131)


tcuongd:
Thanks, but this was addressed is https://github.com/facebook/prophet/pull/2114 which has since been released. v1.1.1 should not show deprecation warnings.

The `.append()`s remaining are appends to `list`s, not dataframes.

PR #2342, opened 2023-01-14
Sanitize warm start parameters

tcuongd:
This is a follow-up to https://github.com/facebook/prophet/pull/2335 which tested fine on Ubuntu and Windows but failed on macOS (see https://github.com/facebook/prophet/pull/2336#issuecomment-1379701487). It turns out that `beta` (i.e. seasonal features + extra regressors) had a different length in each of the initial and new models. Ubuntu and Windows dealt with it (I'm not sure exactly how) but macOS did not. Technically though, we shouldn't be using warm start values from models with different-shaped `delta` (number of changepoints) or `beta`, since the warm values are unlikely to line up with what needs to be fitted in the new model.

The solution proposed here is to sanitize the custom inits: if they don't match the type and shape of the default, we revert to the default. I've done this by adding a new method in `CmdStanPyBackend` and refactoring the fit and sampling methods a little.

I've also added the `warm_start_params` function to the utilities module, to save end users having to copy-paste code from the documentation.

Successful local testing on macOS:

```
python-F3JeUycv ❯ python -m pytest python/prophet/tests/test_prophet.py::TestProphet -k "warm"
======================================================== test session starts ========================================================
platform darwin -- Python 3.8.10, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/cuong/projects/prophet/python
plugins: anyio-3.6.1
collected 40 items / 38 deselected / 2 selected

python/prophet/tests/test_prophet.py ..                                                                                       [100%]

================================================= 2 passed, 38 deselected in 11.11s =================================================
```

### DIFF
--- a/docs/_docs/additional_topics.md
+++ b/docs/_docs/additional_topics.md
@@ -82,11 +82,12 @@ A common setting for forecasting is fitting models that need to be updated as ad
 
 ```python
 # Python
-def get_stan_init(m):
-    """Retrieve parameters from a trained model.
-
-    Retrieve parameters from a trained model in the format
-    used to initialize a new Stan model.
+def warm_start_params(m):
+    """
+    Retrieve parameters from a trained model in the format used to initialize a new Stan model.
+    Note that the new Stan model must have these same settings:
+        n_changepoints, seasonality features, mcmc sampling
+    for the retrieved parameters to be valid for the new model.
 
     Parameters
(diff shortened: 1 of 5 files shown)


tcuongd:
Going to merge once this passes CI since it unblocks https://github.com/facebook/prophet/pull/2336

PR #2343, opened 2023-01-15
Build version-agnostic wheels

tcuongd:
As per @WardBrian 's suggestion here: https://github.com/facebook/prophet/pull/2336#issuecomment-1380484642


tcuongd:
@WardBrian To cross-compile for macOS ARM64 (on an x86_64 machine), do we need to pass a certain setting / CPP flags / headers? The `delocate` command that copies external libraries into the wheel is failing with:
```
delocate.delocating.DelocationError: Some missing architectures in wheel
  Required arch arm64 missing from /Users/runner/work/prophet/prophet/python/build/lib.macosx-10.9-universal2-cpython-38/prophet/stan_model/cmdstan-2.26.1/stan/lib/stan_math/lib/tbb/libtbbmalloc_proxy.dylib
  Required arch arm64 missing from prophet/stan_model/cmdstan-2.26.1/stan/lib/stan_math/lib/tbb/libtbbmalloc_proxy.dylib
  Required arch arm64 missing from prophet/stan_model/prophet_model.bin
```

All good if requires too much work, I'm happy to wait for Github Actions to provide M1 runners and do the compilation natively instead.

WardBrian:
I don’t know if anyone has successfully cross-compiled for M1 using an intel Mac. I’m sure it’s possible, but it probably requires a few undiscovered tricks

nicolaerosia:
great, will make a release after this PR as well?

tcuongd:
> I don’t know if anyone has successfully cross-compiled cmdstan for M1 using an intel Mac. I’m sure it’s possible, but it probably requires a few undiscovered tricks

Yeah alright I don't think it's worth the effort to investigate, I'll remove the arm64 option for macOS.

tcuongd:
https://github.com/facebook/prophet/actions/runs/3958615543
🚀

PR #2351, opened 2023-01-22
setuptools is not only a dev dependency

dimbleby:
pyproject.toml declares that setuptools is a dev only dependency, but in fact it is needed unconditionally - per this import: https://github.com/facebook/prophet/blob/0bf05baf3c4ace5e02e18a432ec6c2d734a97034/python/prophet/models.py#L13.


dimbleby:
sorry, I don't care enough about this to be bothering with a CLA.  Take it or leave it!

tcuongd:
Thanks! Seems like this module is deprecated anyways, so I'll switch to using `importlib`. Thanks for raising :)

PR #2352, opened 2023-01-24
Migrate pkg_resources to importlib_resources

tcuongd:
So we don't need to have `setuptools` in the dependencies list, reported here: https://github.com/facebook/prophet/pull/2351

`pkg_resources` is deprecated in favour of `importlib_resources` anyways: https://importlib-resources.readthedocs.io/en/latest/migration.html

### Local testing (macOS)

```
python-F3JeUycv ❯ python -m pytest prophet/tests/
===================================================== test session starts =====================================================
platform darwin -- Python 3.8.10, pytest-7.2.1, pluggy-1.0.0
rootdir: /Users/cuong/projects/prophet/python
plugins: anyio-3.6.1
collected 55 items

prophet/tests/test_diagnostics.py ...........                                                                           [ 20%]
prophet/tests/test_prophet.py ............s...s.......................                                                  [ 92%]
prophet/tests/test_serialize.py ...                                                                                     [ 98%]
prophet/tests/test_utilities.py .                                                                                       [100%]
```

### DIFF
--- a/python/prophet/models.py
+++ b/python/prophet/models.py
@@ -9,8 +9,7 @@ from abc import abstractmethod, ABC
 from typing import Tuple
 from collections import OrderedDict
 from enum import Enum
-from pathlib import Path
-import pkg_resources
+import importlib_resources
 import platform
 
 import logging
@@ -59,11 +58,9 @@ class CmdStanPyBackend(IStanBackend):
     def __init__(self):
         import cmdstanpy
         # this must be set before super.__init__() for load_model to work on Windows
-        local_cmdstan = pkg_resources.resource_filename(
-            "prophet", f"stan_model/cmdstan-{self.CMDSTAN_VERSION}"
-        )
-        if Path(local_cmdstan).exists():
-            cmdstanpy.set_cmdstan_path(local_cmdstan)
+        local_cmdstan = importlib_resources.files("prophet").joinpath(f"stan_model/cmdstan-{self.CMDSTAN_VERSION}")
+        if local_cmdstan.exists():
+            cmdstanpy.set_cmdstan_path(str(local_cmdstan))
         super().__init__()
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


tcuongd:
@WardBrian could you please take a look, should be quick (don't know why I still can't add you as a reviewer 😓)

python/prophet/models.py
```diff
@@ -72,11 +69,8 @@ def get_type():
 
     def load_model(self):
         import cmdstanpy
-        model_file = pkg_resources.resource_filename(
-            'prophet',
-            'stan_model/prophet_model.bin',
-        )
-        return cmdstanpy.CmdStanModel(exe_file=model_file)
+        model_file = importlib_resources.files("prophet").joinpath("stan_model/prophet_model.bin")
```
WardBrian, line 72: The cannonical way to join multiple paths is to either pass them as multiple arguments like `.joinpath("stan_model", "prophet_model.bin")`, or to use the `/` operator, like 
```suggestion
        model_file = importlib_resources.files("prophet") / "stan_model" / "prophet_model.bin"
```

python/prophet/models.py
```diff
@@ -59,11 +58,9 @@ class CmdStanPyBackend(IStanBackend):
     def __init__(self):
         import cmdstanpy
         # this must be set before super.__init__() for load_model to work on Windows
-        local_cmdstan = pkg_resources.resource_filename(
-            "prophet", f"stan_model/cmdstan-{self.CMDSTAN_VERSION}"
-        )
-        if Path(local_cmdstan).exists():
-            cmdstanpy.set_cmdstan_path(local_cmdstan)
+        local_cmdstan = importlib_resources.files("prophet").joinpath(f"stan_model/cmdstan-{self.CMDSTAN_VERSION}")
```
WardBrian, line 61: Same as other comment, avoid using "/" in a path since this is less portable.

WardBrian (commented):
I believe you will also need to add importlib_resources to the set of requirements, at least for pre-3.9 releases. It's a small dependency, at any rate.

After the min version is 3.9, you can just use `importlib.resources` from the standard library to do this

tcuongd:
Thanks heaps @WardBrian ! All makes sense

PR #2363, opened 2023-02-08
change list or tuple data structure to set where possible

FlynnOwen:
When checking for membership, sets provide O(1) lookup time, compared to O(n) for tuples and lists.
I have changed (where I've seen possible) datastructures to sets when lookup is done. 

I've also changed structures when iterating from to set. Definitely open to changing this back.


tcuongd:
Thanks! I don't think it's worth making this change when the list is small though (only a handful of items); it'd make a very negligible runtime difference. Thanks for the thought though :)

Issue #1824, opened 2021-02-18
Cannot serialize numpy.float32


stanton119:
Hi,

I have an issue serialising a model that was fit with a column of type float32.

Version:
fbprophet == 0.7.1
pystan == 2.19.1.1

Example code:
```
import pandas as pd
import numpy as np
from fbprophet import Prophet
from fbprophet.serialize import model_to_json

df = pd.DataFrame()
df["ds"] = pd.date_range(start="2020-01-01", end="2021-01-01", freq="1MS")
df["y"] = np.random.rand(df.shape[0])
df["y"] = df["y"].astype("float32")

model = Prophet()
model.fit(df)
model_json = model_to_json(model)
```

I did a bit of investigation:
```
# after returning the model_json before serialization:
import json
for key in model_json:
    try:
        json.dumps(model_json[key])
    except TypeError:
        print(key)
```

This returns only `y_scale` and `type(model_json["y_scale"]) => numpy.float32`.

Within serialize.py `y_scale` is in the list `SIMPLE_ATTRIBUTES`, as such it is serialised directly rather than using np.to_list()

This may relate to pull request #1656 and issue #1701 

Thanks!

bletham:
Sorry for the slow reply on this! Thanks for providing replication code and figuring out y_scale as the issue.

I'm actually not able to reproduce the issue still, which I'm thinking now might be due to a difference in numpy version? Or python version? I'm not quite sure. Can you give your versions of numpy and python? I have numpy 1.19.1 and python 3.8.3.

This is the full code that works for me but which I expect to error based on your report:
```Python
import pandas as pd
import numpy as np
from fbprophet import Prophet
from fbprophet.serialize import model_to_json, model_from_json

df = pd.DataFrame()
df["ds"] = pd.date_range(start="2020-01-01", end="2021-01-01", freq="1MS")
df["y"] = np.random.rand(df.shape[0])
df["y"] = df["y"].astype("float32")

model = Prophet()
model.fit(df)
model_json = model_to_json(model)
model_new = model_from_json(model_json)  # should error?
```

stanton119:
Maybe it's an edge case with a specific version set... This was using:
Numpy: 1.19.5
Python 3.8.6

bletham:
I think we're making some progress - the issue might be in pandas actually.

This produces an error for me, since numpy float32s don't serialize:
```Python
import numpy as np
import json

a = np.float32(4.0)
json.dumps(a) # Error!
```

[...]

PR #2378, opened 2023-02-23
Make sure 'mu' and 'std' are not numpy floats

lior1990:
Hi,
I encountered a similar issue like https://github.com/facebook/prophet/issues/1824, where the target value (i.e. `y`) was np.float32, then the computed `y_scale` was np.float32 as well, which caused json serialization to fail.

In my scenario, the extra regressors were np.float32, so `mu` and `std` are also np.float32 and I'm unable to serialize the model (of course after calling `model_to_dict`).

The solution is just to convert it to float, same as in https://github.com/facebook/prophet/commit/9e4e87af9baad421442a62206a6cb705ef4f612f.

Thanks!

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -367,8 +367,8 @@ class Prophet(object):
                 else:
                     standardize = True
             if standardize:
-                mu = df[name].mean()
-                std = df[name].std()
+                mu = float(df[name].mean())
+                std = float(df[name].std())
                 self.extra_regressors[name]['mu'] = mu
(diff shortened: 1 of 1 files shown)


tcuongd:
pandas strikes again :joy: Thanks for the fix, I'll kick off CI and it should be good to merge after that.

tcuongd:
One test is failing due to the same issue as in https://github.com/facebook/prophet/pull/2401#issuecomment-1537441500
This is safe to merge

tcuongd (approved):

PR #2401, opened 2023-04-02
Refactor tests - use fixtures and parametrization, allow backend argument

tcuongd:
## Changes

* Use `pytest` style assertions, fixtures, and parametrization.
* Broke up tests into different classes and functions for a clearer grouping of what functionality is being tested.
* Added `backend` argument to all calls to `Prophet()`, which allows us to test end-to-end with any new backends (e.g. https://github.com/facebook/prophet/pull/1973).
  * Note that the `backend` parametrization is done with `pytest_generate_tests()`, which re-runs all tests with the specified backend  argument.
  * This is not the most efficient, since some tests (e.g. fourier series, piecewise function outputs) do not rely on the PPL backend. However, it is easier to manage (we don't need to consider which tests would / wouldn't require a parametrization), and allows us to catch unintended interactions between intermediate outputs and the inputs to the PPL backend.

## Local testing
```
❯ pipenv run python -m pytest prophet/tests/ --backend CMDSTANPY
========================================== test session starts ==========================================
platform linux -- Python 3.8.6, pytest-7.2.1, pluggy-1.0.0
rootdir: /media/storage/cuong/github/prophet/python
plugins: anyio-3.6.1
collected 69 items                                                                                      

prophet/tests/test_diagnostics.py ..................                                              [ 26%]
prophet/tests/test_prophet.py ..s..s.........................................                     [ 94%]
prophet/tests/test_serialize.py ...                                                               [ 98%]
prophet/tests/test_utilities.py .                                                                 [100%]

============================== 67 passed, 2 skipped, 6 warnings in 34.04s ===============================
```

### DIFF
--- a/python/prophet/tests/conftest.py
+++ b/python/prophet/tests/conftest.py
@@ -40,5 +40,12 @@ def pytest_collection_modifyitems(config, items):
 
 
 def pytest_generate_tests(metafunc):
+    """
+    For each test, if `backend` is used as a fixture, add a parametrization equal to the value of the
+    --backend option.
+
+    This is used to re-run the test suite for different probabilistic programming language backends
+    (e.g. cmdstanpy, numpyro).
+    """
     if "backend" in metafunc.fixturenames:
         metafunc.parametrize("backend", metafunc.config.getoption("backend"))
--- a/python/prophet/tests/conftest.py
+++ b/python/prophet/tests/conftest.py
@@ -1,4 +1,19 @@
+from pathlib import Path
+
+import pandas as pd
 import pytest
+
+
+@pytest.fixture(scope="package")
+def daily_univariate_ts() -> pd.DataFrame:
+    """Daily univariate time series with 2 years of data"""
+    return pd.read_csv(Path(__file__).parent / "data.csv", parse_dates=["ds"])
+
+
+@pytest.fixture(scope="package")
+def subdaily_univariate_ts() -> pd.DataFrame:
+    """Sub-daily univariate time series"""
+    return pd.read_csv(Path(__file__).parent / "data2.csv", parse_dates=["ds"])
 
 
 def pytest_configure(config):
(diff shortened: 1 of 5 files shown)


tcuongd:
Hey @WardBrian ! Would you have time to sense check this tests/ refactor? I don't think checking every individual test is worth it, but I'd like your feedback on the configuration in `conftest.py` and the general pattern of the tests.

WardBrian:
This looks quite slick @tcuongd - I would add a comment to the `pytest_generate_tests` function explaining how it works/why it is there, but otherwise seems good.

A couple questions/suggestions:
1. Are you keeping around the class-based style for a particular reason? (e.g., namespacing?)
  I think you could remove them and all the `self` parameters and the test suite would be the same.
2. If you did want to avoid the re-testing you mention in your second bullet point, I think you could add something like

```python
def pytest_generate_tests(metafunc):
    backend = metafunc.config.getoption("backend")
    if "backend" in metafunc.fixturenames:
        metafunc.parametrize("backend", backend)
    elif backend != DEFAULT_BACKEND: # skip re-running non-backend specific tests
        ...
```

tcuongd:
> This looks quite slick @tcuongd - I would add a comment to the `pytest_generate_tests` function explaining how it works/why it is there, but otherwise seems good.

Great call!

> A couple questions/suggestions:
> 
> 1. Are you keeping around the class-based style for a particular reason? (e.g., namespacing?)
>    I think you could remove them and all the `self` parameters and the test suite would be the same.

Yep for namespacing, it helps with test selection as per https://docs.pytest.org/en/6.2.x/usage.html#specifying-tests-selecting-tests

> 2. If you did want to avoid the re-testing you mention in your second bullet point, I think you could add something like
> 

Oh actually I think I misunderstood what the code was doing, it might already skip those tests -- I think it just adds a parametrize mark to the ones that do read from the `backend` fixture.

Anyways thanks heaps for the review Brian, and sorry I got to this a month late :(

tcuongd:
Note that tests have failed due to this bug with pandas 2.0: https://github.com/pandas-dev/pandas/issues/52595

We'll wait for pandas to push a fix instead of patching it, as it seems like a small bug that will be resolved quickly. Any PRs in the meantime that only have that one failing test can be merged.

PR #2414, opened 2023-04-21
Upgraded CMDSTAN_VERSION to 2.32.0.

chukarsten:
Figured it was time to upgrade the CMDSTAN backend.


tcuongd:
Thanks @chukarsten :) I've kicked off the CI runs but I don't expect to be successful; from memory the newest versions of cmdstan use an updated TBB library, so our `TBB_` path references will also need to be updated. The model compilation also failed last time I tried, but I didn't have time to look into it. If you can help dig into those issues, that would be amazing!

tcuongd:
This is now covered by https://github.com/facebook/prophet/pull/2428 !

PR #2423, opened 2023-05-08
OpenCL & Multicore Support

StoffAudioLLC:
I edited the python setup.py to add the option to compile with OpenCL and/or Multicore Support for faster runtimes. 

My dataframe (226x26000) ran 2x as fast on OpenCL vs single core version. 

In order to run this model properly, changes to CMDSTANPY must be made which I will be making a fork and pull request for soon. 

Feel free to reach out if I can be any help: Stoffaudio@gmail.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Prophet: Automatic Forecasting Procedure
+# Prophet: Automatic Forecasting Procedure With python OPENCL/Multi-Core options
 
 ![Build](https://github.com/facebook/prophet/workflows/Build/badge.svg)
 
--- a/python/setup.py
+++ b/python/setup.py
@@ -8,7 +8,8 @@ import platform
 from pathlib import Path
 from shutil import copy, copytree, rmtree
 from typing import List
-
+import cmdstanpy.utils as utils
+print(utils.cmdstan_path())
 from setuptools import find_packages, setup, Extension
 from setuptools.command.build_ext import build_ext
 from setuptools.command.build_py import build_py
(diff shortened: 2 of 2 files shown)


python/setup.py
```diff
@@ -18,12 +19,14 @@
...
 MODEL_TARGET_DIR = os.path.join("prophet", "stan_model")
-
+MODEL_GPU = False
+MODEL_MULTI_CORES = False
 CMDSTAN_VERSION = "2.26.1"
 BINARIES_DIR = "bin"
 BINARIES = ["diagnose", "print", "stanc", "stansummary"]
 TBB_PARENT = "stan/lib/stan_math/lib"
 TBB_DIRS = ["tbb", "tbb_2019_U8"]
+OPEN_CL = "-L"+"\"C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.8/lib/Win32/OpenCL.lib\""
```
AlexandroLuis, line 29: You should not use absolute path, instead use generic:
**Windows:**
`OPEN_CL = "-L\"" + "%CUDA_PATH%/lib/Win32/OpenCL.lib\""`
**Linux:**
`OPEN_CL = "-L\"" + "$CUDA_PATH/lib64/OpenCL.lib\""`

Note: **%CUDA_PATH%** is an environment variable that points to the CUDA installation directory.

StoffAudioLLC: Good idea! I'll work on adding that with a checker for the OS

StoffAudioLLC: Awesome, just added

AlexandroLuis (commented):
please, remove the donation button!

README.md
```diff
@@ -1,5 +1,5 @@
-# Prophet: Automatic Forecasting Procedure
-
+# Prophet: Automatic Forecasting Procedure With python OPENCL/Multi-Core options
+[![Donate with PayPal](https://raw.githubusercontent.com/stefan-niedermann/paypal-donate-button/master/paypal-donate-button.png)](https://www.paypal.com/donate?hosted_button_id=9ELH753DDE98Y)
```
AlexandroLuis, line 2: remove it.

python/setup.py
```diff
@@ -4,11 +4,13 @@
...
 
 import os
+import sys
 import platform
 from pathlib import Path
 from shutil import copy, copytree, rmtree
 from typing import List
-
+import cmdstanpy.utils as utils
+print(utils.cmdstan_path())
```
AlexandroLuis, line 13: remove path info

donggu-kang:
```MODEL_MULTI_CORES = True```

I created a new virtual environment in ubuntu20 and Python 3.8.10
Then I changed this setting to True and built it with the command below.
```python3 setup.py install```

I confirmed that Prophet 1.1.2 version was installed in my virtual environment without an error message.

After making all conditions, including the dataset, as identical as possible, I measured the usage of each Core/Thread used by Prophet and how much time spent on computation was reduced.

However, disappointingly, multi-core was not used and the computation time was almost the same.

**Did I make a mistake?**


**Does there seem to be a problem with my environment?**
```
...
cmdstanpy 1.0.4
pystan 2.19.1.1
pandas 1.5.3
prophet 1.1.2
```

PR #2427, opened 2023-05-12
Updated documentation and README

arcAman07:
While installing prophet through anaconda there was no option to copy the command , like there was for other ways of installation through python or R.

I have fixed that by updating both the README and documentation. Tested the development locally and It was working perfectly.


tcuongd (approved):
Thanks :)

Issue #2402, opened 2023-04-04
 ERROR: Failed building wheel for prophet


apearson415:
I'm running this in git bash and cannot seem to find anything that will fix the error:

$ python -m pip install prophet
Collecting prophet
  Using cached prophet-1.1.2.tar.gz (67 kB)
  Installing build dependencies: started
  Installing build dependencies: still running...
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Collecting convertdate>=2.1.2
  Using cached convertdate-2.4.0-py3-none-any.whl (47 kB)
Collecting LunarCalendar>=0.0.9
  Using cached LunarCalendar-0.0.9-py2.py3-none-any.whl (18 kB)
Collecting holidays>=0.14.2
  Using cached holidays-0.22-py3-none-any.whl (412 kB)
Requirement already satisfied: matplotlib>=2.0.0 in c:\treasure\virt\lib\site-packages (from prophet) (3.7.1)
Requirement already satisfied: numpy>=1.15.4 in c:\treasure\virt\lib\site-packages (from prophet) (1.24.2)
Requirement already satisfied: pandas>=1.0.4 in c:\treasure\virt\lib\site-packages (from prophet) (2.0.0)
Requirement already satisfied: python-dateutil>=2.8.0 in c:\treasure\virt\lib\site-packages (from prophet) (2.8.2)
Requirement already satisfied: cmdstanpy>=1.0.4 in c:\treasure\virt\lib\site-packages (from prophet) (1.1.0)
Requirement already satisfied: tqdm>=4.36.1 in c:\treasure\virt\lib\site-packages (from prophet) (4.65.0)
Collecting pymeeus<=1,>=0.3.13
  Using cached PyMeeus-0.5.12-py3-none-any.whl
Requirement already satisfied: tzdata in c:\treasure\virt\lib\site-packages (from holidays>=0.14.2->prophet) (2023.3)
Collecting korean-lunar-calendar
  Using cached korean_lunar_calendar-0.3.1-py3-none-any.whl (9.0 kB)
Collecting hijri-converter
  Using cached hijri_converter-2.2.4-py3-none-any.whl (14 kB)
Collecting ephem>=3.7.5.3
  Using cached ephem-4.1.4-cp310-cp310-win32.whl (1.4 MB)
[...]

Issue #2250, opened 2022-08-18
[Python] Error when running fit(): Library not loaded: '@rpath/libtbb.dylib'


e-gons:
When running `Prophet().fit()` on a Mac (Monterey `12.5.1`, Apple M1 Max Chip), I get the following error message:

```
cmdstanpy - ERROR - Chain [1] error: terminated by signal 6 Unknown error: -6
Traceback (most recent call last):
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/tests/test_prophet.py", line 67, in <module>
    test_prophet()
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/tests/test_prophet.py", line 29, in test_prophet
    model.fit(df)
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/.venv/lib/python3.8/site-packages/prophet/forecaster.py", line 1171, in fit
    self.params = self.stan_backend.fit(stan_init, dat, **kwargs)
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/.venv/lib/python3.8/site-packages/prophet/models.py", line 112, in fit
    raise e
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/.venv/lib/python3.8/site-packages/prophet/models.py", line 102, in fit
    self.stan_fit = self.model.optimize(**args)
  File "/Users/xxx/xxx/xxx/xxx/xxx/xxx/.venv/lib/python3.8/site-packages/cmdstanpy/model.py", line 731, in optimize
    raise RuntimeError(msg)
RuntimeError: Error during optimization: console log output:

dyld[29330]: Library not loaded: '@rpath/libtbb.dylib'
  Referenced from: '/Users/xxx/xxx/xxx/xxx/xxx/xxx/.venv/lib/python3.8/site-packages/prophet/stan_model/prophet_model.bin'
[...]

Issue #2002, opened 2021-08-16
Can't install Fbprophet with Conda (or pip) in an Apple M1


jjcfrancisco:
I tried installing FBProphet via Conda in an Apple MacBook 13 M1 but I was unsuccessful (I also tried via pip). This is what I get:

```
fj@fj-MacBook-Pro ~ % conda install -c conda-forge fbprophet
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are not available from current channels:

  - fbprophet

Current channels:

  - https://conda.anaconda.org/conda-forge/osx-arm64
  - https://conda.anaconda.org/conda-forge/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

Any thoughts or a feasible way to install it? I am game.

tcuongd:
Hmmm simple one to start with: can you try `conda install -c conda-forge prophet`?

If that doesn't work we might have to do some more digging.

tcuongd:
@adriendod Could you post the rest of the message? It should tell us which package versions aren't compatible

adriendod:
Well this is weird because this is the full message! It doesn't tell us which package is causing the issue. :/

adriendod:
it still doesn't install on Mac m1, like OP, I get this now : 


```
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are not available from current channels:

  - prophet

Current channels:

  - https://conda.anaconda.org/conda-forge/osx-arm64
  - https://conda.anaconda.org/conda-forge/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

jmsmcniff:
Seeing the same issue. Is it possible to get an arm64 build for Apple Silicon via conda-forge? 🙏

jjTudu:
Hey,
Got error for first time, while installing prophet through terminal.
Installed pystan through Anaconda GUI.
restarted terminal, executed:

[...]

PR #2428, opened 2023-05-12
Build CmdStan and Stan Model in temporary directory

WardBrian:
This is a potential solution to #2402, at least on Windows. I believe the failures there are due to source installations ending up in places with very long path names, and that causing internal failures in `g++`.


WardBrian:
@tcuongd at this point the only errors appear to also be errors on the `main` branch. I will try to test this on Windows locally to make sure it works even if the path to the source is very long.

WardBrian:
Tested this on a local windows machine and confirmed installation succeeds even when in a very long/deeply nested path

tcuongd (approved):
Thanks so much for digging into this!! The additional copy-paste logic is a little annoying but not sure if I have a cleaner solution right now, so happy to merge as is.

Separate to the windows issue, upgrading to 2.31.0 should also help resolve the issues people are having on M1s (tested the install and import + fit on my M1 machine and it worked fine):

* https://github.com/facebook/prophet/issues/2002#issuecomment-1460813161
* https://github.com/facebook/prophet/issues/2250#issuecomment-1317709209

So overall this will be a huge win!

PR #2451, opened 2023-06-18
Fix setup.py stan file temp destination

tcuongd:
* We introduced logic to build stan and compile the prophet model in a temporary directory, since long directory names were causing issues in Windows.
* However there was a bug in specifying the temporary destination for the stan file: https://github.com/facebook/prophet/commit/9dd5c8cc4f6492c0385d3c4bf78f6cbed95e9845#r117902674
* @WardBrian fixed the issue with this patch: https://github.com/conda-forge/prophet-feedstock/pull/45#issuecomment-1591293459 and we're now merging the patch into main.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -95,7 +95,6 @@ def install_cmdstan_deps(cmdstan_dir: Path):
             cores=cpu_count(),
             progress=True,
         ):
-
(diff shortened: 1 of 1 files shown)


tcuongd:
Heya @WardBrian could you take a quick look, thanks!

WardBrian:
Seems good to me

PR #2461, opened 2023-07-14
Replace deprecated iteritems() with items()

### LINKED ISSUE #2460: incompatibility with Pandas 2.x / pd.DataFrame.iteritems
Pandas 1.5 [deprecated](https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#other-deprecations) and v2 removed `DataFrame.iteritems` completely since the `items` function has been [present all along](https://github.com/pandas-dev/pandas/blob/v1.0.0/pandas/core/frame.py#L891):

`iteritems` is still used - at least - in [`plot.py`](https://github.com/facebook/prophet/blob/v1.1.4/python/prophet/plot.py#L870) - therefore `get_forecast_component_plotly_props` won't work with Pandas 2.x

see also:
https://github.com/pandas-dev/pandas/pull/45321

michael72:
- pd.DataFrame.iteritems() was deprecated in pandas 1.5.0
- completely removed in pandas 2 -> incompatibility

fixes #2460


tcuongd (approved):
Thanks!

Issue #2437, opened 2023-05-24
[BUG] Prophet confidence intervals strictly overpredict and entirely miss the target values


vandonova:
Prophet is missing the prediction of the actual values `y` by what appears to be a constant intercept.

Plotting the predictions
<img width="777" alt="image" src="https://github.com/facebook/prophet/assets/16643414/34aed4e1-cfc1-4af1-8111-bcab45e82001">

Plotting the components
<img width="719" alt="image" src="https://github.com/facebook/prophet/assets/16643414/619eef7d-6aeb-48e1-aa05-347e335f59b1">

Plotting the residuals
<img width="736" alt="image" src="https://github.com/facebook/prophet/assets/16643414/be770e26-32db-456a-b8bc-8c44f4f30a79">

More information for repro:
`prophet==1.1.1`
data download = [bug_example.csv](https://github.com/facebook/prophet/files/11554777/bug_example.csv)

Repro code:
```
from prophet import Prophet
df = pd.read_csv('~/Downloads/bug_example.csv')

m = Prophet(interval_width = .95, **kwargs)
m.fit(df)
pred = m.predict(df)
m.plot(pred)
m.plot_components(pred)
```
Note: for plotting the residuals I use altair so I haven't included the code as to not require another install but you can easily reproduce by subtracting `y - yhat` and plotting the series.

PR #2470, opened 2023-07-31
Add support for MinMax scaling

yoziru:
Resolves https://github.com/facebook/prophet/issues/2437
This PR adds a new `scaling` parameter, with `minmax` scaling support (on top of the default `absmax` scaling in place right now). This is useful for certain datasets where `AbsMax` will compress the range (see https://github.com/facebook/prophet/issues/2437) to be `[0.99999... - 1.0]`, leading to bad fitting of parameters. This dataset is now added as a test in `data3.csv`.

For now the default `scaling` remains `absmax`, and de-serialization is backwards-compatible.

Here is how this looks like for the 3 test datasets.

| AbsMax | MinMax|
|---|---|
| output-data-absmax | output-data-minmax |
| output-data2-absmax | output-data2-minmax |
| output-data3-absmax | output-data3-minmax |


tcuongd:
This seems like a great contribution, thanks so much! I'll take a proper look soon.

yoziru:
@tcuongd re this change, settled on this approach in the PR because it has the most minimal changes to the code (and fully backwards-compat).

An alternative was to have a more generic scaler implementation so it can support other types of scalers or even user-defined funcs, but not sure if we need to go there. Wdyt?

python/prophet/serialize.py
```diff
@@ -144,6 +144,23 @@ def model_from_dict(model_dict):
...
+            if attribute == 'train_component_cols':
+                # Special handling because of named index column
+                df.columns.name = 'component'
+                df.index.name = 'col'
+            setattr(model, attribute, df)
+    
+    # backwards compat with prophet <=1.1.4
+    if 'scaling' not in model_dict:
+        model_dict['scaling'] = 'absmax'
+        model_dict['y_min'] = 0.
```
tcuongd, line 162: I might be missing something, but is there a reason we need to move this logic up here? I don't think we call / want to call any of the methods setup methods (setup_dataframe, initilize_scales) during deserialization.

Could we instead:

* add a method that handles backwards compatibility for SIMPLE_ATTRIBUTES
* put your logic in lines 160-162 in that method?

There's another PR, https://github.com/facebook/prophet/pull/2477/files, that would benefit from this too.

tcuongd (commented):
Thanks for the thoughtful and impactful contribution, as always! Really great tests too.

I just have a comment about `serialize.py`, let me know if I'm missing something.

Re: other types of scaling, I don't think we should worry about that for now. Seems like there's a lot of moving parts re: fit, predict, and serialize; if we want to allow more custom models in general it'll be a bigger effort.

python/prophet/serialize.py
yoziru: Yep, you're right, we don't call it so this moving around was not necessary. 

This is a lingering change, remembered I had an iteration trying to infer the scale from `df.history` if it wasn't set (using `minmax` as default). 

Added separate method for backwards compat 👍

tcuongd (approved):
Thanks so much again, @yoziru !

PR #2477, opened 2023-08-07
Adds holiday_mode independent of seasonality

CoreyBryant-everi:
Include a `holiday_mode` parameter so that seasonality and holiday can be distinct modes (additive/multiplicative).


tcuongd:
Thanks for raising! Some questions:

* I think at the least we'd want to retain current behaviour when `holiday_mode` is not set (i.e. if `seasonality_mode` is set but `holiday_mode` isn't, we'd want them to be the same).
* What docs, etc. should we update if we make this change?

CoreyBryant-everi:
I think https://github.com/facebook/prophet/pull/2477/commits/6d9b7e1a58b3f8fb7f3a90e37ef7e33f3804bfd3 should take care of the first.

https://github.com/facebook/prophet/pull/2477/commits/7996b5d6045779799cc27758f9dd3d0c3fcc1093 updates the docstring for the `Prophet` object to include `holiday_mode` parameter. I don't see a need to update formal documentation beyond that. But happy to if you feel it's necessary.

CoreyBryant-everi:
@tcuongd On my fork it now looks like it's just failing the backward compatibility test. Not sure what the best solution is for that.

tcuongd:
> @tcuongd On my fork it now looks like it's just failing the backward compatibility test. Not sure what the best solution is for that.

Thanks Corey! There's another PR to add an property to the model object here: https://github.com/facebook/prophet/pull/2470 that needs to do something similar. I think we let that PR go in first, and follow whatever pattern we decide there.

tcuongd:
It'd be great if you could rebase with main and add to this method:

https://github.com/facebook/prophet/blob/415bb6ef976eafa39bf01e71a6dfbdc32edef172/python/prophet/serialize.py#L132-L137

python/prophet/forecaster.py
```diff
@@ -53,6 +53,7 @@ class Prophet(object):
         optionally can have a column prior_scale specifying the prior scale for
         that holiday.
     seasonality_mode: 'additive' (default) or 'multiplicative'.
+    holiday_mode: 'additive' or 'multiplicative'. Defaults to seasonality_mode.
```
tcuongd, line 56: Sorry a small nit: could we change the name to `holidays_mode`, to be consistent with the other holidays parameters (holidays, holidays_prior_scale). Thank you!

python/prophet/forecaster.py
```diff
@@ -88,6 +89,7 @@ def __init__(
             daily_seasonality='auto',
             holidays=None,
             seasonality_mode='additive',
+            holiday_mode=None,
```
tcuongd, line 92: Could we place this parameter at the end so that the change doesn't affect people who are using positional keywords? Unfortunate i know.

tcuongd (approved):
Just a couple of small things, otherwise LGTM, thanks!

python/prophet/forecaster.py
CoreyBryant-everi: Makes sense. Done!

python/prophet/forecaster.py
CoreyBryant-everi: Hadn't even considered that, but I wouldn't want to be on the other side of that kind of change. Done!

CoreyBryant-everi:
Passing tests again on my fork after those changes.

tcuongd (approved):
LGTM

PR #2486, opened 2023-08-24
Add predict_columns parameter to cross validation

dchiang00:
The output of `cross_validation` currently does not return `trend`. In order to access it, this PR adds `predict_columns=None` as a new parameter to `cross_validation`. This way, the user is free to pass in `trend` to the list with `predict_columns=['ds', 'yhat', 'trend']` in order to retrieve it in the output.


python/prophet/diagnostics.py
```diff
@@ -58,7 +58,7 @@ def generate_cutoffs(df, horizon, initial, period):
     return list(reversed(result))
 
 
-def cross_validation(model, horizon, period=None, initial=None, parallel=None, cutoffs=None, disable_tqdm=False):
+def cross_validation(model, horizon, period=None, initial=None, parallel=None, cutoffs=None, disable_tqdm=False, predict_columns=['ds','yhat']):
```
arkid15r, line 61: I suggest making it None by default
```suggestion
def cross_validation(model, horizon, period=None, initial=None, parallel=None, cutoffs=None, disable_tqdm=False, predict_columns=None):
```

and dealing with it after, e.g.

```
if predict_columns is None:
    predict_columns = ['ds', 'yhat']
```
as in general the default params aren't to be mutable. 

The introduced parameter also needs to be documented.

dchiang00: Thanks for the suggestion! Updated the PR

python/prophet/diagnostics.py
```diff
@@ -121,7 +123,9 @@ def map(self, func, *iterables):
     df = model.history.copy().reset_index(drop=True)
     horizon = pd.Timedelta(horizon)
 
-    predict_columns = ['ds', 'yhat']
+    if predict_columns is None:
+        predict_columns = ['ds', 'yhat']
+        
     if model.uncertainty_samples:
         predict_columns.extend(['yhat_lower', 'yhat_upper'])
```
tcuongd, line 130: Would we ever want to _not_ output `ds`, `yhat`, `yhat_lower` and `yhat_upper`? I'm wondering if it's better to specify the argument as `extra_output_columns` and do

```python
predict_columns = predict_columns.extend([c for c in extra_output_columns if c not in predict_columns])
```

after line 126.

tcuongd:
It'd also be great to add a test to `test_diagnostics.py` demonstrating this new functionality. Thank you!

python/prophet/diagnostics.py
```diff
@@ -109,6 +107,10 @@ def map(self, func, *iterables):
                         for args in zip(*iterables)
                      ]
                      return results
+                     
+    disable_tqdm: if True it disables the progress bar that would otherwise show up when parallel=None
```
dchiang00, line 111: Existing docstring was out of order - placed disable_tqdm after parallel

python/prophet/diagnostics.py
dchiang00: Thank you for the suggestion, added!

dchiang00:
> It'd also be great to add a test to `test_diagnostics.py` demonstrating this new functionality. Thank you!
I also added a test `test_check_extra_output_columns_cross_validation` to demonstrate this functionality. Can you please review? Thank you!

python/prophet/diagnostics.py
```diff
@@ -120,8 +122,11 @@ def map(self, func, *iterables):
     
     df = model.history.copy().reset_index(drop=True)
     horizon = pd.Timedelta(horizon)
-
     predict_columns = ['ds', 'yhat']
+    
+    if extra_output_columns is not None:
```
tcuongd, line 127: Small one: could we move this after L130-131? To be sure that we're not creating duplicate columns.

python/prophet/diagnostics.py
```diff
@@ -109,6 +107,10 @@ def map(self, func, *iterables):
                         for args in zip(*iterables)
                      ]
                      return results
+                     
+    disable_tqdm: if True it disables the progress bar that would otherwise show up when parallel=None
+    extra_output_columns: A String or List of Strings e.g. 'trend' or ['trend'].
```
tcuongd, line 112: Hmm I don't think you've handled the String case. If the input is `trend` I think we'd be adding ['t', 'r', 'e', 'n', 'd'] to the output columns

python/prophet/tests/test_diagnostics.py
```diff
@@ -102,6 +102,37 @@ def mock_predict(df, model, cutoff, horizon, predict_columns):
             # check single forecast function called expected number of times
             assert n_calls == forecasts
 
+    def test_check_extra_output_columns_cross_validation(self, ts_short, monkeypatch, backend):
```
tcuongd, line 105: I think copy one of the other tests -- the monkeypatch doesn't actually help us test the real logic. 
We should run the actual function, but just choose time periods such that the run is quick, since we only need to test whether the extra columns appear.

python/prophet/diagnostics.py
tcuongd: Thank you!

tcuongd (changes requested):
Thank you, just a couple of things :)

tcuongd (approved):
LGTM

Issue #2438, opened 2023-05-25
PIP Prophet installation on m1 does not work


anton164:
It is still not possible to run `pip install prophet` on M1 without using conda. I understand this is an open source project and your time is limited, but it would be good to at the very least provide guidance on your installation page for m1 users. Better yet -- if you have bandwidth it would be great if the team would prioritize fixing this, especially given the focus in 2023 and beyond: 
https://medium.com/@cuongduong_35162/facebook-prophet-in-2023-and-beyond-c5086151c138

The team claims that it's seamless to install, but that's clearly not the case for many users (I assume many more will switch to M1 this year). Related issues:
- #2209 
- #2002 
- #2420 

Hope this doesn't come off as expecting too much, I just wanted to bring this to your attention!

Stack trace for Python 3.9:
```
> pip install prophet
<...shortened for brevit...>
          cmd_obj.run()
        File "/private/var/folders/z1/tk8nhdf94639kldpqnysbpvh0000gq/T/pip-build-env-xfkuwmr7/overlay/lib/python3.9/site-packages/setuptools/_distutils/command/build.py", line 131, in run
          self.run_command(cmd_name)
        File "/private/var/folders/z1/tk8nhdf94639kldpqnysbpvh0000gq/T/pip-build-env-xfkuwmr7/overlay/lib/python3.9/site-packages/setuptools/_distutils/cmd.py", line 318, in run_command
          self.distribution.run_command(command)
        File "/private/var/folders/z1/tk8nhdf94639kldpqnysbpvh0000gq/T/pip-build-env-xfkuwmr7/overlay/lib/python3.9/site-packages/setuptools/dist.py", line 1244, in run_command
          super().run_command(command)
        File "/private/var/folders/z1/tk8nhdf94639kldpqnysbpvh0000gq/T/pip-build-env-xfkuwmr7/overlay/lib/python3.9/site-packages/setuptools/_distutils/dist.py", line 988, in run_command
          cmd_obj.run()
        File "<string>", line 144, in run
        File "<string>", line 131, in build_models
        File "<string>", line 110, in build_cmdstan_model
        File "<string>", line 94, in install_cmdstan_deps
      RuntimeError: CmdStan failed to install in repackaged directory
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for prophet
Failed to build prophet
ERROR: Could not build wheels for prophet, which is required to install pyproject.toml-based projects
```

[...]

PR #2497, opened 2023-09-22
change CMDSTAN version for Apple M2

lucentcosmos:
This is needed to build prophet on Apple M2 Pro. 
```
% pip install prophet -e "git+https://github.com/lucentcosmos/prophet.git@441d77e#egg=prophet&subdirectory=python"

Obtaining prophet from git+https://github.com/lucentcosmos/prophet.git@441d77e#egg=prophet&subdirectory=python

  Cloning https://github.com/lucentcosmos/prophet.git (to revision 441d77e) to .....
  WARNING: Did not find branch or tag '441d77e', assuming revision or ref.
  Running command git checkout -q 441d77e
  Resolved https://github.com/lucentcosmos/prophet.git to commit 441d77e
  Installing build dependencies ... done
  Checking if build backend supports build_editable ... done
  Getting requirements to build editable ... done
  Preparing editable metadata (pyproject.toml) ... done
Collecting prophet
..........
Building wheels for collected packages: prophet
  Building editable for prophet (pyproject.toml) ... done
  Created wheel for prophet: filename=prophet-1.1.4a0-0.editable-py3-none-macosx_13_0_arm64.whl size=4722 sha256=33ad80ae8bd2993bff5dc775675e3be5ca6797466ee09d7eed7b32b99bbcc122
  Stored in directory: /private/var/folders/d5/wjs2_3q13f31k5xnnc_t_1bc0000gq/T/pip-ephem-wheel-cache-iv4bmayp/wheels/4d/14/7b/6942cc1367b12354825984ff85aacc56d594630a89181f54cb
Successfully built prophet
Installing collected packages: prophet
  Attempting uninstall: prophet
    Found existing installation: prophet 1.1.4a0
    Uninstalling prophet-1.1.4a0:
      Successfully uninstalled prophet-1.1.4a0
Successfully installed prophet-1.1.4a0

```

### DIFF
--- a/python/prophet/models.py
+++ b/python/prophet/models.py
@@ -54,7 +54,7 @@ class IStanBackend(ABC):
 
 
 class CmdStanPyBackend(IStanBackend):
-    CMDSTAN_VERSION = "2.31.0"
+    CMDSTAN_VERSION = "2.33.0"
     def __init__(self):
         import cmdstanpy
         # this must be set before super.__init__() for load_model to work on Windows
--- a/python/prophet/models.py
+++ b/python/prophet/models.py
@@ -6,7 +6,8 @@
 
 from __future__ import absolute_import, division, print_function
 from abc import abstractmethod, ABC
-from typing import Tuple
+from dataclasses import dataclass
+from typing import Sequence, Tuple
 from collections import OrderedDict
 from enum import Enum
 import importlib_resources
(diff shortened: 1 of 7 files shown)


python/prophet/models.py
```diff
@@ -54,7 +54,7 @@ def sampling(self, stan_init, stan_data, samples, **kwargs) -> dict:
 
 
 class CmdStanPyBackend(IStanBackend):
-    CMDSTAN_VERSION = "2.31.0"
+    CMDSTAN_VERSION = "2.33.0"
```
nicolaerosia, line 57: latest is 2.33.1, https://github.com/stan-dev/cmdstan/releases

lucentcosmos: updated and tested on my M2

WardBrian:
Note that 2.32.1 should be good enough: https://github.com/stan-dev/cmdstan/issues/1158

@tcuongd I recommend merging this if CI passes, it should fix the remaining issues in #2503, #2438, and maybe even #2402

tcuongd:
The RMSE in one of the tests is now a bit off:

```
FAILED prophet/tests/test_prophet.py::TestProphetFitPredictDefault::test_fit_predict[CMDSTANPY-minmax] - AssertionError: backend: <prophet.models.CmdStanPyBackend object at 0x11dc0e160>
assert 11.917838773340984 == 11.13 ± 2.2e-01
  comparison failed
  Obtained: 11.917838773340984
  Expected: 11.13 ± 2.2e-01
```

I don't think this is a big deal especially since the other tests have passed, but @lucentcosmos could you do some quick fit / predict runs on your machine and check that the forecasts aren't doing something weird? If it's just down to noise then we can update the expected rmse value in the tests.

tcuongd:
Actually now that I think about it, it's weird that the test only fails for macOS given that it's seeded. We'd need to put a higher tolerance on the check then, if the fitting is actually working as expected.

WardBrian:
I believe this range of updates includes when we updated to Eigen 3.4, which did lead to some numerical changes on some architectures

tcuongd:
Keen to get this merged, but the test result is still weird to me. Just to correct myself from before, the discrepancy is actually happening on Intel Macs; I tested on my M1 and the result is ~11.04 with the same seeds which is fine. If anyone can test on an Intel that the forecast isn't doing anything silly that would be great. Otherwise I'll increase the tolerance to 0.1 (10% relative) and leave a note to revisit.

Issue #2257, opened 2022-09-01
Should not call `fig.tight_layout()` unconditionally


davidgilbertson:
I'm calling `plot()` on a model and passing in an `ax`, which is then used to plot the function.

This `ax` exists in a figure that is using [constrained layout](https://matplotlib.org/stable/tutorials/intermediate/constrainedlayout_guide.html), rather than [tight layout](https://matplotlib.org/stable/tutorials/intermediate/tight_layout_guide.html).

The issue is that the Prophet plot function calls `fig.tight_layout()` [on this line](https://github.com/facebook/prophet/blob/main/python/prophet/plot.py#L92).

I think it should either:
* Not do this if the user passes in an `ax`. In this case they're essentially saying "this is your canvas, draw onto this, please" and to reach up into the parent figure and change things is overstepping those bounds. OR
* Check `fig.get_tight_layout()` before calling `fig.tight_layout()`.
 
In fact, the internals of `tight_layout()`/`subplots_adjust()` does indeed check for the use of constrained layout and logs a warning, but this warning gets swallowed somewhere.

My workaround in the meantime is to just reset back to the `rcParams` values with:
```py
m.plot(forecast, ax=ax).set_constrained_layout(None)
```

pranith7:
@tcuongd i want to work on this issue

adit299:
@pranith7 Did you make any progress on this issue? I would like to take a look.

alhridoy:
Hi @tcuongd I created a PR addressing this issue. Would you please check that.

sivapungamuru:
nice thank you

richardtorres314:
@tcuongd Would #2504 merged consider this closed? Thank you!

PR #2504, opened 2023-10-05
Check for tight_layout before applying it in plot function

alhridoy:
Issue:  #2257

This pull request addresses an issue with the `plot()` function in the Prophet library. Previously, the function was calling `fig.tight_layout()` unconditionally, which caused issues when the user's figure was using a constrained layout instead of a tight layout.

Changes made:
- Modified the `plot()` function to check if `fig.get_tight_layout()` is `True` before calling `fig.tight_layout()`.
- This change only affects cases where an `ax` (matplotlib axes) is passed to the `plot()` function.

These changes have been tested to ensure they work as expected. Please review and let me know if any further modifications are required.

### DIFF
--- a/python/prophet/plot.py
+++ b/python/prophet/plot.py
@@ -63,6 +63,7 @@ def plot(
     -------
     A matplotlib figure.
     """
+    user_provided_ax = False if ax is None else True  # Track whether ax was provided by the user
     if ax is None:
         fig = plt.figure(facecolor='w', figsize=figsize)
         ax = fig.add_subplot(111)
@@ -89,13 +90,9 @@ def plot(
     ax.set_ylabel(ylabel)
     if include_legend:
         ax.legend()
-    # Check if user passed in an ax
-    if ax is None:
+    # Only call fig.tight_layout() if the user didn't provide an ax
+    if not user_provided_ax:
         fig.tight_layout()
-    else:
-        # Check fig.get_tight_layout() before calling fig.tight_layout()
-        if fig.get_tight_layout():
-            fig.tight_layout()
     return fig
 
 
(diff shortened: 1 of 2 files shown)


alhridoy:
Hi @tcuongd 

I noticed that the `fix-tight-layout-issue` branch was deleted. I wanted to check if this means my changes have been merged, or if there are any further actions I need to take on this pull request or if my changes are not accepted. 

Thank you for your time and consideration.

python/prophet/plot.py
```diff
@@ -89,7 +89,13 @@ def plot(
     ax.set_ylabel(ylabel)
     if include_legend:
         ax.legend()
-    fig.tight_layout()
+    # Check if user passed in an ax
+    if ax is None:
```
tcuongd, line 93: We first check if `ax` is passed in lines 66-70, and assign one if it's not. So we'd have to create a variable there to track whether we're dealing with a user-provided ax.

python/prophet/plot.py
```diff
@@ -89,7 +89,13 @@ def plot(
...
+        fig.tight_layout()
+    else:
+        # Check fig.get_tight_layout() before calling fig.tight_layout()
+        if fig.get_tight_layout():
```
tcuongd, line 97: Could I confirm what this is meant to check? I'm also happy to call tight_layout() only when the user hasn't provided an ax.

alhridoy: I made changes based on your feedback. The changes include:

1. Tracking whether the ax was provided by the user or not. This is done by creating a new variable user_provided_ax at the start of the function.

2. Only calling fig.tight_layout() if the ax was not provided by the user. This is done by wrapping the fig.tight_layout() call in an if-statement that checks the user_provided_ax variable.

tcuongd (approved):
Thanks for this :)

alhridoy:
@tcuongd Would you consider this PR to merge, if it fits well now.

Issue #2072, opened 2021-11-19
refitting already fitted model


GabrieleSW:
I have declared and fitted a model with the following code:
model= Prophet(weekly_seasonality=False, daily_seasonality=True, changepoint_prior_scale=0.5)

and i add 4 regressor
model.add_regressor('regressor1')
model.add_regressor('regressor2')
model.add_regressor('regressor3')
model.add_regressor('regressor4')

i follow rules in "Updating fitted models" section at link: https://facebook.github.io/prophet/docs/additional_topics.html#flat-trend-and-custom-trends

But **sometimes** i try to use this function to refit a new model the following two error pops up:

1. **mismatch in dimension declared and found in context; processing stage=parameter initialization; variable name=delta; position=0; dims declared=(3); dims found=(25)**
2. **invalid index to scalar variable** - this happens in function **stan_init**

Can you help me to understand what are the problems?

tcuongd:
The code in the docs will work if you keep the same number of parameters. By adding 4 regressors, you're adding 4 extra parameters, so if you'll need to pass initial values for these 4 parameters as well. 

In the `def stan_init(m)` function, `res['beta']` picks up the coefficients of the existing seasonality/holiday/regressor features. You'll need to use `np.hstack` to add 4 additional columns to this array. The initial values can all be zeroes, i.e.:

```python
...
res['beta'] = np.hstack(m.params['beta'][0], np.zeros(4))

return res
```

GabrieleSW:
Thanks for the response.  

I don't understand the response, i try to explain how my software works.

**pr= Prophet(weekly_seasonality=False, daily_seasonality=True, changepoint_prior_scale=0.5)
pr.add_regressor('regressor1')
pr.add_regressor('regressor2')
pr.add_regressor('regressor3')
pr.add_regressor('regressor4')

model = pr.fit(dataframe)** 

#dataframe contains 6 columns:
# ds: timestamp
# y: data to predict
# regressor1
# regressor2
# regressor3
# regressor4

 My models contains in beta data for regressor, is it correct?

I save my model in json format:
**with open(path, 'w') as fout:
      json.dump(model_to_json(model), fout)  # Save model** 

where path is the system operative path where i want to save my json file.

-----------------------------------

When new data will come:

**with open(path, 'r') as fin:
       old_model = model_from_json(json.load(fin))**  # Load model, same path as before

# New prophet instance with the same parameters then old_model
**prophet = Prophet(  growth=old_model.growth,
[...]

PR #2527, opened 2023-11-14
Issues that occur when loading model files

donggu-kang:
Hi. The issue below appears to occur when the Prophet model loads an existing trained model on time series data with constant values (e.g. a single value of 1).

An error occurs when loading the model using the original code block below. The same error occurs in experiments such as unit func(1) and unit func(10).

```
def warm_start_params(m):
    res = {}
    for pname in ['k', 'm', 'sigma_obs']:
        if m.mcmc_samples == 0:
            res[pname] = m.params[pname][0][0]
        else:
            res[pname] = np.mean(m.params[pname])
    for pname in ['delta', 'beta']:
        if m.mcmc_samples == 0:
            res[pname] = m.params[pname][0]
        else:
            res[pname] = np.mean(m.params[pname], axis=0)
    return res
```

```
2023-11-13 17:57:52,173 - [CRITICAL] - **********:250 - ERROR - invalid index to scalar variable.
2023-11-13 17:57:52,173 - [CRITICAL] - **********:251 - Traceback (most recent call last):
  File "**********.py", line 204, in **********
    ********** = adf.AdfForeWorker(tag_id=tag_id,
  File "**********.py", line 486, in **********
    warm_start_params(m)
  File "**********.py", line 476, in warm_start_params
    res[pname] = m.params[pname][0][0]
IndexError: invalid index to scalar variable.
```

If a single value is used for model fitting as shown above, the Prophet model cannot learn the various statistical variabilities required for prediction. Therefore, some parameters may be set differently than expected.

Personally, I think this is what's causing the IndexError in the warm_start_params function.

That is, when a model is trained on constant-valued (non-volatile) data, some parameters may be set to scalar values or arrays of lower dimension than expected. (The above exception occurs if the expected value is a two-dimensional array, but is actually a one-dimensional array.)

This means that learning from data that only has constant values may result in some parameters (e.g. trend point (delta), seasonal component (beta)) being set differently than expected.

This means that the parameter can be set to a single value or an array of lower dimensions than expected.

Even if it is saved as a low-dimensional array, the model should be able to be loaded without error.

I registered this PR to update the above code to the official documentation.

Thank you always for creating Prophet.

### DIFF
--- a/docs/_docs/additional_topics.md
+++ b/docs/_docs/additional_topics.md
@@ -204,15 +204,31 @@ def warm_start_params(m):
     """
     res = {}
     for pname in ['k', 'm', 'sigma_obs']:
-        if m.mcmc_samples == 0:
-            res[pname] = m.params[pname][0][0]
-        else:
-            res[pname] = np.mean(m.params[pname])
+        param = m.params.get(pname)
+        if param is not None:
+            if m.mcmc_samples == 0:
+                if np.isscalar(param):
+                    res[pname] = param
+                elif param.ndim == 1:
+                    res[pname] = param[0]
+                else:
+                    res[pname] = param[0][0]
+            else:
+                res[pname] = np.mean(param)
+
     for pname in ['delta', 'beta']:
-        if m.mcmc_samples == 0:
-            res[pname] = m.params[pname][0]
-        else:
-            res[pname] = np.mean(m.params[pname], axis=0)
+        param = m.params.get(pname)
+        if param is not None:
+            if m.mcmc_samples == 0:
+                if np.isscalar(param):
+                    res[pname] = param
+                elif param.ndim == 1:
+                    res[pname] = param[0]
+                else:
+                    res[pname] = param[0]
+            else:
+                res[pname] = np.mean(param, axis=0)
+
     return res
 
 df = pd.read_csv('https://raw.githubusercontent.com/facebook/prophet/main/examples/example_wp_log_peyton_manning.csv')


donggu-kang:
https://github.com/facebook/prophet/issues/2072#issue-1058211390

tcuongd:
When there's a constant value series, does the model fit actually succeed? I'm wondering if the values in `.params` are just the prior values. And if the model fit doesn't succeed / it's not a good model, I'm not sure why people would want to use it as a warm start.

Thanks for catching the out of index error. But I think I'd rather catch those cases and raise an error (e.g. `model used for warm start does not have properly fitted parameters`). It's a better user experience imo than through passing the old parameters, which won't hold much information or won't give any advantages to fitting the new model.

donggu-kang:
Hi,

I apologize for the delay in my response.
Thank you again for developing Prophet.

1) I will conduct further experiments to verify whether the model fit actually succeeds with a constant value series.

2) In my opinion, a warning message should be sufficient to notify users. Our system, which utilizes Prophet, performs time series analysis and forecasting every midnight. Prophet is integrated into our system, analyzing the past n days of time series data for multiple monitoring targets (called `Tag`s. Ex. sensor values) to predict the normal range for the next day and detect anomalies.

**In our manufacturing environment, time series data can occasionally consist of single values.** 
If Prophet throws an error instead of a warning in such scenarios, it will fail to make predictions and thus fail to detect anomalies. 

**Therefore, even if a constant value is observed, the prediction must be a constant value. (Without errors or prediction failures)**

**There are cases where Prophet is integrated into systems for anomaly detection like ours. Even if a warning is issued, the .forecast should still succeed.** Please consider this context. Thanks.

tcuongd:
Sorry for the delay!

> In our manufacturing environment, time series data can occasionally consist of single values.
If Prophet throws an error instead of a warning in such scenarios, it will fail to make predictions and thus fail to detect anomalies.

I think you should check how many values are in your time series before you try to fit a Prophet model. Or we can throw an error with a specific message that end users can match against. But I don't think we should return a Prophet object or fit result; no other ML package does this either. 

I'll close this PR for now.

Issue #2322, opened 2022-12-21
Prophet reordering forecasts when predicting for a data frame that is not ordered by date


cranedroesch:
When running `pred = prophet.predict(future_df)`, the rows of `pred` are not guaranteed to correspond to the rows of `future_df`.

Here is a MWE:

```
import pandas as pd
from prophet import Prophet
import numpy as np

df = pd.DataFrame(
    dict(
        ds=pd.date_range(start="2022-01-01", periods=100),
        t=list(range(100)),  # time trend variable
        x=np.random.uniform(100),  # covariate
    )
)
df["y"] = df.t + 10 * df.x + np.random.normal(100)  # DGP is function of x and the time trend, plus light noise

m = Prophet()
m.add_regressor("x")
m.fit(df)

pred_df = pd.concat(
    [
        pd.DataFrame(
            dict(
                ds=pd.date_range(start="2022-04-11", periods=7),
                x=100,  # covariate = 1
            )
        ),
        pd.DataFrame(
            dict(
                ds=pd.date_range(start="2022-04-11", periods=7),
                x=-100,  # covariate = -1
            )
        ),
    ]
)
```
the input `pred_df` has the following date ordering:
```
pred_df.ds.head()
Out[19]: 
0   2022-04-11
1   2022-04-12
2   2022-04-13
3   2022-04-14
4   2022-04-15
```

however the output of the predict method has the following date ordering:
```
m.predict(pred_df).ds.head()
Out[20]: 
0   2022-04-11
1   2022-04-11
2   2022-04-12
3   2022-04-12
4   2022-04-13
```
I think that this is a bug.  However, if it's not considered a bug then at minimum it's deeply confusing.  Perhaps it could throw a warning and ideally return the index of the input in the output of the `predict` method, so that it could be sorted.

cranedroesch:
Happily, there is an easy fix:

```
class IDXProphet(Prophet):

    def predict(self, df: pd.DataFrame = None, vectorized: bool = True) -> pd.DataFrame:
        """Predict using the prophet model.

        Parameters
        ----------
        df: pd.DataFrame with dates for predictions (column ds), and capacity
            (column cap) if logistic growth. If not provided, predictions are
            made on the history.
        vectorized: Whether to use a vectorized method to compute uncertainty intervals. Suggest using
            True (the default) for much faster runtimes in most cases,
            except when (growth = 'logistic' and mcmc_samples > 0).

        Returns
        -------
[...]

PR #2568, opened 2024-04-04
Use stable sort when sorting values in forecaster

natl:
When Prophet prepares a dataframe for fitting or predicting, it runs a sort on the `ds` column.

This changes the sort algorithm from the pandas default sort to mergesort, which is a stable sort.

This is important in situations where multiple regressors are present and so the fitted dataframe is not unique by date.

Users may do something like the following:

```.py
df = <some dataframe>
# Example df
# ds            x   y   
# 2020-01-01    0   3.2
# 2020-01-01    1   4.1
# 2020-01-02    0   3.7
# 2020-01-02    1   4.3

m = Prophet()
m.add_regressor('x')
m.fit(df)

predict_df = m.predict(df)
df['prediction'] = predict_df['yhat']
```

In `m.predict(df)`, there is a call to `df.sort_values('ds')`. As this is not a stable sort, sometimes the `x` column above can get switched around, even if `df` is sorted. A stable sort addresses this.

This partially addresses #2322 if you construct the input correctly.

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -320,7 +320,7 @@ class Prophet(object):
 
         if df.index.name == 'ds':
             df.index.name = None
-        df = df.sort_values('ds')
+        df = df.sort_values('ds', kind='mergesort')
         df = df.reset_index(drop=True)
 
         self.initialize_scales(initialize_scales, df)


tcuongd (approved):
Thanks! I'm not sure Prophet was intended to be used the way you described (i.e. as a regression engine). Typical guidance is to have one regressor value per date. But I guess your example does just work out of the box, so this change makes sense to help people using it in this way. Only other consideration I'd have is performance but given time series datasets are usually too small for this to matter + this is a training time operation, I don't think it's a big deal.

Issue #2577, opened 2024-05-01
[BUG] Zero Division in diagnostics.performance_metrics() causing failed assertion


ThomasChia:
# Issue

When calculating the SMAPE metric in the `diagnostics.performance_metrics()` function, there is the possibility for a zero division if `y` and `yhat` are both zero.

```python
def smape(df, w):
    """Symmetric mean absolute percentage error
    based on Chen and Yang (2004) formula

    Parameters
    ----------
    df: Cross-validation results dataframe.
    w: Aggregation window size.

    Returns
    -------
    Dataframe with columns horizon and smape.
    """
    sape = np.abs(df['y'] - df['yhat']) / ((np.abs(df['y']) + np.abs(df['yhat'])) / 2)    <---- POSSIBLE ZERO DIVISION
    if w < 0:
        return pd.DataFrame({'horizon': df['horizon'], 'smape': sape})
    return rolling_mean_by_h(
        x=sape.values, h=df['horizon'].values, w=w, name='smape'
    )
```

This does not cause an error directly, however, it results in `np.nan` values where zero division occurs. When the `rolling_mean_by_h()` function is called, there is a `groupby()` which removes any `np.nan` values. This becomes an issue in the main `performance_metrics()` function with the following assert:

```python
assert np.array_equal(res['horizon'].values, res_m['horizon'].values)
```

This is part of a loop that checks each of the metrics ensuring that they are the same length and fails given the above scenario, as `np.nan` values are removed and that metric returns fewer values.

# Replication

Here is how you can replicate this issue:
```python
import pandas as pd
from prophet import Prophet
from prophet.diagnostics import cross_validation, performance_metrics

df = pd.read_csv('https://raw.githubusercontent.com/facebook/prophet/main/examples/example_wp_log_peyton_manning.csv')

df['ds'] = pd.to_datetime(df['ds'])
df.loc[df['ds'].dt.dayofweek == 6, 'y'] = 0

m = Prophet()
m.fit(df)

df_cv = cross_validation(m, '365 days', initial='1825 days', period='365 days')
df_cv['yhat'] = df_cv['yhat'].clip(lower=0)
metrics = performance_metrics(df_cv)
```

We set certain values in the training data to zero and clip negative values to create a scenario where `y` and `yhat` are both zero.

PR #2578, opened 2024-05-01
Fix Issue #2577 - Zero Division Error in diagnostics.performance_metrics() causing failed assertion

ThomasChia:
This PR resolves Issue #2577 

This issue is caused by zero division leading to `np.nan` values, so this PR adds a fix by filling `np.nan` values with zero during the calculation of the SMAPE metric.

This PR also updates the `test_performance_metrics()` method in the `TestPerformanceMetrics` class in `tests.test_diagnostics.py` to cover all metrics handling zero values for `y` and `yhat`.

### DIFF
--- a/python/prophet/diagnostics.py
+++ b/python/prophet/diagnostics.py
@@ -630,6 +630,7 @@ def smape(df, w):
     Dataframe with columns horizon and smape.
     """
     sape = np.abs(df['y'] - df['yhat']) / ((np.abs(df['y']) + np.abs(df['yhat'])) / 2)
+    sape = sape.fillna(0)
     if w < 0:
         return pd.DataFrame({'horizon': df['horizon'], 'smape': sape})
     return rolling_mean_by_h(
--- a/python/prophet/tests/test_diagnostics.py
+++ b/python/prophet/tests/test_diagnostics.py
@@ -234,7 +234,9 @@ class TestPerformanceMetrics:
             metrics=["coverage", "mse"],
         )
         assert set(df_horizon.columns) == {"coverage", "mse", "horizon"}
-        # Handle zero y and yhat and skip MAPE
+        # Skip MAPE
+        df_cv.loc[0, "y"] = 0.0
+        # Handle zero y and yhat
         df_cv["y"] = 0.0
         df_cv["yhat"] = 0.0
         df_horizon = diagnostics.performance_metrics(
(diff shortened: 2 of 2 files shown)


python/prophet/diagnostics.py
```diff
@@ -630,6 +630,7 @@ def smape(df, w):
     Dataframe with columns horizon and smape.
     """
     sape = np.abs(df['y'] - df['yhat']) / ((np.abs(df['y']) + np.abs(df['yhat'])) / 2)
+    sape = sape.fillna(0)
```
mikeedjones, line 633: are there tests covering these methods?

ThomasChia: There was a test looking at the `mape` metric underneath the wider `test_performance_metrics()` method, so I have extended that to test for zero values across all other metrics.

mikeedjones: Great nice one

python/prophet/tests/test_diagnostics.py
```diff
@@ -234,13 +234,13 @@ def test_performance_metrics(self, ts_short, backend):
             metrics=["coverage", "mse"],
         )
         assert set(df_horizon.columns) == {"coverage", "mse", "horizon"}
-        # Skip MAPE
-        df_cv.loc[0, "y"] = 0.0
+        # Handle zero y and yhat and skip MAPE
```
tcuongd, line 237: You can leave the MAPE skip test as-is and do the test for `0.0` y and `0.0` yhat afterwards.

I guess mdape doesn't have the same problem because it uses the rolling_median function?

python/prophet/tests/test_diagnostics.py
```diff
@@ -236,11 +236,13 @@ def test_performance_metrics(self, ts_short, backend):
         assert set(df_horizon.columns) == {"coverage", "mse", "horizon"}
         # Skip MAPE
         df_cv.loc[0, "y"] = 0.0
```
tcuongd, line 238: You'll still want to keep these calls:

```
df_horizon = diagnostics.performance_metrics(
            df_cv,
            metrics=["coverage", "mape"],
        )
assert set(df_horizon.columns) == {"coverage", "horizon"}
```

python/prophet/tests/test_diagnostics.py
ThomasChia: Updated!

Yep, rolling_median doesn't get rid of the np.nan values, they're still included with the groupby.

python/prophet/tests/test_diagnostics.py
ThomasChia: Ah thanks, missed that! Added back in now.

PR #2592, opened 2024-06-21
AttributeError: np.float_ fixed

shamalgithub:
changed the  NDArray[np.float_]  to  NDArray[np.float64] to be compatible with Numpy version 2.0

PR #2594, opened 2024-06-29
fix-numpy-deprecation

lordskyzw:
Fix: Replace deprecated np.float_ with np.float64 making it compatible with NumPy 2.0.


WardBrian:
Duplicate of https://github.com/facebook/prophet/pull/2592

Issue #2471, opened 2023-08-01
How to set only floor value to make sure predictions wont go negative?


RAKSHITH-U-R:
I was doing predictions on some data which is not suppose to go negative but Its going below zero, how to ensure that the forecast stays above zero? I tried using Uncertainty Intervals, but I have to mention cap value as well but I don't want to set any upper limit.

PR #2610, opened 2024-08-17
New parameter added in order to replace negative predictions with 0

mrtergl:
Sometimes, data series contain many values close to zero but are not expected to fall below zero. For example, a data frame with response times will likely have many near-zero values. In such cases, Prophet may produce yhat_lower or trend_lower metric values that are negative.

To address this issue, I have implemented a flag that sets **_trend, yhat, yhat lower and yhat upper_** values to zero in the future data frame. This flag can be configured when defining the model.

### DIFF
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -78,7 +78,7 @@ class Prophet(object):
     stan_backend: str as defined in StanBackendEnum default: None - will try to
         iterate over all available backends and find the working one
     holidays_mode: 'additive' or 'multiplicative'. Defaults to seasonality_mode.
-    negative_values: bool check to set all yhat_lower negative prediction values in the DataFrame to 0.
+    negative_prediction_values: bool check to set all yhat_lower negative prediction values in the DataFrame to 0.
     """
 
     def __init__(
--- a/python/prophet/forecaster.py
+++ b/python/prophet/forecaster.py
@@ -25,6 +25,7 @@ logger = logging.getLogger('prophet')
 logger.setLevel(logging.INFO)
 NANOSECONDS_TO_SECONDS = 1000 * 1000 * 1000
 
+
 class Prophet(object):
     """Prophet forecaster.
 
@@ -77,6 +78,7 @@ class Prophet(object):
     stan_backend: str as defined in StanBackendEnum default: None - will try to
         iterate over all available backends and find the working one
     holidays_mode: 'additive' or 'multiplicative'. Defaults to seasonality_mode.
+    negative_values: bool check to set all yhat_lower negative prediction values in the DataFrame to 0.
     """
 
     def __init__(
@@ -99,6 +101,7 @@ class Prophet(object):
             stan_backend=None,
             scaling: str = 'absmax',
             holidays_mode=None,
+            negative_prediction_values=True
     ):
         self.growth = growth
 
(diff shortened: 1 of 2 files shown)


mrtergl:
This improvement can solve issue #2471

mrtergl:
Hey, @tcuongd , could you please review my pull request?

donggu-kang:
I think this pull request is useful.

I am currently using Prophet for various purposes.
One of them is to predict traffic trends/detect anomalies.

I currently use it as a code to replace negative values ​​with 0 in the post-processing process.

Because even if I do not train Prophet with values ​​less than 0, there were cases where the prediction result was predicted as a negative number less than 0.

So I personally think this suggestion can be useful.

mrtergl:
Hi @HumptyHans,
Thank you for noticing, you are right about seasonality values being able to go below zero. I removed the part and tested it.

HumptyHans:
Hey @mrtergl,
Thanks for making changes! But without seasonality clipping, the resulting value can still go below zero.
Something needs to be done with 'additive_terms'. The easiest suggestion is to ensure that if it is negative, its absolute value doesn't exceed the trend value. 
I’m just not sure if it’s the correct approach from a statistical perspective.

mrtergl:
Hi @HumptyHans ,

I'm not sure I understand you well because of my lack of knowledge in the statistical area. From what I understood, seasonality values such as 'additive_terms' can be negative while the prediction stays positive. I made some refactoring on the code and I hope I got the right result for all cases without negative predictions.

With all seasonalities included, here is the result you will get if you set negative_prediction_values to false.

Screenshot 2024-10-11 at 00 19 02

As you can see seasonalities have both negative and positive values while prediction has always been positive

mrtergl:
@tcuongd Could you please check my pull request?

HumptyHans:
Hey @mrtergl 
I see that your last commit fixes exactly what I meant. Basically, the whole PR could have been reduced to these two lines:
```
if not self.negative_prediction_values:
    df2['yhat'] = df2['yhat'].clip(lower=0)
```

mrtergl:
Hi @HumptyHans ,

We should be able to make **_trend, yhat upper and  yhat lower_** predictions positive while the flag **_negative_prediction_values_** is set to _false_. So the code you provided will be insufficient.

But I removed the part where I clipped seasonalities' lower and upper values. So that, all seasonality predictions (including upper and lower values) can be negative even if the metric is _**non-negative**_ (such as count, latency).

mrtergl:
Hi @tcuongd,

Is there anything I should do before merge my pull request to main? Is this all okay to be merged ?

tcuongd:
Sorry for the delay! Unfortunately I'm not sure I would put this in the package. What is the benefit of this versus the user doing `future_df.clip(min=0)`?

The reason being, it's not completely accurate. "trend" is meant to represent the mean of the prediction distribution. So technically we should be clipping the prediction distribution, then computing trend and lower / upper.

So if this approach is an approximation, and it's a one-liner for the end user, I'd rather not have it as core functionality.

mrtergl:
Hi @tcuongd,

Thanks for the feedback. I understand your point about this being an approximation and easily achievable via `future_df.clip(min=0)`. However, I think it will be better to have it as a centralized flag, especially for users who may not be aware of the implications or who repeatedly encounter the same pattern.

> The reason being, it's not completely accurate. "trend" is meant to represent the mean of the prediction distribution. So technically we should be clipping the prediction distribution, then computing trend and lower / upper.

That makes sense — and I agree that ideally we’d clip the distribution itself before computing the forecast components. But as I see from the issues and what I experienced so far while using Prophet is, many users are concerned with avoiding negative values in outputs like yhat, trend, or intervals, especially when modeling quantities that logically can’t go below zero . That’s why I thought it would be useful, even if it only affects the output values and not the internal computation.

PR #2684, opened 2025-09-08
Use latest CmdStan

WardBrian:
Resolves a build failure on new XCode versions


tcuongd (approved):
Thank you!

WardBrian:
I don't believe there is an error in that code, and it hasn't changed in nearly 5 years. If you enable logging we can probably help you find the issue

PR #2720, opened 2026-01-28
full static typing support

jorenham:
This adds annotations for the entire public API, and validates the correctness using [Pyrefly](https://github.com/facebook/pyrefly).

With this, the type coverage of the public `prophet` API (excluding tests) is now at *almost* 100%. The one exception is `__init__.py`, but that will be resolved once   https://github.com/facebook/prophet/pull/2717 is merged.

This also adds the required `py.typed` and `pyproject.toml` classifier to make sure that users will also be able to fully benefit from all of these static typing goodies. See https://typing.python.org/en/latest/spec/distributing.html#packaging-typed-libraries for the story behind  `py.typed`.

I realize that this is a pretty beefy PR, but for what it's worth, I wrote these all by hand, so there's no need to worry about slop :). But things like this are difficult to split up, because adding/changing an annotation in module B can also affect module A, and vice-versa. That way, splitting things up is likely to come with a lot of overhead of going back and forth, as well as loads of merge-conflicts. So in my experience it's a lot less painful to annotate everything all at once.

This supersedes #2704.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -62,7 +62,7 @@ jobs:
         run: cd python && uv pip install ".[dev,parallel]"
 
       - name: Typecheck with pyrefly
-        run: pyrefly check --output-format=github
+        run: cd python && pyrefly check --output-format=github
 
   build-and-test-r:
 


tcuongd (approved):
This is amazing, thanks heaps for doing it by hand! 

People have asked for a docs site for a while too, which this will help with.

If the tests pass I'm happy to merge.

rrayy-25809:
Hello, I'm  the author of #2704 
Thanks for your interest in my idea and improving it in Great Direction!
I'd be really happy if this PR gets merged.

jorenham:
CI failure seems unrelated

tcuongd:
@jorenham Could you double check this typecheck error on main: https://github.com/facebook/prophet/actions/runs/21552346989/job/62102981975

jorenham:
> @jorenham Could you double check this typecheck error on main: [facebook/prophet/actions/runs/21552346989/job/62102981975](https://github.com/facebook/prophet/actions/runs/21552346989/job/62102981975)

The error seems to be caused by awkward typeshed annotations for `importlib.resources.files`, which returns a `Traversable` protocol (at runtime it's a `path.Path`), which has no `exists` method. I'll address this by casting it to the right type.

PR #2745, opened 2026-08-27
fix: close cross-validation executor pools

### LINKED ISSUE #2355: Using cross_validation results in OSError: [Errno 24] Too many open files
I have a Python script that does a simple hyperparameter search like this (minus irrelevant code):

```python
for df in dataframes:
    hyperparameters = {...}
    for params in product(hyperparameters):
        m = prophet.Prophet(**params)
        m.fit(df)
        cv = cross_validation(m, parallel='processes')
```

For small numbers of dataframes and hyperparameters it works fine, but if I start going above single digit counts of either of those, I start getting tracebacks like this:

```python
  File "/code.py", line 123, in get_prophet_hyperparameters
    cv = cross_validation(
  File "venv/lib/python3.9/site-packages/prophet/diagnostics.py", line 193, in cross_validation
    predicts = pool.map(single_cutoff_forecast, *iterables)
  File "/usr/lib/python3.9/concurrent/futures/process.py", line 726, in map
    results = super().map(partial(_process_chunk, fn),
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 597, in map
    fs = [self.submit(fn, *args) for args in zip(*iterables)]
  File "/usr/lib/python3.9/concurrent/futures/_base.py", line 597, in <listcomp>
    fs = [self.submit(fn, *args) for args in zip(*iterables)]
  File "/usr/lib/python3.9/concurrent/futures/process.py", line 697, in submit
    self._adjust_process_count()
  File "/usr/lib/python3.9/concurrent/futures/process.py", line 675, in _adjust_process_count
    p.start()
  File "/usr/lib/python3.9/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
  File "/usr/lib/python3.9/multiprocessing/context.py", line 277, in _Popen
    return Popen(process_obj)
  File "/usr/lib/python3.9/multiprocessing/popen_fork.py", line 19, in __init__
    self._launch(process_obj)
  File "/usr/lib/python3.9/multiprocessing/popen_fork.py", line 65, in _launch
    child_r, parent_w = os.pipe()
OSError: [Errno 24] Too many open files
```

After some digging, it appears to be because `cross_validation` creates the necessary pool executors, but [never closes them](https://github.com/facebook/prophet/blob/e665430/python/prophet/diagnostics.py#L165-L196).

The docs for `ProcessPoolExecutor` warn that [pools should be closed](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool), either manually or with the context manager:

> __Warning__: [multiprocessing.pool](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.pool) objects have internal resources that need to be properly managed (like any other resource) by using the pool as a context manager or by calling [close()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.close) and [terminate()](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate) manually. Failure to do this can lead to the process hanging on finalization.
>
> Note that it is __not correct__ to rely on the garbage collector to destroy the pool as CPython does not assure that the finalizer of the pool will be called (see [object.\_\_del__()](https://docs.python.org/3/reference/datamodel.html#object.__del__) for more information).

The last note matches my experience with attempting a `gc.collect()` after each loop iteration, which did not help. From what I can tell, the `cross_validation` function needs to close/terminate the pools that it creates, and this issue will be fixed. I did try adding this to each iteration of the loop, which didn't seem to help:

```python
from multiprocessing import active_children
...
for child in active_children():
    child.terminate()
```

So it seems like there isn't an easy way to hack around this outside of fixing it in Prophet itself.

mameikagou:
Fixes #2355.

This applies the executor cleanup proposed in #2738 and adds focused regression coverage.

## What changed

- Shut down internally created thread and process executors after cross-validation, including error paths.
- Preserve caller-owned executors and Dask clients.
- Cover repeated runs, mapping failures, and caller-owned executors.

## Verification

- Focused diagnostics tests: 5 passed
- Python compile check
- `git diff --check`
- Pyrefly 0.50.0: 0 errors (18 existing suppressions), all 9 configured modules checked. The local hidden-directory exclusion heuristic was disabled; project includes/excludes were unchanged.

### DIFF
--- a/python/prophet/diagnostics.py
+++ b/python/prophet/diagnostics.py
@@ -212,15 +212,18 @@ def cross_validation(
 
     if parallel:
         valid = {"threads", "processes", "dask"}
+        _owns_pool = False
 
         if parallel == "threads":
             pool = concurrent.futures.ThreadPoolExecutor()
+            _owns_pool = True
         elif parallel == "processes":
             if sys.platform.startswith("win") or sys.platform == "darwin":
                 ctx = multiprocessing.get_context("spawn")
             else:
                 ctx = multiprocessing.get_context("forkserver")
             pool = concurrent.futures.ProcessPoolExecutor(mp_context=ctx)
+            _owns_pool = True
         elif parallel == "dask":
             try:
                 from dask.distributed import get_client
@@ -228,6 +231,7 @@ def cross_validation(
                 raise ImportError("parallel='dask' requires the optional "
                                   "dependency dask.") from e
             pool = get_client()
+            # get_client() returns a client managed by the caller.
             # delay df and model to avoid large objects in task graph.
             df, model = pool.scatter([df, model])
         elif hasattr(parallel, "map"):
@@ -242,10 +246,14 @@ def cross_validation(
         iterables = zip(*iterables)
 
         logger.info("Applying in parallel with %s", pool)
-        predicts = pool.map(single_cutoff_forecast, *iterables)
-        if parallel == "dask":
-            # convert Futures to DataFrames
-            predicts = cast("dd.Client", pool).gather(predicts)
+        try:
+            predicts = pool.map(single_cutoff_forecast, *iterables)
+            if parallel == "dask":
+                # convert Futures to DataFrames
+                predicts = cast("dd.Client", pool).gather(predicts)
+        finally:
+            if _owns_pool:
+                pool.shutdown(wait=True)
 
     else:
         predicts = [
--- a/python/prophet/diagnostics.py
+++ b/python/prophet/diagnostics.py
@@ -212,18 +212,15 @@ def cross_validation(
 
     if parallel:
         valid = {"threads", "processes", "dask"}
-        _owns_pool = False
 
         if parallel == "threads":
             pool = concurrent.futures.ThreadPoolExecutor()
-            _owns_pool = True
         elif parallel == "processes":
             if sys.platform.startswith("win") or sys.platform == "darwin":
                 ctx = multiprocessing.get_context("spawn")
             else:
                 ctx = multiprocessing.get_context("forkserver")
             pool = concurrent.futures.ProcessPoolExecutor(mp_context=ctx)
-            _owns_pool = True
         elif parallel == "dask":
             try:
                 from dask.distributed import get_client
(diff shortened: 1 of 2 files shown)


python/prophet/diagnostics.py
```diff
@@ -242,10 +246,14 @@ def map(self, func, *iterables):
...
-        if parallel == "dask":
-            # convert Futures to DataFrames
-            predicts = cast("dd.Client", pool).gather(predicts)
+        try:
+            predicts = pool.map(single_cutoff_forecast, *iterables)
+            if parallel == "dask":
+                # convert Futures to DataFrames
+                predicts = cast("dd.Client", pool).gather(predicts)
+        finally:
+            if _owns_pool:
```
tcuongd: At this stage it probably makes more sense to check whether `pool` is an instance of `concurrent.futures.Executor` rather than manually assign `_owns_pool`.

python/prophet/tests/test_diagnostics.py
```diff
@@ -84,6 +98,93 @@ def test_cross_validation(self, ts_short, parallel_method, backend):
...
+
+            def map(self, func, *iterables):
+                if raises:
+                    raise RuntimeError("map failed")
+                return [func(*args) for args in zip(*iterables)]
+
+            def shutdown(self, wait=True):
+                self.shutdown_calls += 1
+
+        monkeypatch.setattr(
```
tcuongd, line 132: Maybe it's easier to just subclass `concurrent.futures.ThreadPoolExecutor`?

tcuongd (approved):
LGTM but with a couple of suggestions. Type checking is also failing, PTAL. Thank you!

mameikagou:
Thanks, updated in 2cd2510. Cleanup now uses the Executor type guard while remaining limited to internally created pools. The tests subclass ThreadPoolExecutor and confirm that a caller-owned executor remains usable after cross-validation. All five focused tests and the configured Pyrefly 0.50.0 check pass locally.

